PR #19, opened 2019-05-21
Refactors system reports table, add sorting/filtering/searching

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-1285
This adds sorting, searching and filtering to the inventory-insights table. Its all done on the client and s all additive (yer filtering on what yer searchin for).

### what we look like (ss are in order)
Screen Shot 2019-05-29 at 12 47 04 PM
Screen Shot 2019-05-29 at 12 47 21 PM
Screen Shot 2019-05-29 at 12 47 46 PM

#### and when you can't find what yer lookin for...
Screen Shot 2019-05-29 at 1 08 20 PM

### PS
Scope has creeped a little, we're adding a new column in this work (as per the mock below)
image (1)


AllenBW:
cc @robcwilliams & @katierik

packages/inventory-insights/src/Insights.js
```diff
@@ -92,46 +109,79 @@ class InventoryRuleList extends Component {
...
-                    parent,
+                    parent: key,
                     fullWidth: true,
                     cells: [{
                         title: <ReportDetails key={ `child-${key}` } report={ value } kbaDetail={ kbaDetail }/>
                     }]
                 }
             ];
+
+            const isValidSearchValue = searchValue.length === 0 || rule.description.search(new RegExp(searchValue, 'i')) !== -1;
```
karelhala: You don't really need to check `length === 0` right? Also can you split this into special function?

packages/inventory-insights/src/Insights.js
```diff
@@ -92,46 +109,79 @@ class InventoryRuleList extends Component {
...
+            const isValidFilterValue = Object.keys(filters).length === 0 || Object.keys(filters).map((key) => {
```
karelhala: Can you move this to special function? And I think that `length === 0` is not needed as well.

packages/inventory-insights/src/Insights.js
```diff
@@ -92,46 +109,79 @@ class InventoryRuleList extends Component {
...
+                const filterValues = filters[key];
+                let rowValue;
+                switch (key) {
```
karelhala: I am not that big of a fan of switch, what do you think about this
```JS
function rowValue(value, key) {
  return ({
  'has_playbook': value.resolution[key],
  risk: value.resolution.resolution_risk[key]
}) || value[key]
}
```

And getting the value would be `rowValue(value, key)`

packages/inventory-insights/src/Insights.js
```diff
@@ -165,24 +215,96 @@ class InventoryRuleList extends Component {
...
                 </DropdownItem>
             ] }
         />;
     };
 
+    onSort = (_event, index, direction) => {
+        const { activeReports, kbaDetailsData, filters, searchValue } = this.state;
+        const attrIndex = { 2: 'publish_date', 3: 'total_risk' };
+        let sortedReports;
+        switch (index) {
```
karelhala: Same as above, can you perhaps change this to use hash instead of switch?

packages/inventory-insights/src/Insights.js
AllenBW: yeah! need to check length === 0 here, when there's no text input we would like to show all results (rather than no results).  as for splitting it into a function, can do! wondering what yer looking to achieve by that?

packages/inventory-insights/src/Insights.js
AllenBW: Same as above 😬 we need the zero compare, gotta show all the results

packages/inventory-insights/src/Insights.js
AllenBW: OH yeah i can do that! <3 <3 <3

packages/inventory-insights/src/Insights.js
karelhala: Oh okay, well with the function you don't have to recreate it for each row.

packages/inventory-insights/src/Insights.js
AllenBW: eehhhh doesn't quite work like that, would a pile of ifs be preferred to the switch?

AllenBW: ok lets just make em objects, thats a lil simpler

PR #38, opened 2019-06-04
Allow filter to use components as labels

dLobatog:


akofink:
The associated PR is merged, so this one should be good to merge as well, right @dLobatog?

PR #40, opened 2019-06-05
Fix some remaining UX issues for RHICOMPL-126

akofink:
Fixes issues 2, 4, 5, 8

Before:

image

image

After:

image

image

### DIFF
--- a/packages/inventory-compliance/src/SystemPolicyCard.js
+++ b/packages/inventory-compliance/src/SystemPolicyCard.js
@@ -40,28 +40,28 @@ class SystemPolicyCard extends React.Component {
         return (
             <Card>
                 <CardBody>
-                    <TextContent style={ { marginBottom: 'var(--pf-global--spacer--md)' } } >
-                        <Text style={ { marginBottom: '0' } } component={ TextVariants.small }>External policy</Text>
-                        <Text style={ { marginTop: '0', marginBottom: '0' } } component={ TextVariants.h4 }>{ this.state.policy.name }</Text>
+                    <TextContent className='margin-bottom-md'>
+                        <Text className='margin-bottom-none' component={ TextVariants.small }>External policy</Text>
+                        <Text className='margin-bottom-top-none' component={ TextVariants.h4 }>{ this.state.policy.name }</Text>
                     </TextContent>
-                    <div style={ { marginBottom: 'var(--pf-global--spacer--md)' } } >
+                    <div className='margin-bottom-md' >
                         { this.complianceIcon(this.state.policy.compliant) }
                         <Text component={ TextVariants.small }>
                             { this.state.policy.rules_passed } of { this.state.policy.rules_passed + this.state.policy.rules_failed } rules passed
                         </Text>
                     </div>
-                    <div style={ { marginBottom: 'var(--pf-global--spacer--md)' } } >
+                    <div className='margin-bottom-md' >
                         <Text
                             component={ TextVariants.medium }
                             onMouseEnter={ this.onMouseover.bind(this) }
                             onMouseLeave={ this.onMouseout.bind(this) }
-                            style={ { wordWrap: 'break-word' } }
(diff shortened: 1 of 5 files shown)


katierik:
Not sure if it was on the jira, but policy should be sentence case: "External policy"

Other than that, all good from me!

akofink:
Thanks @katierik! I've updated to sentence case.

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -178,15 +178,20 @@ class SystemRulesTable extends React.Component {
         cells: [{
             title: (
                 <React.Fragment key={ key }>
-                    <Stack id={ `rule-description-${key}` } gutter="md">
```
akofink: This gutter feature seems broken. It puts spacing between components, but it's the same regardless of sm,md,lg that you pass. So I just did it manually on the `StackItem` bottom margin.

packages/inventory-compliance/src/SystemPolicyCard.js
```diff
@@ -39,31 +40,31 @@ class SystemPolicyCard extends React.Component {
         return (
             <Card>
                 <CardBody>
-                    <TextContent>
-                        <Text style={ { marginBottom: '0' } } component={ TextVariants.small }>External Policy</Text>
-                        <Text style={ { marginTop: '0' } } component={ TextVariants.h4 }>{ this.state.policy.name }</Text>
+                    <TextContent style={ { marginBottom: 'var(--pf-global--spacer--md)' } } >
```
karelhala: Can you create class in stylesheet instead of direct style changes?

packages/inventory-compliance/src/SystemPolicyCard.js
```diff
@@ -39,31 +40,31 @@ class SystemPolicyCard extends React.Component {
...
+                    <div style={ { marginBottom: 'var(--pf-global--spacer--md)' } } >
+                        { this.complianceIcon(this.state.policy.compliant) }
                         <Text component={ TextVariants.small }>
-                          Last scanned: { this.state.policy.last_scanned }
+                            { this.state.policy.rules_passed } of { this.state.policy.rules_passed + this.state.policy.rules_failed } rules passed
                         </Text>
-                    </TextContent>
-                </CardFooter>
+                    </div>
+                    <div style={ { marginBottom: 'var(--pf-global--spacer--md)' } } >
```
karelhala: Same as above, perhaps moving all these hard coded styles to stylesheet?

packages/inventory-compliance/src/SystemPolicyCard.js
```diff
@@ -39,31 +40,31 @@ class SystemPolicyCard extends React.Component {
...
+                        <Text
+                            component={ TextVariants.medium }
+                            onMouseEnter={ this.onMouseover.bind(this) }
```
karelhala: You used fat arrow previous, no need to bind this.

akofink: Hmm... it doesn't seem to work without these two binds (onMouseEnter/Leave). Any idea?

karelhala: So changing it to `onMouseEnter={ this.onMouseover }` or `onMouseEnter={() => this.onMouseover() }` is not helping?

akofink: Oh, I removed the two lines completely :laughing: Yes, it works with `onMouseEnter={ this.onMouseover }`. Of course, I didn't change any of these lines in this PR, so it's kind of out of scope. But, I've added it as a separate commit.

akofink:
This should be good to go.

packages/inventory-compliance/src/SystemRulesTable.js
dLobatog: :+1:

PR #41, opened 2019-06-06
Use JS client for Vulnerability

jiridostal:
Frontend component now uses vulnerability client. 
Can be merged, dependency met https://github.com/RedHatInsights/javascript-clients/pull/21


packages/inventory-vulnerabilities/src/api/interceptors.js
```diff
@@ -0,0 +1,23 @@
+import { authInterceptor, interceptor401, responseDataInterceptor } from '@redhat-cloud-services/frontend-components-utilities/files/interceptors';
+import axios from 'axios';
+const instance = axios.create();
+
+export function errorInterceptor(err) {
+    if (!axios.isCancel(err)) {
+        const errObject = { ...err };
+
+        if (errObject.response && errObject.response.data && errObject.response.data.errors) {
```
karelhala: I don't think that this is a good idea, what if you are going to get 2 errors from server (user has no access to CVE and host is not present for instance) you will hide the second error. What do you think about using the general `errorInterceptor` from utils and we will change notifications in a way that whenever you dispatch array to notifications it will show toast notification for each dispatched item in array.

jiridostal: This definitely works for me. Currently notifications take one action only and needs to be changed to array. Processing them and showing toast for each of them would be a great feature.

PR #43, opened 2019-06-06
Adds shorthand: true to lodash webpack plugin plugin

AllenBW:
related to https://github.com/patternfly/patternfly-react/issues/1035

Need this for the victory charts used by pfcharts to be happy


AllenBW:
@karelhala @ryelo can one of yah release this after merge?

PR #47, opened 2019-06-07
Use correct API function to change display name

karelhala:
* Changing display name and ansible hostname is using new version of API and function name has changed
* When display name is changed dispatch 2 actions, reload inventory detail and reload system profile


karelhala:
Published in version [@redhat-cloud-services/frontend-components-inventory-general-info/v/0.0.7](https://www.npmjs.com/package/@redhat-cloud-services/frontend-components-inventory-general-info/v/0.0.7)

PR #51, opened 2019-06-12
Send all server errors to sentry

karelhala:
Screenshot from 2019-07-31 11-16-50

### DIFF
--- a/packages/notifications/src/Notification.js
+++ b/packages/notifications/src/Notification.js
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { Alert, Button } from '@patternfly/react-core';
+import { Alert, Button, TextContent, Text, TextVariants } from '@patternfly/react-core';
 import { CloseIcon } from '@patternfly/react-icons';
 import PropTypes from 'prop-types';
 import './notifications.scss';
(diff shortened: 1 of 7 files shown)


AllenBW:
@ryelo is this the kinda catchall we should have in each of the platform apps?

karelhala:
> @ryelo is this the kinda catchall we should have in each of the platform apps?

You should be using axios and this error interceptor so you should be covered.

AllenBW:
Hmm so maybe we ought to stop throwin em at the user here? https://github.com/RedHatInsights/insights-advisor-frontend/blob/a8f8efd65cef96eb71e8b36f6837557a2e5e9f55/src/Store/index.js#L13

karelhala:
Not really, you should still show toast notification message for user so he understands what broke, but at the same time we are sending the error to Sentry so you can see if users are getting errors and which errors.

karelhala:
@ryelo what do you think about that sentryID is it enough or should I also include sentry URL?

karelhala:
Released in new packages [@redhat-cloud-services/frontend-components-utilities/v/0.0.8](https://www.npmjs.com/package/@redhat-cloud-services/frontend-components-utilities/v/0.0.8) and [@redhat-cloud-services/frontend-components-notifications/v/0.0.7](https://www.npmjs.com/package/@redhat-cloud-services/frontend-components-notifications/v/0.0.7)

PR #57, opened 2019-06-14
Remove index.css from all packages

karelhala:
This PR removed index.css from all packages because we don't really need to track them in git.


karelhala:
Release no needed.

PR #58, opened 2019-06-17
Ensures correct sanitization of >, <, &

AllenBW:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1721265

We gotta sanitize, but also gotta accurately represent code elements... (sanitation seeeeees these a open brackets, kills em)

### See now we render it correctly!
Screen Shot 2019-06-17 at 5 38 31 PM

### DIFF
--- a/packages/inventory-insights/package.json
+++ b/packages/inventory-insights/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@redhat-cloud-services/frontend-components-inventory-insights",
-    "version": "0.1.2",
+    "version": "0.1.3",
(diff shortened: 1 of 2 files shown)


AllenBW:
There's an issue related to this over here https://github.com/swagger-api/swagger-ui/issues/3510 

tldr; sanitize-html isn't broken, rather our expectations, of _course_ an open tag should be cleaned up (in this case substituted for the text value), so the solution is to replace that text value with the tag post sanitation 🤷‍♀️

PR #64, opened 2019-06-24
Vulnerability empty states should be in UL, add links

jiridostal:
image
image


katierik:
First shot needs a colon after "This may be happening to for the following reasons:"

Also remove "The" before the link to product security group (since its not a team anymore, the grammar doesn't make sense). 

otherwise 👍

PR #68, opened 2019-07-02
Adds pick-ah-row for ansible remediation

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-1704

When no rows are highlighted, our button is disabled, if you select all rows, we only process those with playbooks (duh 😏 )

### Looks like...
Screen Shot 2019-07-02 at 8 53 47 AM
Screen Shot 2019-07-02 at 8 53 34 AM
Screen Shot 2019-07-02 at 10 06 26 AM

### DIFF
--- a/packages/inventory-insights/src/Insights.js
+++ b/packages/inventory-insights/src/Insights.js
@@ -44,7 +44,6 @@ class InventoryRuleList extends Component {
             { title: 'Risk of change', transforms: [ sortable ]},
             { title: <span className='ansibleCol'>{ ANSIBLE_ICON } Ansible</span>, transforms: [ sortable ]}
         ],
-        remediation: false,
         isKebabOpen: false,
         sortBy: {},
         activeReports: [],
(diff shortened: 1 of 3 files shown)


packages/inventory-insights/src/Insights.js
```diff
@@ -275,8 +266,40 @@ class InventoryRuleList extends Component {
...
+                oneRow.selected = isSelected;
+                return oneRow;
+            });
+        } else {
+            rows = [ ...this.state.rows ];
+            rows[rowId].selected = isSelected;
+        }
+
+        this.setState({
+            rows
```
karelhala: Spread might be better:
```JS
onSelect = (event, isSelected, rowId) => {
  this.setState({
    rows: this.state.rows.map((oneRow, rowKey) => {
        if (rowId === -1 || rowKey === rowId) {
           oneRow.selected = isSelected;
        }
        return row;
      })
  })
}
```

AllenBW: yeah! we can do this, but probs want to return `oneRow` rather than row 👍

AllenBW: we could even go further... turn it into a one liner `             rows: this.state.rows .map((oneRow, rowKey) => rowId === -1 || rowKey === rowId ? { ...oneRow, selected: isSelected } : { ...oneRow })`

PR #69, opened 2019-07-02
Fixes RHICOMPL-223: Handle errored remediations

akofink:
If the remediations API fails for whatever reason, we should handle it
as gracefully as possible in the system details rules table.

If the remediations API is down, the table will just show that no remediation is available. I think this is fine, because the remediation isn't available _at this time_, at least.


akofink:
Rebased

PR #72, opened 2019-07-08
 RHICOMPL-220 - Inconsistency in Compliance severity capitalization 

dLobatog:
Depends on https://github.com/RedHatInsights/frontend-components/pull/71 because otherwise the page would not even load because of the GraphQL changes.

### DIFF
--- a/packages/inventory-compliance/src/SystemRulesTable.js
+++ b/packages/inventory-compliance/src/SystemRulesTable.js
@@ -312,7 +312,7 @@ class SystemRulesTable extends React.Component {
     filterByPolicy = (policy, rows) => {
         const filteredRows = [];
         rows.forEach((row, i) => {
-            if (row.hasOwnProperty('isOpen') && policy.includes(row.cells[POLICY_COLUMN])) {
+            if (row.hasOwnProperty('isOpen') && policy.includes(row.cells[POLICY_COLUMN].toLowerCase())) {
                 filteredRows.push(row);
                 if (!rows[i + 1].hasOwnProperty('isOpen')) {
                     let child = rows[i + 1];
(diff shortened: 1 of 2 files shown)


packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -180,7 +180,7 @@ class SystemRulesTable extends React.Component {
         cells: [
             title,
             profile.name,
-            severity,
+            severity.charAt(0).toUpperCase() + severity.slice(1),
```
akofink: Why not just make it all lowercase? Or, even if you want it capitalized, why not apply a CSS class with `text-transform: capitalize;`? If it's just for the purposes of presentation, I think the CSS style makes more sense.

akofink:
#71 is merged. Please rebase

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -457,6 +457,7 @@ class SystemRulesTable extends React.Component {
                         />
                     </TableToolbar>
                     <Table
+                        className='compliance-rules-table'
```
dLobatog: I've changed the capitalization to happen through CSS here

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -312,7 +312,7 @@ class SystemRulesTable extends React.Component {
     filterByPolicy = (policy, rows) => {
         const filteredRows = [];
         rows.forEach((row, i) => {
-            if (row.hasOwnProperty('isOpen') && policy.includes(row.cells[POLICY_COLUMN])) {
+            if (row.hasOwnProperty('isOpen') && policy.includes(row.cells[POLICY_COLUMN].toLowerCase())) {
```
dLobatog: There was a bug here - when a report contained capitalized `severity` fields, they were not filtered properly with this. The filter was checking exclusive for lowercase: https://github.com/RedHatInsights/frontend-components/blob/master/packages/inventory-compliance/src/RulesComplianceFilter.js#L21 - so we can convert the severity here to lowercase to fix the bug.

PR #78, opened 2019-07-11
Hides reports if show_satellite_managed is false

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-1547

We need the 
* ~`show_satellite_hosts` attribute to show up on the entityDetails state~ ends up satellite_id is sufficient
* ~to GET to api/insights/v1/account_setting/ (but maybe account settings should also be on the platform state)~ maybe one day we'll pull it from the state, till then this'll work

### (updated) what it looks like when we don't show reports for satellite managed systems
Screen Shot 2019-07-15 at 11 50 29 AM

### DIFF
--- a/packages/inventory-insights/package-lock.json
+++ b/packages/inventory-insights/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@redhat-cloud-services/frontend-components-inventory-insights",
-    "version": "0.1.8",
+    "version": "0.1.10",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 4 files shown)


AllenBW:
cc @robcwilliams @katierik  you ok with the icon? or want it to be something different?

katierik:
Re: icon - there is a pficon-satellite in PF4 we can use for this, which adds a bit of character. 

https://www.patternfly.org/v4/design-guidelines/styles/icons

katierik:
Text comments: 
- Should be sentence case "Satellite managed system"

- "KCS link" is a little odd. Would it make more sense to call it a "knowledgebase article"?

PR #82, opened 2019-07-16
Empty state for compliance system details page

dLobatog:
Apparently we can currently get in an state where the host is in the compliance db but there are no reports for it. The page currently fails to load. Instead, I think something like this should be shown. Also, this is a new emptystate for the rules table whenever you select a filter that does not have any rows. 

Screenshot from 2019-07-16 10-34-06


packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -463,7 +499,7 @@ class SystemRulesTable extends React.Component {
                         onSort={ this.onSort }
                         sortBy={ sortBy }
                         onSelect={ this.onSelect }
```
karelhala: You should add `onSelect={ (currentRows.length !== 0) && this.onSelect }` so you wouldn't see that checkbox when table is empty.

dLobatog: Nice, we have had that in our backlog for a while but wasn't sure how to get rid of that, thanks!

PR #83, opened 2019-07-16
RHICOMPL-229 - Compliance Remediation available icon should match others

dLobatog:
Screenshot from 2019-07-16 15-38-40
Screenshot from 2019-07-16 15-38-15

As per https://marvelapp.com/c2bcd9h/screen/58340297


packages/inventory-compliance/src/ComplianceRemediationButton.js
```diff
@@ -10,17 +10,16 @@ class ComplianceRemediationButton extends React.Component {
...
+        id: `ssg:rhel7|${profile}|${refId}`,
         description: title,
         systems: [
             system
         ]
     })
 
-    findRule = (rules, ref_id) => {
-        return rules.find(rule => rule.ref_id === ref_id.split('|')[2]);
+    findRule = (rules, refId) => {
```
dLobatog: This is just cosmetic

packages/inventory-compliance/src/ComplianceRemediationButton.js
```diff
@@ -32,27 +31,27 @@ class ComplianceRemediationButton extends React.Component {
         });
     }
 
-    removeRefIdPrefix = (ref_id) => {
-        const splitRefId = ref_id.split('xccdf_org.ssgproject.content_profile_')[1];
+    removeRefIdPrefix = (refId) => {
+        const splitRefId = refId.toLowerCase().split('xccdf_org.ssgproject.content_profile_')[1];
```
dLobatog: Unfortunately I noticed that remediations for the C2S do not work without making the identifier lowercase, because calling remediations with `ssg:rhel7|C2S|xccdf_org.ssgproject.content_rule_ensure_gpgcheck_never_disabled` does not work. It has to be `c2s`

packages/inventory-compliance/src/ComplianceRemediationButton.js
```diff
@@ -32,27 +31,27 @@ class ComplianceRemediationButton extends React.Component {
...
         } else {
             // Sometimes the reports contain IDs like "stig-rhel7-disa" which we can pass
             // directly
-            return ref_id;
+            return refId;
         }
     }
 
-    rulesWithRemediations = (rules, system_id) => {
+    rulesWithRemediations = (rules, systemId) => {
```
dLobatog: This method changes are just cosmetic to remove the lint exception

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -76,8 +87,16 @@ class SystemRulesTable extends React.Component {
         };
 
     }
+
     removeRefIdPrefix = (refId) => {
-        return refId.split('xccdf_org.ssgproject.content_profile_')[1];
+        const splitRefId = refId.toLowerCase().split('xccdf_org.ssgproject.content_profile_')[1];
```
dLobatog: Again - we have to call `.toLowerCase`. Another problem is that we were not doing the right split here - if splitting fails, we have to return the original refId

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -202,7 +222,7 @@ class SystemRulesTable extends React.Component {
         } else {
             // One rule was selected
             const index = ((page - 1) * itemsPerPage * 2) + Number(key);
-            if (!currentRows[index].cells[4].title.props.unsupported) {
+            if (currentRows[index].cells[REMEDIATIONS_COLUMN].original) {
```
dLobatog: This would not work since we removed the `<Ansible>` component, we no longer have an `unsupported` prop

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -214,11 +234,19 @@ class SystemRulesTable extends React.Component {
     calculateParent = ({ profile }, { title, severity, compliant }) => ({
         isOpen: false,
         cells: [
-            title,
-            profile.name,
-            severity,
-            { title: (compliant ? <CheckCircleIcon className='ins-u-passed' /> :
-                <ExclamationCircleIcon className='ins-u-failed' />) }
+            { title, original: title },
```
dLobatog: Adding the `original` field here allows us to freely use the `original` to sort, filter, etc.. easily - without it you have to look inside the component, remove the icon.. it's much more of a hassle

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -315,20 +341,9 @@ class SystemRulesTable extends React.Component {
...
                 (a, b) => {
                     if (direction === SortByDirection.asc) {
-                        if (column === COMPLIANT_COLUMN) {
-                            return a.parent.cells[column].title.props.className === b.parent.cells[column].title.props.className ? 0 :
-                                a.parent.cells[column].title.props.className < b.parent.cells[column].title.props.className ? 1 : -1;
-                        } else {
-                            return a.parent.cells[column].localeCompare(b.parent.cells[column]);
-                        }
-
+                        return String(a.parent.cells[column].original).localeCompare(String(b.parent.cells[column].original));
```
dLobatog: Here's an example of code cleaner by using the `original` version of each row. Before we were checking the props of the compliant column

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -270,13 +298,11 @@ class SystemRulesTable extends React.Component {
...
         const { currentRows, profiles, refIds } = this.state;
-        return currentRows.filter(row => row.selected && !row.cells[4].title.props.unsupported).map(row => ({
+        return currentRows.filter(row => row.selected && !row.cells[REMEDIATIONS_COLUMN].title.props.unsupported).map(row => ({
             // We want to match this response with a similar response from GraphQL
-            /* eslint-disable camelcase */
-            ref_id: refIds[row.cells[0]],
-            profiles: [{ ref_id: profiles[row.cells[0]] }],
-            /* eslint-enable camelcase */
-            title: row.cells[0] // This is the rule title, the description is too long
+            refId: refIds[row.cells[0].original],
```
dLobatog: Cosmetic

PR #94, opened 2019-07-25
Add UI for rule references/identifiers

akofink:
image

If no references:

image

With some references that don't have an href:

image

@katierik please take a look as well. Do you think we should hide labels if no data exists? @dLobatog should we display references is any particular order?

### DIFF
--- a/packages/inventory-compliance/src/Compliance.js
+++ b/packages/inventory-compliance/src/Compliance.js
@@ -46,9 +46,17 @@ query System($systemId: String!){
                 refId
                 description
                 compliant(systemId: $systemId)
+                identifier {
+                    label
+                    system
+                }
+                references {
+                    label
+                    href
+                }
             }
         }
-	}
+    }
 }
 `;
 
--- a/packages/inventory-compliance/src/SystemRulesTable.js
+++ b/packages/inventory-compliance/src/SystemRulesTable.js
@@ -14,6 +14,8 @@ import {
     EmptyStateVariant,
     Level,
     LevelItem,
+    Grid,
+    GridItem,
     Stack,
     StackItem,
     Pagination,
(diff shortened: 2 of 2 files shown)


akofink:
I just realized this doesn't quite match the mock (adding identifiers inline in each row). I'll be updating this PR to add that.

katierik:
If there are no references - Hide
If there are multiple "TBD" -> can we consolidate this down to "TBD (3 references)" or just a single "TBD"

What is the identifier linking to right now? It isn't a link in the mock but if there is info to be linked to...

akofink:
@dLobatog I messed with the data manually to get the three cases (with/without href). There are multiple TBD references on this rule as displayed; again, I did this manually in the DB, not from the parsed report. I'm unsure if multiple of the same reference can be on a rule, but we could enforce uniqueness in the model. The NaN comes from the XML (i.e.):
```xml
<reference href="http://iase.disa.mil/stigs/cci/Pages/index.aspx">NaN</reference>
```

About hiding, I've asked @katierik above. I'm fine with hiding the label if UX approves that behavior.

@dLobatog What do you think about the order in which we display references? Should I sort by label?

dLobatog:
@akofink I didn't realize but yeah, if you can sort by label that sounds easier to read.

akofink:
Yes, some identifiers have links, and others do not. This is the same for references. The screenshots are contrived examples; I don't think we should see multiple of the same reference on a rule, so we shouldn't need to handle that case.

Identifiers look like this in the xml:

```xml
<ident system="https://nvd.nist.gov/cce/index.cfm">CCE-27337-5</ident>
```

I'll work on hiding the label and sorting the references.

akofink:
Updated to hide the references label unless there are references to show.

https://github.com/RedHatInsights/compliance-backend/pull/190 solves sorting. I don't see a reason why we should make the front end do that when we will likely always want to sort by label.

akofink:
I also found some realistic screenshots (not contrived) and updated the description.

katierik:
Looks good!

PR #95, opened 2019-07-25
Update Summary screen according to design

rvsia:
**Description**

Updates summary screen according to design

- the summary is generated from values, labels are taken from API schema
- boolean values are converted into Yes / No
- exceptions (types and name) are handled separately

**Design**

image

**After**

image

image

**edit**: added Review title

image

@terezanovotna

I have a few questions:

1. Should we show empty fields in the summary?
2. Is okay to converts boolean (checkbox) to Yes, No?
3. What about the styling? I wasn't able to find the correct PF4 component to match your design.
4. What about ordering? Right now, because of the algorithm which gets the values, the first field in the summary is the first field which was changed by a user (so, name can be first or second..) Should I implement some hardcoded order?

### DIFF
--- a/packages/sources/src/addSourceWizard/SourceAddSchema.js
+++ b/packages/sources/src/addSourceWizard/SourceAddSchema.js
@@ -296,20 +296,22 @@ const sourceTypeSteps = sourceTypes =>
     sourceTypes.map(t => fieldsToSteps(sourceTypeSchema(t), t.name, 'summary'))
     .flatMap((x) => x);
 
-const summaryStep = () => ({
+const summaryStep = (sourceTypes) => ({
     fields: [
         {
             component: 'description',
             name: 'description-summary',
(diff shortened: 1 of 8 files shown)


terezanovotna:
This looks awesome!

Answering your questions

1. yes, we want to include either n/a or - or not available. We want to notify the user the info is missing.
2. yes
3. looks great! are we using the right font?
4. from what I see, this looks great. the we are seeing the right ordering. 

other notes:

- the title on the left should match the title on the right. for ex: "Review source details" should same in both left and right.


NICE JOB!

rvsia:
@terezanovotna Updated
* Source name is always first, type is second
* Header is matching title in the step mapper

Thank you for your feedback! :+1: :cake: 

image


@Hyperkid123 please, review! :pray:

rvsia:
@martinpovolny

PR #97, opened 2019-07-26
RHICOMPL-275 - Allow to select all rules

dLobatog:
The previous method does not work since we don't have the `<Ansible>` component with `title.props.unsupported`. Right now we can just check the `original` value - which tells us whether the remediation is available or not with `true` or `false`

Peek 2019-07-30 09-26


dLobatog:
@karelhala It's not really correct from the UI standpoint, because only fixes for rules with `Ansible remediation` are really submitted once you click, but the resulting playbook is correct.

PR #101, opened 2019-07-29
Add rule identifier in each table row in system details

akofink:
image

### DIFF
--- a/packages/inventory-compliance/src/SystemRulesTable.js
+++ b/packages/inventory-compliance/src/SystemRulesTable.js
@@ -21,6 +21,7 @@ import {
     Pagination,
     PaginationVariant,
     Text,
+    TextContent,
     TextVariants,
     Title
 } from '@patternfly/react-core';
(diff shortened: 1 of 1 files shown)


packages/inventory-compliance/src/compliance.scss
```diff
@@ -63,6 +63,12 @@
...
 
-.compliance-rules-table td[data-label=Severity] {
-    text-transform: capitalize;
+.compliance-rules-table {
+    td[data-label=Severity] {
+        text-transform: capitalize;
+    }
+
+    td[data-label=Rule] small {
+        color: #72767b; // var(--pf-c-content--small--Color)
```
akofink: I have no clue why this variable isn't working. It seems like insights chrome is overriding the value to be `rgb(26, 26, 26)`

karelhala: I can't find any variable `--pf-c-content--small--Color` in PF core, perhaps you should use `var(--pf-global--Color--dark-200)` instead. Or you should wrap `Text` with `TextContent` component
```JSX
ruleTitleCell = (title, identifier) => <Stack>
        <StackItem>{ title }</StackItem>
        { identifier && <StackItem><TextContent><Text component={ TextVariants.small }>{ identifier.label }</Text></TextContent></StackItem> || '' }
    </Stack>
```

akofink: Awesome, thanks!

PR #102, opened 2019-07-29
Turn on source-map devtool

akofink:


akofink:
My intent is to enable source maps for development, where the source is locally available (via `npm link`). I don't really see a need to ship the source maps in the package. Do npms usually contain the maps?

akofink:
Good to merge this one?

PR #109, opened 2019-08-01
Add CardSelect

rvsia:
**Description**

- Adds CardSelect component to schema
- allows multi select
- replaces Select in Source Type selection

**Docs**

## CardSelect

Use in schema as `component: 'card-select`

**Props**

This components accepts all formGroup props `(label, helperText, isDisabled, isRequired, ...)`

|Prop|Type|Description|
|----|:--:|----------:|
|options|array|Array of options with keys `value`, `label`|
|DefaultIcon|node, element, func|Default icon (default is `ServerIcon`)|
|iconMapper*|func|You can use your own mapper `(value, DefaultIcon) => ...` |
|multi/isMulti|bool|Allows to select more items|

* In the future it could be replaced by data obtained from API

**Design**

image

**After**

image

Todo:
- [x] OnChange
- [x] build styles (need help by @karelhala )
- [x] UX review
- [x] Tests
- [x] Docs 
- [x] expose outside

**More GIFs!** :tada: :champagne: :chart_with_upwards_trend: :money_mouth_face: 

cardselect

@terezanovotna

Please, tell me what you think about it! I have made cards bigger: the labels are too long, they wrap into second line and it looks _meh_ And with this width I can easily use a grid system (span 4). 

Icons are manually mapped and used by PatternFly Icons, however, this solution is not future-proof. (New source types = new icons.. but with only three sources it works now.)

### DIFF
--- a/packages/sources/src/sourceFormRenderer/components/CardSelect.js
+++ b/packages/sources/src/sourceFormRenderer/components/CardSelect.js
@@ -1,0 +1,138 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Card, CardHeader, CardBody, FormGroup, Grid, GridItem, Bullseye } from '@patternfly/react-core';
+import { ServerIcon } from '@patternfly/react-icons';
+
+class CardSelect extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            selectedValues: this.props.input.value ? this.props.input.value : this.isMulti ? [] : undefined
+        };
+    }
+
+    isMulti = this.props.isMulti || this.props.multi;
+
+    isDisabled = this.props.isDisabled || this.props.isReadOnly;
+
+    change = () => this.props.input.onChange(this.state.selectedValues);
+
+    handleMulti = (value) => this.state.selectedValues.includes(value) ?
+        this.setState(({ selectedValues }) => ({ selectedValues: selectedValues.filter(valueSelect => valueSelect !== value) }), this.change)
+        : this.setState(({ selectedValues }) => ({ selectedValues: [ ...selectedValues, value ]}), this.change);
+
+    handleSingle = (value) => this.setState(() => ({ selectedValues: this.state.selectedValues === value ? undefined : value }), this.change);
+
+    handleClick = (value) => {this.isMulti ? this.handleMulti(value) : this.handleSingle(value);}
+
+    handleCompare = (value) => this.isMulti ? this.state.selectedValues.includes(value) : this.state.selectedValues === value;
+
+    onClick = value => {
+        if (this.isDisabled) {
+            return undefined;
+        }
+
+        this.handleClick(value);
+        this.props.input.onBlur();
+    }
+
+    handleKeyPress = (event, value) => {
+        if (event.key === 'Enter') {
+            this.onClick(value);
+        }
+    }
+
+    prepareCards = () => this.props.options.map(({ value, label }) => {
+        const { iconMapper, DefaultIcon } = this.props;
+
+        if (!value) {
+            return undefined;
+        }
+
+        const Component = iconMapper(value, DefaultIcon);
+
+        return (
+            <GridItem sm={ 1 } md={ 4 } key={ value }>
+                <Card
+                    className={ `ins-c-sources__wizard--card${this.handleCompare(value) ? ' selected' : ''}${this.isDisabled ? ' disabled' : ''}` }
+                    onClick={ () => this.onClick(value) }
+                    tabIndex={ this.isDisabled ? -1 : 0 }
+                    onKeyPress={ (e) => this.handleKeyPress(e, value) }
+                    isHoverable={ !this.isDisabled }
+                    isCompact={ true }
+                >
+                    <div className={ this.isDisabled ? 'disabled' : '' }>
+                        <CardHeader>
+                            { label }
+                        </CardHeader>
+                        <CardBody>
+                            <Bullseye>
+                                <Component size="xl"/>
+                            </Bullseye>
+                        </CardBody>
+                    </div>
+                </Card>
+            </GridItem>
+        );
+    })
+
+    render() {
+        const { FieldProvider, isRequired, label, helperText, hideLabel, name, meta, input, ...rest } = this.props;
+        const { error, touched } = meta;
+        const showError = touched && error;
+
+        return (
+            <FormGroup
+                isRequired={ isRequired }
+                label={ !hideLabel && label }
+                fieldId={ input.name }
+                isValid={ !showError }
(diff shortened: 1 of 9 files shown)


rvsia:
@terezanovotna What about signaling error?

image

And this is how it looks on smaller screen

image

I also added an ability to select by TABing and pressing ENTER. Should I add moving across the cards by pressing Left, Right?

rvsia:
**Final set of screenshots and gifs** :money_mouth_face: :chart_with_upwards_trend: 

Normal

image

Hover

finalhover

Keyboard selection (TAB >, SHIFT+TAB < )

finalkeyboard

Selected

image

packages/sources/src/sourceFormRenderer/components/CardSelect.js
```diff
@@ -0,0 +1,134 @@
...
+        };
+    }
+
+    isMulti = this.props.isMulti || this.props.multi;
+
+    isDisabled = this.props.isDisabled || this.props.isReadOnly;
+
+    change = () => this.props.input.onChange(this.state.selectedValues);
+
+    handleMulti = (value) => this.state.selectedValues.includes(value) ?
```
Hyperkid123: I believe that selecting multiple sources was deprecated.

packages/sources/src/sourceFormRenderer/components/CardSelect.js
```diff
@@ -0,0 +1,134 @@
...
+        : this.setState(({ selectedValues }) => ({ selectedValues: [ ...selectedValues, value ]}), this.change);
+
+    handleSingle = (value) => this.setState(() => ({ selectedValues: this.state.selectedValues === value ? undefined : value }), this.change);
+
+    handleClick = (value) => {this.isMulti ? this.handleMulti(value) : this.handleSingle(value);}
+
+    handleCompare = (value) => this.isMulti ? this.state.selectedValues.includes(value) : this.state.selectedValues === value;
+
+    onClick = value => {
+        if (this.isDisabled) {
```
Hyperkid123: Pointer events should not be even allowed when in disabled state.

packages/sources/src/sourceFormRenderer/components/CardSelect.js
```diff
@@ -0,0 +1,134 @@
...
+    handleKeyPress = (event, value) => {
+        if (event.key === 'Enter') {
+            this.onClick(value);
+        }
+    }
+
+    prepareCards = () => this.props.options.map(({ value, label }) => {
+        const { iconMapper, DefaultIcon } = this.props;
+
+        if (!value) {
```
Hyperkid123: This condition should not be necessary, users should follow defined API. I would not be against actually throwing an error.

packages/sources/src/sourceFormRenderer/components/CardSelect.js
rvsia: It is for now. However, I am preparing for the future. :space_invader:

packages/sources/src/sourceFormRenderer/components/CardSelect.js
```diff
@@ -0,0 +1,134 @@
...
+            </GridItem>
+        );
+    })
+
+    render() {
+        const { FieldProvider, isRequired, label, helperText, hideLabel, name, meta, input, ...rest } = this.props;
+        const { error, touched } = meta;
+        const showError = touched && error;
+
+        return (<FormGroup
```
Hyperkid123: You should have new line after and before `(`, `)`

packages/sources/src/sourceFormRenderer/components/CardSelect.js
```diff
@@ -0,0 +1,134 @@
...
+};
+
+CardSelect.defaultProps = {
+    options: [],
+    DefaultIcon: ServerIcon,
+    iconMapper: (_value, DefaultIcon) => DefaultIcon
+};
+
+const CardSelectProvider = ({ FieldProvider, ...rest }) =>
+    (<FieldProvider { ...rest }>
```
Hyperkid123: new lines :wink:

packages/sources/src/sourceFormRenderer/components/CardSelect.js
rvsia: It was made to add a option to easily share the options prop between Select (which supports undefined values) and this component.

packages/sources/src/styles/cardSelect.scss
```diff
@@ -0,0 +1,28 @@
...
+      box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.42), 0 4px 11px 0 rgba(0, 0, 0, 0.42);
+      outline: none;
+    }
+
+    &.selected {
+      border:1px solid;
+      border-color: var(--pf-global--primary-color--100);
+    }
+  
+    &.disabled {
```
Hyperkid123: You can disable pointer events here.

rvsia: Will look on it! :microscope:

packages/sources/src/tests/__mocks__/FieldProvider.js
```diff
@@ -0,0 +1,24 @@
+import { createElement } from 'react';
```
Hyperkid123: I am wondering if we should expose some testUtils in data driven forms.

rvsia: I was thinking about some test framework

Hyperkid123: stuff like FieldProvider, some simple layout mapper etc. Let me think about it. We could also use it for testing our mappers etc.

packages/sources/src/sourceFormRenderer/components/CardSelect.js
Hyperkid123: that makes sense :hurtrealbad:

packages/sources/src/sourceFormRenderer/components/CardSelect.js
rvsia: :+1: I disabled it in CSS, i think here it could serve as an another check.

packages/sources/src/styles/cardSelect.scss
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    &.selected {
+      border:1px solid;
+      border-color: var(--pf-global--primary-color--100);
+    }
+
+    &.disabled {
+      cursor: not-allowed;
+      background-color: rgba(210, 210, 210, .5);
```
karelhala: Can you use PF's var in here? `var(--pf-global--BorderColor--100)`

rvsia: fixed :+1:

PR #112, opened 2019-08-05
Compliance: Support filters without search

dLobatog:


packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -439,9 +439,9 @@ class SystemRulesTable extends React.Component {
             result = policyRows;
         }
 
-        if (title.length > 0 && result.length > 0) {
+        if (title && title.length > 0 && result.length > 0) {
```
dLobatog: If you currently click on a filter without writing some search, it will fail with `title is undefined` when it gets to this point.

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -457,7 +457,7 @@ class SystemRulesTable extends React.Component {
             passedRows = originalRows;
         }
 
-        const severityRows = this.filterBy(severity, originalRows, SEVERITY_COLUMN);
+        const severityRows = this.filterBy(severity.join('|'), originalRows, SEVERITY_COLUMN);
```
dLobatog: Multiple severity filters were not handled properly - with this, we call `.match` on `medium|high` for example, which works

PR #117, opened 2019-08-08
Add tests to config generation

karelhala:


ryelo:
@karelhala Lint and then it's good!

PR #119, opened 2019-08-09
Initial inventory-compliance tests

dLobatog:
This puts us above 50% test coverage:

```
File                                         |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
 inventory-compliance/src                    |    58.14 |    52.78 |    49.54 |     58.4 |                   |
  Compliance.js                              |        0 |        0 |        0 |        0 |... 44,154,167,177 |
  ComplianceRemediationButton.js             |        5 |        0 |        0 |     5.41 |... 96,115,116,117 |
  Constants.js                               |      100 |      100 |      100 |      100 |                   |
  RulesComplianceFilter.js                   |      100 |    85.71 |      100 |      100 |             53,65 |
  SystemPolicyCard.js                        |    71.43 |       50 |       60 |    71.43 |             32,36 |
  SystemPolicyCards.js                       |    28.57 |        0 |        0 |    28.57 |     9,11,18,20,25 |
  SystemRulesTable.js                        |    69.28 |       63 |    67.74 |    70.29 |... 39,441,451,538 |

```

cc @akofink @quarckster

### DIFF
--- a/packages/inventory-compliance/package.json
+++ b/packages/inventory-compliance/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@redhat-cloud-services/frontend-components-inventory-compliance",
-    "version": "0.0.14",
+    "version": "0.0.22",
     "description": "Detail part of compliance for RedHat Cloud Services.",
     "main": "index.js",
     "publishConfig": {
--- a/packages/inventory-compliance/src/RulesComplianceFilter.test.js
+++ b/packages/inventory-compliance/src/RulesComplianceFilter.test.js
@@ -1,0 +1,55 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import toJson from 'enzyme-to-json';
+import RulesComplianceFilter from './RulesComplianceFilter';
+import { IntlProvider } from 'react-intl';
+
+describe('RulesComplianceFilter component', () => {
+    const policies = [
+        {
+            refId: 'xccdf_org.ssgproject.content_profile_pci-dss',
+            name: 'PCI-DSS v3 Control Baseline for Red Hat Enterprise Linux 7'
+        },
+        {
+            refId: 'xccdf_org.ssgproject.content_profile_pci-dss',
+            name: 'PCI-DSS v3 Control Baseline for Red Hat Enterprise Linux 8'
+        }
+    ];
+
+    it('should render filter', () => {
+        const wrapper = shallow(
+            <RulesComplianceFilter availablePolicies={ policies } />
+        );
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+
+    it('should not allow to filter if there is only one policy', () => {
+        const wrapper = shallow(
+            <RulesComplianceFilter availablePolicies={ [ policies[0] ] } />
+        );
+        const instance = wrapper.instance();
+        expect(wrapper.state('filterCategories').length).toEqual(2);
+    });
+
+    it('should allow to add and remove filters', () => {
+        const wrapper = shallow(
+            <RulesComplianceFilter availablePolicies={ policies } />
+        );
+        const instance = wrapper.instance();
+        instance.addFilter('hidePassed', true);
+        expect(wrapper.state('hidePassed')).toEqual(true);
+        instance.addFilter('severity', 'medium');
+        instance.addFilter('severity', 'low');
+        expect(wrapper.state('hidePassed')).toEqual(true);
+        expect(wrapper.state('severity')).toEqual([ 'medium', 'low' ]);
+        instance.removeFilter('severity', 'low');
+        expect(wrapper.state('severity')).toEqual([ 'medium' ]);
+        instance.addFilter('policy', policies[0].name);
+        instance.addFilter('policy', policies[1].name);
+        expect(wrapper.state('policy')).toEqual([ policies[0].name, policies[1].name ]);
+        instance.removeFilter('policy', policies[0].name);
+        expect(wrapper.state('policy')).toEqual([ policies[1].name ]);
+        instance.removeFilter('hidePassed', false);
+        expect(wrapper.state('hidePassed')).toEqual(false);
+    });
+});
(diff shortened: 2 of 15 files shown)


akofink:
Don't you need to re-record the snapshots for RulesComplianceFilter and SystemRulesTable?

karelhala:
Nice work, you have to run `npm run test -- -u` to update snapshots so they won't fail in travis.

packages/inventory-compliance/src/RulesComplianceFilter.test.js
```diff
@@ -0,0 +1,55 @@
...
+        expect(wrapper.state('filterCategories').length).toEqual(2);
+    });
+
+    it('should allow to add and remove filters', () => {
+        const wrapper = shallow(
+            <RulesComplianceFilter availablePolicies={ policies } />
+        );
+        const instance = wrapper.instance();
+        instance.addFilter('hidePassed', true);
+        expect(wrapper.state('hidePassed')).toEqual(true);
```
karelhala: Can we limit the amount of expects per test to 2 or 3? This way if one of these fails the entire test case fails and you'd have hard time debugging what actually failed out of these 7 expects.

packages/inventory-compliance/src/SystemRulesTable.test.js
```diff
@@ -0,0 +1,157 @@
...
+import { shallow, mount } from 'enzyme';
+import toJson from 'enzyme-to-json';
+import SystemRulesTable from './SystemRulesTable';
+import { IntlProvider } from 'react-intl';
+jest.mock('@redhat-cloud-services/frontend-components-remediations/remediationsApi');
+import * as remediationsApi from '@redhat-cloud-services/frontend-components-remediations/remediationsApi';
+import { SortByDirection } from '@patternfly/react-table';
+import { TITLE_COLUMN } from './Constants';
+
+const windowInsightsSpy = {
```
karelhala: Can you move this function to `/config/setUpTests`? So we don't have to mock it everytime?

packages/inventory-compliance/src/SystemRulesTable.test.js
```diff
@@ -0,0 +1,157 @@
...
+};
+
+const system = {
+    id: 'aa9c4497-5707-4233-9e9b-1fded5423ef3',
+    name: '3.example.com',
+    profiles: {},
+    __typename: 'System'
+};
+
+const profileRules = [
```
karelhala: Perhaps moving this to a JS with export file under `data/` folder might be better. So the test files are just for testing and we can change data much easily.

packages/inventory-compliance/src/SystemRulesTable.test.js
```diff
@@ -0,0 +1,157 @@
...
+                __typename: 'Rule'
+            }
+        ]
+    }
+];
+
+describe('SystemRulesTable component', () => {
+    it('should render', () => {
+        global.insights = windowInsightsSpy;
+        remediationsApi.getResolutionsBatch.mockImplementation(
```
karelhala: This yields a HTTP get right? If so, use [jest-fetch-mock](https://www.npmjs.com/package/jest-fetch-mock) instead of mocking promises and mocking implementation.

packages/inventory-compliance/src/SystemRulesTable.test.js
```diff
@@ -0,0 +1,157 @@
...
+        );
+        const wrapper = shallow(
+            <SystemRulesTable
+                profileRules={ profileRules }
+                loading={ false }
+                system={ system }
+            />
+        );
+        const instance = wrapper.instance();
+        return instance.setInitialCurrentRows().then(() => {
```
karelhala: Please rewrite this to more streamlined approach. If one of these requests fails the entire test fails and we won't be able to find out which actually failed.

packages/inventory-compliance/src/SystemRulesTable.test.js
```diff
@@ -0,0 +1,157 @@
...
+            expect(wrapper.state('currentRows').length / 2).toEqual(2);
```
karelhala: Same as above, refactor this to more streamlined approach.

dLobatog: Not sure if this is possible here, the nesting is minimal here - just setting the rows and then sorting.

packages/inventory-compliance/src/SystemRulesTable.test.js
dLobatog: :+1: - that can be done

packages/inventory-compliance/src/SystemRulesTable.test.js
dLobatog: Sounds good

packages/inventory-compliance/src/SystemRulesTable.test.js
dLobatog: I tried by creating a file `config/setupJest.js`, and then changing package.lock with: 
```
    "jest": {
        "automock": false,
        "setupFiles": [
            "./setupJest.js"
        ]
    },
```

But it didn't quite work because we run the tests from the root of the monorepo, any ideas? So far I set `global.insights = global.window.insights;` in the global `config/setupTests.js` and that seems to be enough.

karelhala: Actually, the chrome config has been moved to global config so no need to define it here. And you can start using `window.insights` or as you mentioned setting `global.insights = global.window.insights;` should be enough.

packages/inventory-compliance/package.json
```diff
@@ -32,6 +32,12 @@
         "react-content-loader": "^3.4.1",
         "lodash": "4.17.15"
     },
+    "jest": {
```
karelhala: You don't really need to set jest for monorepo in package, instead if you want to run only compliance tests you can run `npm run test packages/inventory-compliance` from root folder.

PR #127, opened 2019-08-22
Allow search by references and identifiers in UI

akofink:
It should work exactly as before and play nice with other filters, but it now also searches references and identifiers!

image

### DIFF
--- a/packages/inventory-compliance/src/SystemRulesTable.js
+++ b/packages/inventory-compliance/src/SystemRulesTable.js
@@ -272,6 +272,8 @@ class SystemRulesTable extends React.Component {
     calculateChild = ({ description, rationale, identifier, references }, key) => ({
         parent: key * 2,
         cells: [{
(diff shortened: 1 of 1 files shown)


packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -272,6 +272,8 @@ class SystemRulesTable extends React.Component {
     calculateChild = ({ description, rationale, identifier, references }, key) => ({
         parent: key * 2,
         cells: [{
+            originalIdentifier: identifier ? identifier.label : '',
+            originalReferences: references && references.length ? references.map((r) => r.label).join() : '',
```
akofink: This is the same paradigm as is used for the parent row cells.

akofink:
@bastilian I was able to do this without any refactoring; however, we really should tackle some of this tech debt and try simplifying this component.

PR #146, opened 2019-09-12
Data toolbar

karelhala:
### Large screen
Screenshot from 2019-09-12 17-35-04

### Large screen - actions
Screenshot from 2019-09-12 17-38-58

### Large screen - bulk select
Screenshot from 2019-09-12 17-39-03

### Smaller screen
Screenshot from 2019-09-12 17-36-54

### Smaller screen - actions
Screenshot from 2019-09-12 17-37-53

### Smaller screen - bulk select
Screenshot from 2019-09-12 17-38-05

* waiting on https://github.com/RedHatInsights/frontend-components/pull/138


katierik:
🎉 This is looking awesome
Some feedback:
1 - Since the original design, PF addressed the problem we had with the pagination (aka, it is HUGE) and released a compact variant. We should use this one: https://github.com/patternfly/patternfly-next/issues/2110

2 - Tiny thing: Sentence case! It should be lower case "s" on selected e.g. "10 selected"

karelhala:
@katierik thanks for the review!

Updated look with filter chips:
Screenshot from 2019-09-26 15-53-54
Screenshot from 2019-09-26 15-55-08

karelhala:
We should work on styling of export and count on mobile, but I think that is looks OK for now.

packages/components/src/Components/FilterChips/filter-chips.scss
```diff
@@ -0,0 +1,9 @@
+.ins-c-chip-group__plain > li {
+    background: none;
+    padding: 0;
+    background-color: initial !important;
```
ryelo: Can we make this parent more specific to eliminate !important

packages/components/src/Components/FilterChips/filter-chips.scss
```diff
@@ -0,0 +1,9 @@
...
+
+    h4 {
+        display: none !important;
```
ryelo: same as above

packages/components/src/Components/PrimaryToolbar/Actions.js
```diff
@@ -0,0 +1,127 @@
...
+                                    { firstAction.label }
+                                </Button> :
+                                firstAction
+                        }
+                    </DataToolbarItem>
+                }
+                {
+                    ((actions && actions.length > 1) || (overflowActions.length > 0)) &&
+                    <DataToolbarItem
+                        className={ `${actions.length <= 1 && 'ins-m-empty-actions'}` }
```
ryelo: nitpick: the modifier comes last, so this should be `ins-m-actions--empty`

packages/components/src/Components/FilterChips/filter-chips.scss
karelhala: Good point. I was frustrated that PF rewrite these styles, well I went with more specific rule and it works, thanks!

ryelo:
@karelhala fix Travis and code looks good!

karelhala:
Released in new version [@redhat-cloud-services/frontend-components/v/0.0.20](https://www.npmjs.com/package/@redhat-cloud-services/frontend-components/v/0.0.20)

PR #147, opened 2019-09-13
UI enhancements of AddSourceWizard

rvsia:
**Description**

Enhancements:
- Options (sourcetypes, apptypes) are sorted by alphabet
- Adds Microsoft Icon for Azure
- CardSelect allow to disable an option
- Add URL validator for Openshift
- Adds Async validation for name field

### DIFF
--- a/packages/sources/src/addSourceWizard/SourceAddSchema.js
+++ b/packages/sources/src/addSourceWizard/SourceAddSchema.js
@@ -4,6 +4,22 @@ import { Popover, TextContent, TextList, TextListItem, Text, TextVariants, Title
 import { QuestionCircleIcon } from '@patternfly/react-icons';
 import SSLFormLabel from './SSLFormLabel';
 import { AwsIcon, OpenshiftIcon, MicrosoftIcon } from '@patternfly/react-icons';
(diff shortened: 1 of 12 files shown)


packages/sources/src/addSourceWizard/SourceAddSchema.js
```diff
@@ -3,11 +3,27 @@ import { componentTypes, validatorTypes } from '@data-driven-forms/react-form-re
...
+export const asyncValidator = (value, sourceId = undefined) => findSource(value).then(({ data: { sources }}) => {
+    if (sources.find(({ id }) => id !== sourceId)) {
+        return 'Name has already been taken';
+    }
+
+    if (value === '' || value === undefined) {
+        return 'Name can\'t be blank';
+    }
+
+    return undefined;
```
karelhala: Undefined is returned by default. So no need for it here.

rvsia: I think this is more readable. Like "it is our intention to return undefined and it's not a mistake."

rvsia:
@karelhala Yeah, please merge it but you don't need to release it. I am preparing another PR, so there is no need to release this yet.

PR #148, opened 2019-09-16
Update pf-icon version to support ansible icon in inv-insights table

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-2340

### Looks like
Screen Shot 2019-09-16 at 10 17 41 AM
Screen Shot 2019-09-16 at 10 17 30 AM


AllenBW:
Travis fails aren't related to this pr 😭

packages/inventory-insights/src/Insights.js
```diff
@@ -331,7 +333,9 @@ class InventoryRuleList extends Component {
                             <RemediationButton
                                 isDisabled={ this.getSelectedItems(rows).length === 0 }
                                 dataProvider={ this.processRemediation }
-                                onRemediationCreated={ result => this.props.addNotification(result.getNotification()) } />
+                                onRemediationCreated={ result => this.props.addNotification(result.getNotification()) } >
+                                <AnsibeTowerIcon size='sm' color={ global_BackgroundColor_100.value } /> Remediation
```
karelhala: Can you make this into a new style in `.scss` file? That way we don't have to import this const from pf styles.

AllenBW: So this const comes from the `reac-tokens`, and we're already using it in another place...  🤔 `global_success_color_200` yah want me to move both of those into the scss?

packages/inventory-insights/src/Insights.js
```diff
@@ -41,7 +41,9 @@ class InventoryRuleList extends Component {
             { title: 'Added', transforms: [ sortable, cellWidth(15) ]},
             { title: 'Total risk', transforms: [ sortable ]},
             { title: 'Risk of change', transforms: [ sortable ]},
-            { title: <span className='ansibleCol'>{ ANSIBLE_ICON } Ansible</span>, transforms: [ sortable ]}
+            { title: <span className='ansibleCol'>
+                <AnsibeTowerIcon size='md' /> Ansible
```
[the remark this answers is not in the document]
AllenBW: Can confirm it exists in this version (see ss), are yah talkin bout a programmatic check to employing it here?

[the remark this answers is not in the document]
karelhala: Yup, exactly. Just check `{AnsibleTowerIcon && <AnsibleatoweIcon...`

packages/inventory-insights/src/Insights.js
```diff
@@ -19,17 +19,17 @@ import {
     ToolbarGroup,
     ToolbarItem
 } from '@patternfly/react-core';
-import { CheckIcon, ExternalLinkAltIcon, SearchIcon, TimesCircleIcon, PficonSatelliteIcon } from '@patternfly/react-icons';
+import { CheckIcon, ExternalLinkAltIcon, SearchIcon, TimesCircleIcon, PficonSatelliteIcon, AnsibeTowerIcon } from '@patternfly/react-icons';
 import { cellWidth, sortable, SortByDirection, Table, TableBody, TableHeader } from '@patternfly/react-table';
 import { flatten, sortBy } from 'lodash';
-import { global_success_color_200 } from '@patternfly/react-tokens';
+import { global_success_color_200, global_BackgroundColor_100 } from '@patternfly/react-tokens';
```
karelhala: We should create styles for this, but that is not in scop of this PR so good to merge. Just to remember for future.

PR #150, opened 2019-09-17
Handle rules without identifiers

akofink:
While this shouldn't ever happen, we have no checks to ensure a rule has
an identifier, so we should handle the case at least in the frontend.


akofink:
Rebased

karelhala:
@akofink you gotta release a new version, do you know how to do it?

akofink:
I'm working on figuring it out :)

PR #153, opened 2019-09-20
Allow sorting by remediation available

dLobatog:
This removes the need to call for `remediationAvailable` as we get that attribute directly from GraphQL now, which improves the speed of the page by a good margin. Additionally, this lets us filter and sort by remediation available, which is a nice feature.


packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -133,15 +132,14 @@ class SystemRulesTable extends React.Component {
...
-        this.currentRows(page, itemsPerPage).then((currentRows) => {
-            this.setState(() => (
-                {
-                    page,
-                    itemsPerPage,
-                    currentRows
-                }
-            ));
-        });
+        const currentRows = this.currentRows(page, itemsPerPage);
```
karelhala: No need to assign to `currentRows`, you can directly set this in state: `currentRows: this.currentRows(page, itemsPerPage)`

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -387,16 +374,16 @@ class SystemRulesTable extends React.Component {
...
-            this.setState(() => ({
-                currentRows,
-                sortBy: {
-                    index,
-                    direction
-                },
-                rows: sortedRows
-            }));
-        });
+        const currentRows = this.currentRows(page, itemsPerPage, { rows: sortedRows, profiles, refIds });
```
karelhala: Same as above, no need to assign to variable.

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -484,23 +471,22 @@ class SystemRulesTable extends React.Component {
...
-                hidePassed,
-                severity,
-                policy,
-                searchTerm
-            }));
-        });
+        );
+
+        return this.setState(() => ({
+            currentRows,
```
karelhala: And here as well.

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -253,6 +231,11 @@ class SystemRulesTable extends React.Component {
                 title: (compliant ? <CheckCircleIcon className='ins-u-passed' /> :
                     <ExclamationCircleIcon className='ins-u-failed' />),
                 original: compliant
+            },
+            {
+                title: (remediationAvailable ? <CheckIcon style={ { color: 'var(--pf-global--success-color--200)' } }/> :
```
karelhala: Can you make this into a style in `.scss` file? Inline styles are not really good, you can add `ins-c-compliance-system-rule-check` className to this component and style it in your stylesheet file.

PR #155, opened 2019-09-23
Add a tag count component

ryandillinfelton:
This PR adds a new component that shows the number of tags associated with a given system. Currently planned for use on the 'system show page' and the 'inventory list'


ryelo:
@ryandillinfelton is this a dupe of #157 ?

ryandillinfelton:
> @ryandillinfelton is this a dupe of #157 ?

They we're going to be seperate PRs, but that ended up not making sense. I'll close this one

PR #166, opened 2019-10-01
Unit tests for FilterChips

tumido:
Add unit tests for FilterChips

Covers:
- [x] rendering
- [x] API `onDelete` handler
- [x] defaults

Not covered:
- Group delete via `onDelete` since this is missing from design

cc @karelhala


tumido:
The merge conflict requires a decision on how you'd like to approach the plain filters:
1) as a separate group
or
2) as a plain list of chips

The conflict in code:
```js
onDelete(event, [{ category: 'group_plain', chips: [ chip ]}])
# or
onDelete(event, [ chip ]);
```

Which API fits you better? I assumed the plain filter structure, like the one shown in the `FilterChips.test.js`...

karelhala:
@tumido you can go with the second one (covered by tests) `onDelete(event, [ chip ]);` we can always identify if the chip is part of group or not.

PR #173, opened 2019-10-01
Update linter to consume subdirectories

karelhala:


karelhala:
cc @rvsia Sources needs to update as well, can you look at it if it's correct?

rvsia:
@karelhala looks ok! :trophy:

PR #177, opened 2019-10-01
Modular compliance rules table

dLobatog:
These changes are required for https://github.com/RedHatInsights/compliance-frontend/pull/197

### DIFF
--- a/packages/inventory-compliance/src/Compliance.js
+++ b/packages/inventory-compliance/src/Compliance.js
@@ -4,7 +4,7 @@ import routerParams from '@redhat-cloud-services/frontend-components-utilities/f
 import SystemPolicyCards from './SystemPolicyCards';
 import SystemRulesTable from './SystemRulesTable';
 import { sortable } from '@patternfly/react-table';
-import { ANSIBLE_ICON } from './Constants'
+import { ANSIBLE_ICON } from './Constants';
 import { Query } from 'react-apollo';
 import gql from 'graphql-tag';
 import { ApolloProvider } from 'react-apollo';
@@ -59,13 +59,13 @@ query System($systemId: String!){
 }
 `;
 
-const columns = [
+export const columns = [
     { title: 'Rule', transforms: [ sortable ] },
     { title: 'Policy', transforms: [ sortable ] },
     { title: 'Severity', transforms: [ sortable ] },
     { title: 'Passed', transforms: [ sortable ] },
     { title: <React.Fragment>{ ANSIBLE_ICON } Ansible</React.Fragment>, original: 'Ansible', transforms: [ sortable ] }
-]
+];
 
 const SystemQuery = ({ data, loading, hidePassed }) => (
     <React.Fragment>
--- a/packages/inventory-compliance/src/Compliance.js
+++ b/packages/inventory-compliance/src/Compliance.js
@@ -59,7 +59,7 @@ query System($systemId: String!){
 }
 `;
 
-export const columns = [
+const columns = [
     { title: 'Rule', transforms: [ sortable ] },
     { title: 'Policy', transforms: [ sortable ] },
     { title: 'Severity', transforms: [ sortable ] },
--- a/packages/inventory-compliance/src/Compliance.js
+++ b/packages/inventory-compliance/src/Compliance.js
@@ -3,6 +3,8 @@ import propTypes from 'prop-types';
 import routerParams from '@redhat-cloud-services/frontend-components-utilities/files/RouterParams';
 import SystemPolicyCards from './SystemPolicyCards';
 import SystemRulesTable from './SystemRulesTable';
+import { sortable } from '@patternfly/react-table';
+import { ANSIBLE_ICON } from './Constants'
 import { Query } from 'react-apollo';
 import gql from 'graphql-tag';
 import { ApolloProvider } from 'react-apollo';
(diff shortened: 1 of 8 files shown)


packages/inventory-compliance/src/index.js
```diff
@@ -1 +1,2 @@
 export { default as default } from './Compliance';
+export { default as SystemRulesTable } from './SystemRulesTable';
```
dLobatog: @akofink the line above was `export { Compliance as default } ` before I changed it, which is what caused the problem you saw. Since compliance-frontend uses the default export, it was failing to pick up any as it was undefined.

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -211,28 +180,31 @@ class SystemRulesTable extends React.Component {
...
-                original: remediationAvailable
-            }
-        ]
-    })
+    calculateParent = ({ profile }, { title, severity, compliant, identifier, remediationAvailable }) => {
+        const { columns } = this.state;
+        return {
+            isOpen: false,
+            cells: [
+                columns.find((column) => column.title === 'Rule') ? { title: this.ruleTitleCell(title, identifier), original: title } : null,
```
bastilian: `columns.find(...) ? ... : null` could be extracted into a function.

packages/inventory-compliance/src/Compliance.js
```diff
@@ -57,12 +59,21 @@ query System($systemId: String!){
 }
 `;
 
+const columns = [
```
bastilian: It wouldn't be the worst to export this to use in the tests, but just a suggestion.

packages/inventory-compliance/src/SystemRulesTable.test.js
```diff
@@ -117,14 +145,15 @@ describe('SystemRulesTable component', () => {
...
+                columns={ columns }
             />
         );
         const instance = wrapper.instance();
         await instance.setInitialCurrentRows();
         await instance.updateFilter(false, [ 'high', 'medium' ], []);
         expect(wrapper.state('currentRows').length / 2).toEqual(50);
         wrapper.state('currentRows').forEach((row, i) => {
-            if (row.hasOwnProperty('parent')) {
+            if (Object.prototype.hasOwnProperty.call(row, 'parent')) {
```
bastilian: Curious why this change is needed? calling `hasOwnProperty` on `row` should be the same.

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -559,8 +533,8 @@ class SystemRulesTable extends React.Component {
                         onCollapse={ this.onCollapse }
                         onSort={ this.onSort }
                         sortBy={ sortBy }
-                        onSelect={ (currentRows.length !== 0) ? this.onSelect : () => null }
-                        rows={ (currentRows.length === 0) ? emptyRows : currentRows }>
+                        onSelect={ (remediationsEnabled && currentRows.length !== 0) ? this.onSelect : undefined }
```
bastilian: Passing undefined could cause issues. The `Table` component could still try to call what is passed to `onSelect` and result in an unwanted error or warning. Using the mute function is not optimal. Optimal would be to only pass the prop if it's set, that would require to prepare the props before hand and spread them in.

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -67,13 +42,7 @@ class SystemRulesTable extends React.Component {
...
         super(props);
         this.state = {
-            columns: [
-                { title: 'Rule', transforms: [ sortable ] },
-                { title: 'Policy', transforms: [ sortable ] },
-                { title: 'Severity', transforms: [ sortable ] },
-                { title: 'Passed', transforms: [ sortable ] },
-                { title: <React.Fragment>{ ANSIBLE_ICON } Ansible</React.Fragment>, transforms: [ sortable ] }
-            ],
+            columns: props.columns,
```
bastilian: The `columns` prop seems to be missing a type validation in the prop-types. It might be good to verify the shape of each `column` item as well. and possibly a default value.

packages/inventory-compliance/src/SystemRulesTable.test.js
```diff
@@ -12,8 +12,17 @@ debounce.mockImplementation(fn => fn);
 global.fetch = require('jest-fetch-mock');
 
 describe('SystemRulesTable component', () => {
+    let columns;
```
bastilian: It doesn't really change through out the tests, why not define it with `const`?

packages/inventory-compliance/src/SystemRulesTable.test.js
dLobatog: New `eslint` requirement for the whole project :man_shrugging:

dLobatog:
@akofink I didn't check with `LOCAL_CHROME=false` but this PR isn't supposed to change how the systems table works (other than passing in the column headers as props) - it's meant mostly to make the rules table without remediations available in the Wizard

dLobatog:
@akofink Just checked and it doesn't work for me either, there seems to be some problem with `withRouter` - that's on SystemDetails though, on `compliance-frontend` , not here :disappointed_relieved:  - this PR doesn't have much to do with it as without it it throws the same error.

akofink:
Okay, yeah - I agree this PR doesn't seem to make the problem better or worse. I'm glad it's not just me! :octopus:

packages/inventory-compliance/src/Compliance.js
```diff
@@ -144,6 +146,13 @@ SystemDetails.propTypes = {
             inventoryId: propTypes.string
         })
     }),
+    columns: propTypes.shape([
+        {
+            title: propTypes.oneOfType([ propTypes.string, propTypes.object ]).isRequired,
```
bastilian: If the title should also accept React elements you could use `PropTypes.element`, instead of the `propTypes.oneOfType(...)`, which allows (i think both) strings and objects/elements.

PR #179, opened 2019-10-03
RHICOMPL-316 - FailedRules doesn't filter correctly

dLobatog:
This PR is meant to fix a regression related to auto-filtering passed rules whenever a user clicks on "failed rules"  on the compliance policy page. It's not really relevant when the compliance component is displayed on the inventory.

Peek 2019-10-03 11-34

### DIFF
--- a/packages/inventory-compliance/src/SystemRulesTable.js
+++ b/packages/inventory-compliance/src/SystemRulesTable.js
@@ -77,7 +77,7 @@ class SystemRulesTable extends React.Component {
             page: 1,
             itemsPerPage: 10,
             rows: [],
-            hidePassed: false,
+            hidePassed: props.hidePassed,
             severity: [],
             policy: [],
             currentRows: [],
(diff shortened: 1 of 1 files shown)


packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -129,13 +128,11 @@ class SystemRulesTable extends React.Component {
...
 
     changePage = (page, itemsPerPage) => {
-        this.setState(() => (
-            {
-                currentRows: this.currentRows(page, itemsPerPage),
-                page,
-                itemsPerPage
-            }
-        ));
+        this.setState({
```
dLobatog: I took the chance also to remove the callback in all of the `setState`. Before, we needed the callback because `currentRows` was making an async call to the Remediations API. But that's no longer relevant  since https://github.com/RedHatInsights/frontend-components/commit/3c6c2762227b96b80edf70b1aebcc39f13249683

PR #185, opened 2019-10-07
Hide select if only one filter is used

karelhala:

PR #184, opened 2019-10-07
Allow custom components in Primary toolbar

karelhala:
This PR will enable custom somponents be part of Primary toolbar in order to allow rewriting of certain parts.

PR #186, opened 2019-10-07
Add new data toolbar to inventory

karelhala:
### Removed Filters from inv
Since we never really used filters in inventory this PR removes them altogether.

### Before
Screenshot from 2019-10-09 14-52-21

### After 
Screenshot from 2019-10-09 14-51-49

On hold until:
* [x] https://github.com/RedHatInsights/frontend-components/pull/184
* [x] https://github.com/RedHatInsights/frontend-components/pull/185

Jira issue RHCLOUD-2503


packages/inventory/src/EntityTableToolbar.js
```diff
@@ -1,61 +1,111 @@
...
+
+    render() {
+        const {
+            total,
+            page,
+            onRefresh,
+            perPage,
+            filters,
+            hasItems,
+            // eslint-disable-next-line no-unused-vars
```
AllenBW: we do this more than once, time to throw `/* eslint-disable no-unused-vars */` on the top

packages/inventory/src/EntityTableToolbar.js
```diff
@@ -1,61 +1,108 @@
...
+const TEXT_FILTER = 'hostname_or_id';
+
+class EntityTableToolbar extends Component {
+    state = {
+        textFilter: ''
+    }
+
+    debouncedRefresh = debounce((config) => {
+        this.props.onRefresh && this.props.onRefresh(config);
+    }, 800);
```
shdunning: is that 800ms pause between typing? is that typical for a typeahead / autocomplete solution (I have no idea)?

https://ux.stackexchange.com/questions/36231/when-should-you-start-autocompleting-a-search-in-autocomplete-search-field?rq=1

shdunning: i realize that we can always tweak this later...

karelhala: Yes, the 800 is in milliseconds between requests to server. If user does not type anything in this window request is sent. On the other hand if user is still typing no new requests are sent to server. In the past we chose 800 as the base for this, however we should perhaps add this number to utils constant in order to make this persistent accross entire platform - https://github.com/RedHatInsights/frontend-components/issues/197

karelhala:
Released in package version [@redhat-cloud-services/frontend-components-inventory/v/0.0.23](https://www.npmjs.com/package/@redhat-cloud-services/frontend-components-inventory/v/0.0.23)

PR #191, opened 2019-10-08
Fix problem with first action throwing error if button was not used

karelhala:
There was a nasty bug that was triggered when only object was sent to PrimaryToolbar instead of React component as fisrt action.


karelhala:
Released in new package version [@redhat-cloud-services/frontend-components/v/0.0.24](https://www.npmjs.com/package/@redhat-cloud-services/frontend-components/v/0.0.24)

PR #201, opened 2019-10-10
Prevent default and buble on Label in Group filter

karelhala:
### UI changes
None

### Fixes
When user clicks on label in Grouped filter it will fire 2 actions - `onClick` and `onChange` each of them are fired from PF component so we have to prevent buble if user clicks on label otherwise the value will be selected and deselected right after.


karelhala:
Released in new package version [@redhat-cloud-services/frontend-components/v/0.0.26](https://www.npmjs.com/package/@redhat-cloud-services/frontend-components/v/0.0.26)

PR #208, opened 2019-10-11
Send inifity to chipCount and merge props to selected in primary filter

karelhala:
### Ui changes
None

### Fixes
* When more than 3 filters were passed to PF filter group `Show more` or `Show less` was shown - pass `Infinity` to chip group to prevent this
* Incorrect calculation of groupped filter, by merging only state's selected values instead of combindation of props and state


karelhala:
Released in new package version [@redhat-cloud-services/frontend-components/v/0.0.26](https://www.npmjs.com/package/@redhat-cloud-services/frontend-components/v/0.0.26)

PR #210, opened 2019-10-14
AddSourceWizard UX enhancements

rvsia:
**Description**

- updates DDF to latest
- `showTitles` prop is used to generate title (no custom logic)
- A `product_name` is added to step title
- Nested subforms have padding left
- when only one auth_type, it is selected automatically

**Before**

initial

image

selected

image

**After**

initial/selected

image

@Hyperkid123 
@terezanovotna


packages/sources/src/styles/subform.scss
```diff
@@ -0,0 +1,3 @@
+.ins-c-sources__wizard--subform-nested {
+    padding-left: var(--pf-global--spacer--md);
```
Hyperkid123: you can use pf class instead `pf-u-pl-md` and use it instead. This class does not need to exist.

rvsia: Thanks! :taco:

PR #213, opened 2019-10-14
Add filters to inventory toolbar

karelhala:
### TODO
* [x] merge after https://github.com/RedHatInsights/frontend-components/pull/218

### UI changes
*Before*
Screenshot from 2019-10-14 17-03-38

*After*
Screenshot from 2019-10-14 16-58-03
Screenshot from 2019-10-14 16-58-11
Screenshot from 2019-10-14 16-58-32

### DIFF
--- a/packages/inventory/src/EntityTableToolbar.js
+++ b/packages/inventory/src/EntityTableToolbar.js
@@ -3,31 +3,133 @@ import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import { Skeleton, SkeletonSize, PrimaryToolbar } from '@redhat-cloud-services/frontend-components';
+import { fetchAllTags, clearFilters } from './redux/actions';
 import debounce from 'lodash/debounce';
 import { InventoryContext } from './Inventory';
-
-const TEXT_FILTER = 'hostname_or_id';
+import {
+    mapGroups,
+    TEXT_FILTER,
+    reduceFilters,
+    constructGroups,
+    TEXTUAL_CHIP,
+    TAG_CHIP
+} from './constants';
+import flatMap from 'lodash/flatMap';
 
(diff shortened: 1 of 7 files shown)


packages/inventory/src/constants.js
```diff
@@ -0,0 +1,68 @@
...
+            label: `${tagName} - ${tagValue}`
+        }))
+    }));
+}
+
+export function reduceFilters(filters) {
+    return filters.reduce((acc, oneFilter) => {
+        if (oneFilter.value === TEXT_FILTER) {
+            return { ...acc, textFilter: oneFilter.filter };
+        } else if (oneFilter.hasOwnProperty('tagFilters')) {
```
Hyperkid123: Since user can have any keys as the object attribute, maybe it would be better to use `Object.hasOwnProperty` to prevent possible runtime crash.

packages/inventory/src/EntityTableToolbar.js
```diff
@@ -3,32 +3,128 @@ import React, { Component } from 'react';
...
+            onDelete: (_e, [ deleted ], isAll) => {
+                if (isAll) {
+                    onRefresh({ page, perPage, filters: [] });
+                    onClearFilters();
+                    this.setState({
+                        selected: {},
+                        textFilter: ''
+                    });
+                } else {
+                    if (deleted.type === 'textual') {
```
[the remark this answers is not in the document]
karelhala: Sure thing!

packages/inventory/src/constants.js
karelhala: I don't really like `Object.hasOwnProperty` because it's so long, however [JS#Operators/in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in) could be better here.

Hyperkid123: Works to

PR #219, opened 2019-10-16
Update export button to show on new place and use correct styling

karelhala:
### UI changes
*Before*
Screenshot from 2019-10-16 16-27-24

*After spacing updates*
Screenshot from 2019-10-21 17-39-32

*After*
Screenshot from 2019-10-16 16-13-46
Screenshot from 2019-10-16 16-13-59

### DIFF
--- a/packages/components/src/Components/DownloadButton/DownloadButton.js
+++ b/packages/components/src/Components/DownloadButton/DownloadButton.js
@@ -26,16 +26,20 @@ class DownloadButton extends Component {
         return (
             <Dropdown
                 { ...props }
+                isPlain
                 onSelect={ this.onSelect }
                 toggle={
-                    <DropdownToggle onToggle={ this.onToggle } >
+                    <DropdownToggle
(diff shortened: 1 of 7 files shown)


karelhala:
Yeah, I've updated the first spacing to use 24px spacer
Screenshot from 2019-10-16 18-18-31

katierik:
@mattnolting is there anything you can think of around toolbar that could sort out some of this weird spacing? It doesn't seem to match the PF documentation and not sure if docs are wrong or there's something missing.

ryelo:
My guess is that it's because those icons are all "squares", even if they don't look like it. So even though the kebab is very vertical, it'll still take up the same horizontal space as the other icons, thus leaving extra white space to the left and right of the kebab "inside" the icon itself

karelhala:
This one should be good to review @RedHatInsights/ui-reviewers @katierik ACK on spacing updates.

Screenshot from 2019-10-21 17-39-32

PR #220, opened 2019-10-17
RHICOMPL-395 - Update Remediate with Ansible button to use icon

dLobatog:
Screenshot from 2019-10-17 11-54-20

New look to match our mocks (@katierik) https://marvelapp.com/c2bcd9h/screen/58340320


dLobatog:
@akofink This `ComplianceRemediationButton` is not the same that's used in there, this one is only used in the SystemDetails page. Actually now with the cached remediations I think we might be able to merge them into 1 as the code has been simplified a lot. 

https://github.com/RedHatInsights/compliance-frontend/pull/201 contains the changes for the other one.

Issue #221, opened 2019-10-17
[BUG] Inventory row actions visible even without rows


jiridostal:
**Describe the bug**
Inventory table row actions are visible even for a table that has zero rows to display.
![image](https://user-images.githubusercontent.com/6286045/67006451-74d7fd80-f0e5-11e9-8a8a-4f878a7b5095.png)
 

**To Reproduce**
1. Have Inventory table with actions for each row
2. Filter/remove rows so that empty state is displayed
3. See error

**Expected behavior**
No action kebab visible

PR #222, opened 2019-10-17
Hide actions if no rows present

karelhala:
### UI changes
*Before*
Screenshot from 2019-10-17 14-45-48

*After*
Screenshot from 2019-10-17 14-42-55

### Fixes
Closes: #221

### DIFF
--- a/packages/inventory/src/EntityTable.js
+++ b/packages/inventory/src/EntityTable.js
@@ -181,7 +181,7 @@ class EntityTable extends React.Component {
                         { ...{
                             ...hasCheckbox && rows.length !== 0 ? { onSelect: this.onItemSelect } : {},
                             ...expandable ? { onCollapse: onExpandClick } : {},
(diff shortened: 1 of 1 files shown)


packages/inventory/src/EntityTable.js
```diff
@@ -181,7 +181,7 @@ class EntityTable extends React.Component {
                         { ...{
                             ...hasCheckbox && rows.length !== 0 ? { onSelect: this.onItemSelect } : {},
                             ...expandable ? { onCollapse: onExpandClick } : {},
-                            ...actions ? { actions } : {}
+                            ...actions && rows.length > 0 ? { actions } : {}
```
AllenBW: looks like this could also be written `...actions && rows.length > 0 && { actions } ` 🤔

karelhala: Good point, I'll update it

PR #223, opened 2019-10-17
Fix some linting rules in codebase

karelhala:


packages/components/src/Components/Battery/Batery.test.js
```diff
@@ -42,6 +42,7 @@ describe('Battery component', () => {
         it('NullBatery, default', () => {
             const wrapper = shallow(<Battery severity={''} label={''}/>);
             expect(toJson(wrapper)).toMatchSnapshot();
+            // eslint-disable-next-line no-console
```
Hyperkid123: I think that we should permit warn and error but blacklist log.

We could use this rule globally: `no-console: ["error", { allow: ["warn", "error"] }]`

What do you think?

packages/components/src/Components/Breadcrumbs/Breadcrumbs.js
```diff
@@ -4,6 +4,7 @@ import classnames from 'classnames';
 import { Breadcrumb, BreadcrumbItem } from '@patternfly/react-core';
 
 const Breadcrumbs = ({ items, current, className, onNavigate, ...props }) => {
+    // eslint-disable-next-line no-console
```
Hyperkid123: ^^

packages/components/src/Components/Main/Main.js
```diff
@@ -35,6 +35,7 @@ export class Main extends Component {
         };
     }
     render () {
+        // eslint-disable-next-line no-unused-vars
```
Hyperkid123: You can also use this rule to permit spread operator siblings
https://eslint.org/docs/rules/no-unused-vars#ignorerestsiblings

Hyperkid123 (commented):
I think avoid some of those `/* eslint-...` by adding/editing of certain rules.

packages/components/src/Components/Battery/Batery.test.js
karelhala: I don't really like `error` for logs, but allowing warnings and errors is a good idea. I'll update it!

Hyperkid123 (approved):
:name_badge: :  :name_badge:  :name_badge:  :name_badge:  :snake:

PR #226, opened 2019-10-18
Show error message in AddSourceWizard

rvsia:
**Description**

- Shows error message in the error step:

image

- Tries to remove the source, if something goes wrong, 
       - if the deletion is unsuccessful, there is a message telling user to try to remove it again later

TODO:
- [x] tests

@Hyperkid123 @karelhala @terezanovotna

### DIFF
--- a/packages/sources/src/api/handleError.js
+++ b/packages/sources/src/api/handleError.js
@@ -1,0 +1,30 @@
+import { getSourcesApi } from './index';
+
+export const handleError = (id, error) => {
+    if (!error) {
+        return 'Undefined error';
+    }
+
+    if (typeof error === 'string') {
+        return error;
+    }
+
+    const url = error.config.url;
+    let detail;
+
+    if (typeof error.response.data === 'string') {
+        detail = error.response.data;
+    } else {
+        detail = error.response.data.errors[0].detail;
+    }
+
+    const detailInfo = error.detailInfo ? `(${error.detailInfo}) ` : ``;
+
+    return getSourcesApi().deleteSource(id)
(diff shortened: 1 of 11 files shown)


packages/sources/src/api/handleError.js
```diff
@@ -0,0 +1,10 @@
+import { getSourcesApi } from './index';
+
+export const handleError = (id, { config: { url }, response: { data: { errors: [{ detail }] } } }) => getSourcesApi()
+.deleteSource(id)
+.catch((errorDelete) => {
+    throw `${url}: ${detail}. The source was not removed, try remove it later: ${errorDelete.response.data.errors[0].detail}`; }
+)
+.then(() => {
+    throw `${url}: ${detail}`;
```
Hyperkid123: I suppose you want to throw multiple errors?

rvsia: `Promise.All `returns only the first error, so no, I just want to show the first returned error.

Hyperkid123: Yeah i know i am just curious why there is throw in `then` branch.

rvsia: This is like an error 'post-process', the error is then catched by the outer catch branch.

Hyperkid123: You can do mutations via graphql? I though its not supported.

rvsia: I don't understand what you mean

rvsia:
@Hyperkid123 updated! :+1:

PR #243, opened 2019-10-25
Use DateFormat component to correctly show times in inventory

karelhala:
### UI Changes
* **Inventory table**
*Before*
Screenshot from 2019-10-25 15-08-53
*After*
Screenshot from 2019-10-25 14-54-41
* **Inventory detail**
*Before*
Screenshot from 2019-10-25 15-08-59
*After*
Screenshot from 2019-10-25 14-54-48

* **Inventory detail**
*Before*
Screenshot from 2019-10-25 15-09-03
*After*
Screenshot from 2019-10-25 15-07-27

### DIFF
--- a/packages/inventory-general-info/src/CollectionCard.js
+++ b/packages/inventory-general-info/src/CollectionCard.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import LoadingCard from './LoadingCard';
 import { collectionInformationSelector } from './selectors';
+import { DateFormat } from '@redhat-cloud-services/frontend-components';
 
 const CollectionCard = ({ detailLoaded, collectionInformation, entity }) => (<LoadingCard
     title="Collection information"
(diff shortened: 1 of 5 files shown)


karelhala:
Ping @katierik @bclarhk

packages/inventory/src/EntityTable.js
```diff
@@ -63,7 +63,11 @@ class EntityTable extends React.Component {
             }
 
             if (isTime) {
-                const [ day, date, month, year, time ] = new Date(get(col, key, ' ')).toUTCString().split(' ');
+                if (DateFormat) {
+                    return { title: <DateFormat date={get(col, key, ' ')} type="relative" /> };
+                }
+
+                const [ , date, month, year, time ] = new Date(get(col, key, ' ')).toUTCString().split(' ');
```
AllenBW: is leading comma intentional? 🤔

packages/inventory-general-info/src/CollectionCard.js
```diff
@@ -3,15 +3,24 @@ import PropTypes from 'prop-types';
...
         { title: 'Egg', value: collectionInformation.egg },
-        { title: 'Last check-in', value: entity && new Date(entity.updated).toLocaleString() },
-        { title: 'Registered', value: entity && new Date(entity.created).toLocaleString() }
+        { title: 'Last check-in', value: entity && (
+            DateFormat ?
+                <DateFormat date={ entity.updated } type="onlyDate" /> :
+                new Date(entity.updated).toLocaleString()
+        ) },
+        { title: 'Registered', value: entity && (
+            DateFormat ?
```
AllenBW: Just a question... is `DateFormat` loading so slow that yah gotta test for its availability before using?

karelhala: That's to check if the component is actually accessible, since it's a new-ish component and the consumer could have outdated FEC bundle. If that would happen users could have see a broken UI.

packages/inventory/src/EntityTable.js
karelhala: Yup, that's to omit first element from array.

PR #247, opened 2019-10-28
Updates inv-insights toolbar to use primarytoolbar component

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-2901


AllenBW:
cc @katierik @robcwilliams

packages/inventory-insights/src/Constants.js
```diff
@@ -1,24 +1,19 @@
...
-            { label: 'No support', value: false }
+export const FILTER_CATEGORIES = {
+    has_playbook: {
+        type: 'radio', title: 'Ansible support', urlParam: 'has_playbook', values: [
+            { label: 'Ansible remediation support', value: 'true' },
+            { label: 'No Ansible remediation support', value: 'false' },
+            { label: 'All rules', value: 'all' }
         ]
-    }, {
+    }, total_risk: {
```
karelhala: Nitpick: can you put this to new line? It will look slightly better.

packages/inventory-insights/src/Constants.js
```diff
@@ -1,24 +1,19 @@
...
 
-export const FILTER_CATEGORIES = [
-    {
-        type: 'checkbox', title: 'Ansible support', urlParam: 'has_playbook', values: [
-            { label: 'Has support', value: true },
-            { label: 'No support', value: false }
+export const FILTER_CATEGORIES = {
+    has_playbook: {
+        type: 'radio', title: 'Ansible support', urlParam: 'has_playbook', values: [
+            { label: 'Ansible remediation support', value: 'true' },
```
karelhala: Is it intentional to have these `value`s boolean string?

AllenBW: yeah, conditional filter will eat up the false and or undefined (resort to displaying the array position of the entry) so we do strings to avoid that

packages/inventory-insights/src/Constants.js
AllenBW: resolved!

packages/inventory-insights/src/Constants.js
karelhala: Oh I get it, might want to open an issue to address this.

karelhala: https://github.com/RedHatInsights/frontend-components/issues/250

PR #248, opened 2019-10-29
Date format tooltip

PanSpagetka:
Add tooltip for DataFormat when it is relative.

Also add more tests with different input types (string, number, date)

### DIFF
--- a/packages/components/src/Components/DateFormat/DateFormat.test.js
+++ b/packages/components/src/Components/DateFormat/DateFormat.test.js
@@ -18,4 +18,14 @@ describe('DateFormat component', () => {
         const wrapper = shallow(<DateFormat date={new Date('Dec 31 2019 00:00:00 UTC')} />);
         expect(toJson(wrapper)).toMatchSnapshot();
     });
+
+    it('DateFormat renders with date integer', () => {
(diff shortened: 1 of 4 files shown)


packages/components/src/Components/DateFormat/DateFormat.test.js
```diff
@@ -4,8 +4,18 @@ import toJson from 'enzyme-to-json';
 import DateFormat from './DateFormat';
 
 describe('DateFormat component', () => {
-    it('DateFormat renders', () => {
-        const wrapper = shallow(<DateFormat />);
+    it('DateFormat renders with date integer', () => {
```
karelhala: Can you also add tests to render `exact` and `onlyDate`?

packages/components/src/Components/DateFormat/helper.js
```diff
@@ -16,9 +19,25 @@ const relativeTimeTable = [
 
 const exact = (value) => value.toUTCString().split(',')[1].slice(0, -4).trim();
 
+export const addTooltip = (date, element) => (
+    <Tooltip
+        content={
+            <div>date</div>
+        }
```
karelhala: This should be wrapped in `{}`, also no need to space it that much. `content={ <div>{date}</div> }`

PR #249, opened 2019-10-29
Enable comment driven releases

karelhala:
### New feature
This PR brins in auto release to our packages, it is triggered by adding comment starting with `release` and specifying either `minor` or `bugfix`, by default it will fallback to `bugfix`. Major is not present because we want to control major releases by ourselves.

This is brought to you by https://probot.github.io/ using https://github.com/karelhala/comment-travis

Only certain people can trigger release and only merged PRs are allowed.

I might need a bit of help on styling comments (I am not that good with emoticons), currently these values are used:
`We will trigger new Release: ${type}!`
`Sorry no release, PR has not been merged.`
`New version of package has been released [${pckgName}/v/${newVersion}](https://www.npmjs.com/package/${pckgName}/v/${newVersion})`


karelhala:
@ryelo sure thing!

PR #254, opened 2019-11-01
Add date string exact to tooltip of date format component

karelhala:
## Changes
* it will add UTC string to tooltip

## Fixes
* error when using this component since react can't render date correctly, it needs a parsed string


Hyperkid123 (approved):
:hot_pepper:

PR #257, opened 2019-11-04
Add direct components to PrimaryToolbar

karelhala:
### Changes
None

### Fixes
When new render happens each component is re-rendered and text filters lost focus.

### DIFF
--- a/packages/components/src/Components/ConditionalFilter/ConditionalFilter.js
+++ b/packages/components/src/Components/ConditionalFilter/ConditionalFilter.js
@@ -32,7 +32,7 @@ class ConditionalFilter extends Component {
             items[0]
         );
         const onChangeCallback = onChange || this.onChange;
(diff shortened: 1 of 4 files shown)


packages/components/src/Components/ConditionalFilter/ConditionalFilter.js
```diff
@@ -32,7 +32,7 @@ class ConditionalFilter extends Component {
             items[0]
         );
         const onChangeCallback = onChange || this.onChange;
-        const ActiveComponent = activeItem && typeMapper(activeItem.type);
+        const ActiveComponent = activeItem && (typeMapper[activeItem.type] || typeMapper.text);
```
AllenBW: oh oh oh  awesome, so we can either explicitly declare text type or just not n its the default 👍

PR #258, opened 2019-11-05
Adds rules to support ts and tsx files using ts-loader

josejulio:
I was trying to add support for typescript into a copy of the starter-app and had to patch [1] some configuration from webpack to have it working.
I was working if it would make sense to have it here instead.

If this makes sense, i could write a README on how to use typescript with the starter-app or have a branch that uses typescript.

Note that this `module.rules[]` could be added by default to every project and shouldn't matter as long as there is no ts file. If there is one, it will be handled by `ts-loader` (fail if `ts-loader` is not installed, or missing a `tsconfig.json` file)

[1] https://github.com/josejulio/insights-frontend-starter-app/commit/c296da118f7d6fce6315689bfcdf9ce07c580e98#diff-15993b2c8c6dac58acfc99f3e39f91f4R17-R33

### DIFF
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -34,6 +34,8 @@
         "sass-loader": "^7.1.0",
         "source-map-loader": "^0.2.4",
         "style-loader": "^0.23.1",
+        "ts-loader": "^6.2.1",
+        "webpack": "^4.41.2",
         "webpack-dev-server": "^3.4.1",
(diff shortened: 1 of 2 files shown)


josejulio:
Thanks @karelhala !

Followed all your feedback, and also added webpack as a dependecy, noticed it wasn't there and is used in `plugins.js`. It was also giving me troubles when trying to use as a local dependency with `yarn link`.

PR #259, opened 2019-11-05
Allow passing an object/component for Filters label prop

bastilian:
In the compliancce frontend we pass an icon component as `label`<sup>[1]</sup>, which causes a prop validation error, as it is expecting a string. This adds objects as a valid prop type for label.

[1] https://github.com/RedHatInsights/compliance-frontend/blob/1531fb26ec0a67805d9c4d6b4e7a1526c4009e77/src/SmartComponents/SystemsComplianceFilter/SystemsComplianceFilter.js#L70

### DIFF
--- a/packages/components/src/Components/Filters/FilterDropdown.js
+++ b/packages/components/src/Components/Filters/FilterDropdown.js
@@ -82,7 +82,7 @@ FilterDropdown.propTypes = {
             )
         })
     ),
-    label: PropTypes.string
+    label: PropTypes.oneOfType([ PropTypes.string, PropTypes.object ])
 };
 
 FilterDropdown.defaultProps = {
--- a/packages/components/src/Components/Filters/FilterDropdown.js
+++ b/packages/components/src/Components/Filters/FilterDropdown.js
@@ -82,7 +82,10 @@ FilterDropdown.propTypes = {
             )
         })
     ),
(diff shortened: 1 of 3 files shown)


packages/components/src/Components/Filters/FilterDropdown.js
```diff
@@ -82,7 +82,7 @@ FilterDropdown.propTypes = {
             )
         })
     ),
-    label: PropTypes.string
+    label: PropTypes.oneOfType([ PropTypes.string, PropTypes.object ])
```
karelhala: This should be `PropTypes.node` React can't render plain object.

packages/components/src/Components/Filters/FilterDropdown.test.js
```diff
@@ -3,15 +3,30 @@ import { shallow } from 'enzyme';
 import toJson from 'enzyme-to-json';
 import FilterDropdown from './FilterDropdown';
 
+const TestComponent = () => {
```
karelhala: This is not really a testComponent and allows objects to be rendered, which is not correct. You can redner `<React.Fragment/> ` instead of this `TestComponent`

bastilian:
@karelhala Thank you for reviewing! The proptype is corrected and the test now uses `<React.Fragment />`.

bastilian:
🤦‍♂ I updated the snapshots and then changed the test title. It's now updated also with the new title.

PR #261, opened 2019-11-06
Build was failing on importing apolo deps

karelhala:
Failing build can be seen https://github.com/RedHatInsights/insights-inventory-frontend/pull/171

### DIFF
--- a/packages/config/src/config.js
+++ b/packages/config/src/config.js
@@ -63,10 +63,15 @@ module.exports = ({
                         outputPath: 'fonts/'
                     }
                 }]
+            },
+            {
+                test: /\.mjs$/,
+                include: /node_modules/,
+                type: 'javascript/auto'
             }]
         },
         resolve: {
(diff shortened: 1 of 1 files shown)


packages/config/src/config.js
```diff
@@ -63,10 +63,15 @@ module.exports = ({
...
+            },
+            {
+                test: /\.mjs$/,
+                include: /node_modules/,
+                type: 'javascript/auto'
             }]
         },
         resolve: {
-            extensions: [ '.ts', '.tsx', '.js', '.scss' ]
+            extensions: [ '.ts', '.tsx', '.mjs', '.js', '.scss' ]
```
akofink: Aren't ECMAScript modules an experimental feature?

karelhala: That's actually what the preferes solution is about https://github.com/graphql/graphql-js/issues/1272

The problem is that you can't import named exports multiple times and this was breaking it.

PR #274, opened 2019-11-14
Sources: Modify export to support edit form

rvsia:
Additional things needed to be exported to be able to generate edit form in sources ui


rvsia:
@karelhala

PR #275, opened 2019-11-14
Fix rules expand when filters are applied

dLobatog:
Before:
Peek 2019-11-14 18-25

After:
Peek 2019-11-14 20-42

This needs tests to ensure we don't break this further IMO. Don't merge this until that's done.

### DIFF
--- a/packages/inventory-compliance/src/SystemRulesTable.js
+++ b/packages/inventory-compliance/src/SystemRulesTable.js
@@ -342,7 +342,7 @@ class SystemRulesTable extends React.Component {
     compressRows = (rows) => {
         const compressedRows = [];
         rows.forEach((row, i) => {
-            if (row.hasOwnProperty('isOpen')) {
+            if (this.isParent(row)) {
                 compressedRows.push({ parent: row, child: rows[i + 1] });
             }
         });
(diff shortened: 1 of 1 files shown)


packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -342,7 +342,7 @@ class SystemRulesTable extends React.Component {
     compressRows = (rows) => {
         const compressedRows = [];
         rows.forEach((row, i) => {
-            if (row.hasOwnProperty('isOpen')) {
+            if (this.isParent(row)) {
```
dLobatog: There's a function for `row.hasOwnProperty('isOpen')` that's more meaningful - I'm just using it, there's no actual change in behavior here

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -400,14 +400,15 @@ class SystemRulesTable extends React.Component {
...
+            if (this.isParent(row) && String(row.cells[column].original).toLowerCase().match(
                 String(attribute).toLowerCase()
             )) {
                 filteredRows.push(row);
-                if (!rows[i + 1].hasOwnProperty('isOpen')) {
-                    let child = rows[i + 1];
-                    child.parent = filteredRows.length - 1;
-                    filteredRows.push(child);
+                if (!this.isParent(rows[i + 1])) {
+                    filteredRows.push({
```
dLobatog: This is the gist of the fix.

Apparently, `child.parent = filteredRows.length - 1` was not respected properly, so the parent ID would not be 0, 2, 4, etc... it would be the parent ID of the row in the unfiltered array (so it could be something stupid like 12, 18...). 

Actually even doing `.push { parent: filteredRows.length - 1, ...rows[i + 1] }`, the parent attribute was still set to the `rows[i + 1]` attribute. I'm not even sure why, so I decided to pass `cells` explicitly and ignore the parent

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -442,7 +443,7 @@ class SystemRulesTable extends React.Component {
         let result;
 
         if (severity.length > 0 && hidePassed) {
-            result = passedRows.filter(row => severityRows.includes(row));
+            result = passedRows.filter(row => severityRows.map((row) => row.cells).includes(row.cells));
```
dLobatog: Here's where we combine filtered results. The filtered results, however, will have *different* parent for the children rows. So what we're doing here is making sure we only compare the `cells` and forget about the parent. Combining results worked well before because of the bug - `parent` was the same in any filtered row.

dLobatog:
Release bugfix

PR #276, opened 2019-11-15
RHICOMPL-417 - Fix compliance rule expand for Search

dLobatog:
Tests for https://github.com/RedHatInsights/frontend-components/pull/275

### DIFF
--- a/packages/inventory-compliance/src/SystemRulesTable.test.js
+++ b/packages/inventory-compliance/src/SystemRulesTable.test.js
@@ -44,13 +44,14 @@ describe('SystemRulesTable component', () => {
                 profileRules={ profileRules }
                 loading={ false }
                 system={ system }
+                itemsPerPage={ 100 }
             />
         );
         const instance = wrapper.instance();
         await instance.setInitialCurrentRows();
+        expect(wrapper.state('currentRows').length / 2).toEqual(52);
+        await instance.updateFilter(false, [ 'low' ], []);
         expect(wrapper.state('currentRows').length / 2).toEqual(2);
-        await instance.updateFilter(false, [ 'high' ], []);
-        expect(wrapper.state('currentRows').length / 2).toEqual(0);
     });
 
     it('should render filtered rows by multiple severities', async () => {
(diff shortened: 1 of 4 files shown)


bastilian:
@dLobatog there seems to be a linting error: https://travis-ci.org/RedHatInsights/frontend-components/builds/612289225#L1104

dLobatog:
@bastilian :see_no_evil:  - I just updated the PR to add the missing semicolon

packages/inventory-compliance/src/SystemRulesTable.test.js
```diff
@@ -105,4 +105,19 @@ describe('SystemRulesTable component', () => {
...
+                profileRules={ profileRules }
+                loading={ false }
+                system={ system }
+            />
+        );
+        const instance = wrapper.instance();
+        await instance.setInitialCurrentRows();
+        await instance.updateFilter(false, [ 'high', 'medium' ], []);
+        expect(wrapper.state('currentRows').length / 2).toEqual(1);
+        expect(wrapper.state('currentRows')[1].parent).toEqual(0);
```
akofink: Sometimes the behavior appears correct for one or two rows, but then another row will be messed up. Maybe we could go ahead and check that this is correct in a loop for `profileRules.length`

dLobatog: There are only 2 rules in the sample request so I'll have to change that

akofink: Yeah, I noticed that. :/ You could just set `profileRules = profileRules.concat(profileRules, profileRules, profileRules, ...)`

dLobatog:
I'll make the Fixtures file bigger to reflect reality a bit better, with 20 rules or so. With that, it should be much easier to test for these kind of things.

packages/inventory-compliance/src/SystemRulesTable.test.js
```diff
@@ -89,20 +92,83 @@ describe('SystemRulesTable component', () => {
...
         expect(wrapper.state('currentRows')[0].cells[TITLE_COLUMN].original).
         toEqual('Use direct-lvm with the Device Mapper Storage Driver');
         expect(wrapper.state('currentRows')[2].cells[TITLE_COLUMN].original).
         toEqual('Enable the Docker service');
         await instance.onSort(null, TITLE_COLUMN + 2, SortByDirection.asc);
-        expect(wrapper.state('currentRows').length / 2).toEqual(2);
+        expect(wrapper.state('currentRows').length / 2).toEqual(52);
         expect(wrapper.state('currentRows')[0].cells[TITLE_COLUMN].original).
-        toEqual('Enable the Docker service');
+        toEqual('Add nodev Option to /dev/shm');
```
dLobatog: After adding more results to the response, the first alphabetical row title has changed

PR #277, opened 2019-11-15
[RHICOMPL-420] - Re-enable search by references on Compliance

dLobatog:
Given the amount of rules we load now is much smaller than before, and we don't have to wait for a 2nd request to get `remediation_available`, it seems that we can turn back on this feature.  

Peek 2019-11-15 18-02

### DIFF
--- a/packages/inventory-compliance/src/SystemRulesTable.js
+++ b/packages/inventory-compliance/src/SystemRulesTable.js
@@ -228,8 +228,8 @@ class SystemRulesTable extends React.Component {
     )
 
     referencesList = (references) => references.reduce((acc, reference, i) => ([
-        acc, ', ', this.conditionalLink(reference.label, reference.href, { target: '_blank', key: i + 1 })
-    ]), this.conditionalLink(references[0].label, references[0].href, { target: '_blank', key: 0 }));
+        acc, i !== 0 ? ', ' : '', this.conditionalLink(reference.label, reference.href, { target: '_blank', key: i + 1 })
+    ]), []);
 
     calculateChild = ({ description, rationale, identifier, references }, key) => ({
         parent: key * 2,
@@ -520,7 +520,7 @@ class SystemRulesTable extends React.Component {
                                         updateFilter={ this.updateFilter } />
                                     <SimpleTableFilter buttonTitle={ null }
                                         onFilterChange={ this.handleSearch }
-                                        placeholder="Search by name or identifier" />
+                                        placeholder="Search by name, identifer, or reference" />
                                 </InputGroup>
                             </LevelItem>
                             <LevelItem>
--- a/packages/inventory-compliance/src/SystemRulesTable.js
+++ b/packages/inventory-compliance/src/SystemRulesTable.js
@@ -3,9 +3,8 @@ import propTypes from 'prop-types';
 import ComplianceRemediationButton from './ComplianceRemediationButton';
 import { CheckIcon, CheckCircleIcon, ExclamationCircleIcon } from '@patternfly/react-icons';
 import { SimpleTableFilter, TableToolbar } from '@redhat-cloud-services/frontend-components';
-import { Table, TableHeader, TableBody, sortable, SortByDirection } from '@patternfly/react-table';
+import { Table, TableHeader, TableBody, SortByDirection } from '@patternfly/react-table';
 import {
-    Checkbox,
     InputGroup,
     Level,
     LevelItem,
@@ -32,30 +31,25 @@ import {
     SEVERITY_COLUMN,
     POLICY_COLUMN,
     TITLE_COLUMN,
-    ANSIBLE_ICON,
     HIGH_SEVERITY,
     MEDIUM_SEVERITY,
     LOW_SEVERITY
 } from './Constants';
 
 class SystemRulesTable extends React.Component {
-    constructor(props) {
-        super(props);
-        this.state = {
-            columns: props.columns,
-            page: 1,
-            itemsPerPage: props.itemsPerPage,
-            rows: [],
-            hidePassed: props.hidePassed,
-            severity: [],
-            policy: [],
-            currentRows: [],
-            refIds: {},
-            profiles: {},
-            sortBy: {}
-        };
-
-    }
+    state = {
+        columns: this.props.columns,
+        page: 1,
+        itemsPerPage: this.props.itemsPerPage,
+        rows: [],
+        hidePassed: this.props.hidePassed,
+        severity: [],
+        policy: [],
+        currentRows: [],
+        refIds: {},
+        profiles: {},
+        sortBy: {}
+    };
 
     removeRefIdPrefix = (refId) => {
         const splitRefId = refId.toLowerCase().split('xccdf_org.ssgproject.content_profile_')[1];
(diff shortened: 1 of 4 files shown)


packages/inventory-compliance/src/__snapshots__/SystemRulesTable.test.js.snap
```diff
@@ -30,7 +30,7 @@ exports[`SystemRulesTable component should render 1`] = `
             onButtonClick={[Function]}
             onFilterChange={[Function]}
             onOptionSelect={[Function]}
-            placeholder="Search by name or identifer"
+            placeholder="Search by name, identifer, or reference"
```
akofink: This will get cut off on narrow screens. It was already being cut off before this change. @katierik I think this is a UX issue

akofink:
Actually, @dLobatog I'm seeing this graphql response and this UI with duplicate references for some reason:

```
{
  "title": "Verify that All World-Writable Directories Have Sticky Bits Set",
  "severity": "low",
  "rationale": "Failing to set the sticky bit on public directories allows unauthorized users to delete files in the directory structure. The only authorized public directories are those temporary directories supplied with the system, or those designed to be temporary file repositories.  The setting is normally reserved for directories used by the system, by users for temporary file storage (such as /tmp), and for directories requiring global read/write access.",
  "refId": "xccdf_org.ssgproject.content_rule_dir_perms_world_writable_sticky_bits",
  "description": "When the so-called 'sticky bit' is set on a directory, only the owner of a given file may remove that file from the directory. Without the sticky bit, any user with write access to a directory may remove any file in the directory. Setting the sticky bit prevents users from removing each other's files. In cases where there is no reason for a directory to be world-writable, a better solution is to remove that permission rather than to set the sticky bit. However, if a directory is used by a particular application, consult that application's documentation instead of blindly changing modes. To set the sticky bit on a world-writable directory DIR, run the following command: $ sudo chmod +t DIR",
  "compliant": true,
  "remediationAvailable": false,
  "references": [
    {
      "label": "AC-6",
      "href": "http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-53r4.pdf",
      "__typename": "RuleReference"
    },
    {
      "label": "Test attestation on 20120929 by swells",
      "href": "https://github.com/OpenSCAP/scap-security-guide/wiki/Contributors",
      "__typename": "RuleReference"
    }
  ],
  "identifier": {
    "label": "CCE-26840-9",
    "system": "http://cce.mitre.org",
    "__typename": "RuleIdentifier"
  },
  "__typename": "Rule"
}
```

image

Any idea what might be causing this?

packages/inventory-compliance/src/__snapshots__/SystemRulesTable.test.js.snap
katierik: Ick, that is no good. I don't have an immediate solution for this with the filter. The attribute value really isn't set up for multiple things.
We could break it into 3 filters: name; identifiers; references; but that is non-ideal. 
I'll bring this up with PF and also see if I can come up with a better solution.

dLobatog:
Let's wait a sec for this one - I want to figure out the double references showing up, and also it appears to make the GraphQL request timeout...

dLobatog:
Indeed the duplication comes from the API:
```
[2] pry(#<Types::Rule>)> object.references.pluck(selected_columns).map do |reference|
[2] pry(#<Types::Rule>)*   selected_columns.zip([reference].flatten).to_h
[2] pry(#<Types::Rule>)* end
DEPRECATION WARNING: Dangerous query method (method whose arguments are used as raw SQL) called with non-attribute argument(s): [:label, :href]. Non-attribute arguments will be disallowed in Rails 6.0. This method should not be called with user-provided values, such as request parameters or model attributes. Known-safe values can be passed by wrapping them in Arel.sql(). (called from references at (pry):3)
   (18.9ms)  SELECT label, href FROM "rule_references" INNER JOIN "rule_references_rules" ON "rule_references"."id" = "rule_references_rules"."rule_reference_id" WHERE "rule_references_rules"."rule_id" = $1 ORDER BY "rule_references"."label" ASC  [["rule_id", "7fa6c199-ff7f-4d75-b608-ab5465675d97"]]
  ↳ (pry):3
=> [{:label=>"11", :href=>"https://www.cisecurity.org/wp-content/uploads/2017/03/Poster_Winter2016_CSCs.pdf"},
 {:label=>"11", :href=>"https://www.cisecurity.org/wp-content/uploads/2017/03/Poster_Winter2016_CSCs.pdf"},
 {:label=>"14", :href=>"https://www.cisecurity.org/wp-content/uploads/2017/03/Poster_Winter2016_CSCs.pdf"},
 {:label=>"14", :href=>"https://www.cisecurity.org/wp-content/uploads/2017/03/Poster_Winter2016_CSCs.pdf"},
 {:label=>"3", :href=>"https://www.cisecurity.org/wp-content/uploads/2017/03/Poster_Winter2016_CSCs.pdf"},
 {:label=>"3", :href=>"https://www.cisecurity.org/wp-content/uploads/2017/03/Poster_Winter2016_CSCs.pdf"},
 {:label=>"4.3.3.5.1", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.5.1", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.5.2", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.5.2", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.5.3", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.5.3", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.5.4", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.5.4", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.5.5", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.5.5", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.5.6", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.5.6", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.5.7", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.5.7", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.5.8", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.5.8", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.6.1", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.6.1", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.6.2", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.6.2", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.6.3", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},
 {:label=>"4.3.3.6.3", :href=>"https://www.isa.org/templates/one-column.aspx?pageid=111294&productId=116731"},

```

akofink:
@dLobatog This drastically reduces the response time of this API. I'm concerned that we don't have any UI handling of the graphql timeout nor the gateway timeout. IMO we should really figure out how to make this quicker, but before we do, we need the defensive frontend in place.

dLobatog:
@akofink Absolutely - I don't want to merge this until it's clear the API can respond fast enough

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -239,8 +239,8 @@ class SystemRulesTable extends React.Component {
     )
 
     referencesList = (references) => references.reduce((acc, reference, i) => ([
-        acc, ', ', this.conditionalLink(reference.label, reference.href, { target: '_blank', key: i + 1 })
-    ]), this.conditionalLink(references[0].label, references[0].href, { target: '_blank', key: 0 }));
+        acc, i !== 0 ? ', ' : '', this.conditionalLink(reference.label, reference.href, { target: '_blank', key: i + 1 })
```
karelhala: Is this correct? Beacause if I try it on sample array it looks kinda strange
```JS
['first', 'second', 'third'].reduce((acc, ref, i) => ([acc, i !== 0 ? ', ' : '', ref]), [])
// output is
[
 [ [ [], '', 'first' ], ', ', 'second' ],
 ', ',
 'third'
]
```
I guess you want to have flat array
```JS
referencesList = (references) => references.reduce((acc, reference, i) => ([
  ...acc,
  this.conditionalLink(reference.label, reference.href, { target: '_blank', key: i + 1 })
])
```

karelhala: @dLobatog ^

packages/inventory-compliance/src/__snapshots__/SystemRulesTable.test.js.snap
katierik: FOLLOW UP: For now, we can increase the max width of the filter box to see more of the content

packages/inventory-compliance/src/SystemRulesTable.js
dLobatog: Fixed, thanks! I guess this wasn't noticed because on the UI once it's cast to a string you don't see it

PR #278, opened 2019-11-18
Use tags from API instead of mocked ones

karelhala:
TODO:
* [ ] https://github.com/RedHatInsights/insights-host-inventory/issues/517

### UI changes
Screenshot from 2019-11-20 16-29-59


packages/inventory/src/api/index.js
```diff
@@ -24,12 +24,28 @@ export const mapData = ({ facts = {}, ...oneResult }) => ({
...
 });
 
+export const mapTags = (data = { results: [] }, { orderBy, orderDirection } = {}) => {
+    if (data.results.length > 0) {
+        return hosts.apiHostGetHostTags(data.results.map(({ id }) => id), data.per_page, 1, orderBy, orderDirection)
+            .then(({ results: tags }) => ({
+                ...data,
+                results: data.results.map(row => ({ ...row, tags: tags[row.id] || [] }))
+            }))
+            .catch(() => ({
```
karelhala: This is safe way how not to fail on fetching tags (they are something trivial and shouldn't break the UI)

karelhala:
@yashnalla mhhh, that is trange. I can't seem to be able to find any reference to it in lock file https://raw.githubusercontent.com/RedHatInsights/frontend-components/d14a0e3b12899a3922cd5737445d3d46c7069d4c/package-lock.json

yashnalla:
Yea i noticed that too; I didn't see it when I was npm installing; just npm linking. No idea why. Might just be some npm behavior I don't know about.

PR #280, opened 2019-11-18
Disable Endpoint step and configuration

rvsia:
**Description**

- allow to disable endpoint step (and so authentication) by using  `skipEndpoint` in hardcoded schemas
- refactors createSource to use async/await for better maintainability
- updates DDF to be able to use initiliazeOnMount


rvsia:
@Hyperkid123

PR #283, opened 2019-11-25
feat(Openshift): Add openshift steps for cost management

boaz0:
closes https://github.com/RedHatInsights/sources-ui/issues/127
//cc @rvsia

### DIFF
--- a/packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costManagement.js
+++ b/packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costManagement.js
@@ -1,0 +1,243 @@
+import { TextContent, Text, TextVariants, TextListItem, TextList, Button, ClipboardCopy, Popover, ButtonVariant } from '@patternfly/react-core';
+import { QuestionCircleIcon } from '@patternfly/react-icons';
+import React from 'react';
+
+const OPENSHIFT_INSTALLATION_URL = 'https://docs.openshift.com/container-platform/4.2/getting_started/install_openshift.html';
+const OPERATOR_METERING_INSTALLATION_URL = 'https://docs.openshift.com/container-platform/4.2/metering/metering-installing-metering.html';
+const ANSIBLE_INSTALLATION_URL = 'https://access.redhat.com/articles/3174981';
+const EPEL_INSTALLATION_URL = 'https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html#basics-what-will-be-installed';
+const OC_CMD_URL = 'https://docs.openshift.com/container-platform/4.2/cli_reference/openshift_cli/getting-started-cli.html#cli-installing-cli_cli-developer-commands';
+const USAGE_COLLECTOR_URL = 'https://github.com/project-koku/korekuta/archive/master.zip';
+const OPERATOR_METERING_URL = 'https://github.com/operator-framework/operator-metering/blob/master/Documentation/metering-config.md';
+const REPORTING_OPERATOR_URL = 'https://docs.openshift.com/container-platform/4.2/metering/configuring-metering/metering-configure-reporting-operator.html';
+
+export const PrerequisiteDescription = () => (
+    <TextContent key='description'>
+        <Text component={TextVariants.p}>
+            Before you begin, install the following prerequisites if not present:
+        </Text>
+    </TextContent>
+);
+
+export const PrerequisiteOCPText = () => (
+    <TextContent key='prerequisite-ocp'>
+        <Text component={TextVariants.p}>
+            On your OpenShift cluster, install:
+        </Text>
+    </TextContent>
+);
+
+export const PrerequisiteSystemText = () => (
+    <TextContent key='prerequisite-system'>
+        <Text component={TextVariants.p}>
+            On a system with network access to your OpenShift cluster, install:
+        </Text>
+    </TextContent>
+);
+
+export const PrerequisiteOCPList = () => (
+    <TextContent className='list-align-top'>
+        <TextList component='ul'>
+            <TextListItem component='li'>
+                <Text
+                    component={TextVariants.a}
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    href={OPENSHIFT_INSTALLATION_URL}
+                >
+                    OpenShift Container Platform 4.2 or newer
+                </Text>
+            </TextListItem>
+            <TextListItem component='li'>
+                <Text
+                    component={TextVariants.a}
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    href={OPERATOR_METERING_INSTALLATION_URL}
+                >
+                        Operator Metering
+                </Text>
+            </TextListItem>
+        </TextList>
+    </TextContent>
+);
+
+export const PrerequisiteSystemList = () => (
+    <TextContent className='list-align-top'>
+        <TextList component='ul'>
+            <TextListItem component='li'>
+                <Text
+                    component={TextVariants.a}
+                    rel="noopener noreferrer"
+                    target="_blank"
+                    href="htps://access.redhat.com/products/red-hat-insights#getstarted"
+                >
+                    Red Hat Insights Client
+                </Text>
+            </TextListItem>
+            <TextListItem component='li'>
+                <Text
+                    component={TextVariants.a}
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    href={ANSIBLE_INSTALLATION_URL}
+                >
+                    Ansible
+                </Text>
+                &nbsp;and the&nbsp;
+                <Text
+                    component={TextVariants.a}
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    href={EPEL_INSTALLATION_URL}
+                >
+                    EPEL repository
+                </Text>
+            </TextListItem>
+            <TextListItem component='li'>
+                <Text
+                    component={TextVariants.a}
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    href={OC_CMD_URL}
+                >
+                    OpenShift command line tools (oc)
+                </Text>
+            </TextListItem>
+        </TextList>
+    </TextContent>
+);
+
+export const ObtainLoginDescription = () => (
+    <TextContent>
+        <Text component={TextVariants.p}>
+            To gather OpenShift Container Platform metering data, Cost Management requires the login token for the Reporting Operator service account.
+            <Popover
+                aria-label="more-information-popover"
+                position="top"
+                bodyContent={
+                    <TextContent>
+                        When you install Operator Metering for Red Hat OpenShift Container Platform, the&nbsp;
+                        <i>reporting-operator</i> service account is created in the Reporting Operator namespace.&nbsp;
+                        <a href={REPORTING_OPERATOR_URL} target="_blank" rel="noopener noreferrer">Learn more</a>
+                    </TextContent>
+                }
+            >
+                <Button variant={ButtonVariant.plain}><QuestionCircleIcon /></Button>
+            </Popover>
+        </Text>
+        <TextContent className='list-align-left'>
+            <TextList component='ol'>
+                <TextListItem component='li'>
+                    Log in to the Red Hat OpenShift Container Platform cluster with an account
+                        that has access to the Reporting Operator namespace.
+                </TextListItem>
+                <TextListItem component='li'>
+                    To obtain your login token, run:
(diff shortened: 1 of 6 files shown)


boaz0:
@dccurtis - I get this error message

image

does the backend support OCP?

dccurtis:
@boaz0 `billing_source` is not used for OCP.  There are no cost management specific fields for OCP.  The only thing you will have to do is to send the cluster_id to platform sources into the `source_ref` field of the `/sources` API [Sources API soure_ref](https://github.com/RedHatInsights/sources-api/blob/07044a23797cd2c811e2af590edae80bf2a4efbc/public/doc/openapi-3-v1.0.0.json#L1539)

boaz0:
Thanks @dccurtis - I will do that.

boaz0:
OK - tested that passing the cluster id in the "source_ref" field works and it looks OK.

- This is the POST request sent to https://ci.foo.redhat.com:1337/api/sources/v1.0/graphql

```
{
  "query": "{ sources(filter: {name: \"boaz-shuster-test-source\"})\n        { id, name }\n    }"
}
```

- This is the POST request that was sent to https://ci.foo.redhat.com:1337/api/sources/v1.0/sources:

```
{
  "name": "boaz-shuster-test-source",
  "source_type_id": "1"
}
```

- This is the POST request sent to https://ci.foo.redhat.com:1337/api/sources/v1.0/applications

```
{
  "application_type_id": "2",
  "source_id": "1059"
}
```

//cc @rvsia @dccurtis feel free to review the code

boaz0:
//cc @nlcwong can you review this too?

image

image

image

image

image

dccurtis:
@boaz0 great progress!  I see your source in the CI logs, but I think you forgot to create the endpoint on the platform sources side. I logged this `ERROR: Unable to find endpoint for Source ID: 1059`

The flow for platform source creation should be:
This is just the gist of what you need to do.  I don't know specifically what the UI is setting @rvsia could help to keep me honest here.
1. Create Source `POST /api/sources/v1.0/sources` with data `{'source_type_id': <OCP Type ID>, 'name': source_name, 'source_ref': <cluster id>}`
2. Create Endpoint `POST /api/sources/v1.0/endpoints` with data `{'host': 'www.example.com', 'path': '/api/v1', 'source_id': <source_id>}`. Returns <endpoint_id> for next step
3. Create Authentication `POST /api/sources/v1.0/authentications` with data `{'authtype': 'tenant_id_client_id_client_secret', 'name': 'Azure default', 'password': <client_secret>,
                     'status': 'valid', 'status_details': 'Details Here', 'username': <client_id>,
                     'extra': {'azure': {'tenant_id': <tenant_id>}}, 'resource_type': 'Endpoint',
                     'resource_id': <endpoint_id>}`

packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costMangement.js
```diff
@@ -0,0 +1,222 @@
...
+    <TextContent key='prerequisite-system'>
+        <Text component={TextVariants.p}>
+            On a system with network access to your OpenShift cluster, install:
+        </Text>
+    </TextContent>
+);
+
+export const PrerequisiteOCPList = () => (
+    <TextContent>
+        <TextList component='ul' style={{ marginLeft: '16px', marginTop: '-16px' }}>
```
rvsia: Please, dont use inline styles, you can use PF4 class names https://www.patternfly.org/v4/documentation/core/utilities/spacing and it is strange to overrides default margins in a standard list. It should look according to default Patternfly design.

packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costMangement.js
```diff
@@ -0,0 +1,222 @@
...
+            </TextListItem>
+            <TextListItem component='li'>
+                <Text
+                    component={TextVariants.a}
+                    target="_blank"
+                    href="https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html#basics-what-will-be-installed"
+                >
+                    Ansible
+                </Text>
+                {' '}and the{' '}
```
rvsia: Can we someway do it without blocks with empty strings? At least we could use html constants for spaces

packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costMangement.js
```diff
@@ -0,0 +1,222 @@
...
+                    When configuration is complete, you will receive a confirmation message that contains a cluster identifier.<br />
+                    Enter that value in the cluster identifier field.
+                </TextListItem>
+            </TextList>
+        </TextContent>
+    </TextContent>
+);
+
+export const ClusterIdentifierLabel = () => (
+    <>
```
rvsia: Please, dont use the shorthand. Shouldnt it be a Text?

packages/sources/src/addSourceWizard/hardcodedSchemas.js
```diff
@@ -7,6 +7,8 @@ import * as OpenshiftToken from './hardcodedComponents/openshift/token';
 import * as AwsSecret from './hardcodedComponents/aws/access_key';
 import * as AwsArn from './hardcodedComponents/aws/arn';
 
+import * as CMOpenshift from  './hardcodedComponents/openshift/costMangement';
```
rvsia: Typo in costManagement file name

rvsia:
@dccurtis There was an additional step with setting data for the endpoint, but I was told that it does not need to have an endpoint, so changes was made to allow disabling endpoint’s steps and its creation (+ it disables authentication creation as well.)

dccurtis:
@rvsia the authentication may not be needed for the platform but it is needed for cost management. It’s how we will get the client and tenant secrets.

dccurtis:
@rvsia Question on the AWS wizard page.  It looks like an endpoint is getting created automatically by the UI (even though there is no page for the user to add a URL).  I think we will have to do the same thing for OCP since cost mgmt will need the values out of the authentication object and the endpoint is the glue between a source and authentication.

Does that seem right?

rvsia:
@dccurtis It is not so automatic as it seems, all endpoints are defined here https://github.com/RedHatInsights/sources-api/blob/45046bf46c946ab486ce9a1f3a8dbc74d263287c/db/seeds/source_types.rb#L19 and if the endpoint is `:hidden => true`, then the field and values are hidden in the first authentication step.

However, it is possible to do it, even in this PR without any platform changes, however I see one problem which can appear - the URL is required (not in the seeds, but that's probably just a missing attribute) - and if we created an empty endpoint for the source and an user decided to edit this source, he would see an empty and required field in the form and it would block submitting this form. Of course it's a problem I can solve, but I would rather like to avoid it. (As I asked (when we started working on CM/Sources integration) if application can change the endpoint schema and I was told that it cannot.)

So, by my standpoint, the best solution would be to integrate the endpoint step into the flow, but if it is a problem, we can do it the other way.

@boaz0 If you decide to do that by creating an empty endpoint, just append a few fields to the first additional step of your flow:

```jsx
{
  name: 'noEndpoint', // this name is disabling sending values `endpoint.*`, `authentication.*` if its value is true
  hideField: true, // will hide the field in the form
  initializeOnMount: true, // will always rewrite the value
  initialValue: false, // it will be probably rewritten to 'undefined', but that also should work,
  component: "text-field"
},{
  component: "text-field", 
  name: "endpoint.role",
  hideField: true,
  initializeOnMount: true,
  initialValue: "kubernetes", // ? probably, don't know
}
```

@dccurtis However, I don't see any values which are sent to the authentication in this flow.

dccurtis:
@boaz0 @rvsia So sorry for the confusion here!  I was thinking of Azure when I was making the comments on the authentication object being needed.  You are correct, for OCP Cost Management doesn't need the authentication and therefore doesn't need the endpoint.

That being said.  I'm going to have to make a small change on the backend since I was operating under the assumption that there would always be an endpoint for a source.

rvsia:
@dccurtis Awesome!

nlcwong:
Sorry, last minute changes...

Step 3 - change step title from "Prerequisite installation" to "Install prerequisites", this way it conforms to the other step titles of the action being first and then the noun.

Step 4 - change the bulleted list into a numbered list

Step 5 - change the bulleted list into a numbered list; capitalize "cost" to "Cost"

Step 6 - change the bulleted list into a numbered list

Step 7 - the Cluster ID that the user entered into Step 5 is missing in the review step

boaz0:
@nlcwong done
@rvsia added tests

packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costManagement.js
```diff
@@ -0,0 +1,222 @@
...
+    <TextContent key='prerequisite-system'>
+        <Text component={TextVariants.p}>
+            On a system with network access to your OpenShift cluster, install:
+        </Text>
+    </TextContent>
+);
+
+export const PrerequisiteOCPList = () => (
+    <TextContent>
+        <TextList component='ul' style={{ marginLeft: global_spacer_md.value, marginTop: `-${global_spacer_md.value}` }}>
```
rvsia: Could we extract this into a classname and define it in css? It would be better to have all styles together.

boaz0: You want me to create `costManagement.styles.js` and use react-styles's css for that?

rvsia: Nope, here https://github.com/RedHatInsights/frontend-components/tree/master/packages/sources/src/styles you can find styles, so add a new css file, add your styles there and then import this file in index.js https://github.com/RedHatInsights/frontend-components/blob/201dfcd4818179a4fe887d698ea92d4e089be5fd/packages/sources/src/index.js#L13, so it will be exported together with the wizard.

boaz0: Thanks that was helpful

packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costManagement.js
```diff
@@ -0,0 +1,222 @@
...
+
+export const PrerequisiteSystemList = () => (
+    <TextContent className='list-align-top'>
+        <TextList component='ul'>
+            <TextListItem component='li'>
+                <Text component={TextVariants.a} target="_blank" href="htps://access.redhat.com/products/red-hat-insights#getstarted" >Red Hat Insights Client</Text>
+            </TextListItem>
+            <TextListItem component='li'>
+                <Text
+                    component={TextVariants.a}
```
karelhala: You should also add `rel="noopener noreferrer"` to all `a` elements.

packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costManagement.js
```diff
@@ -0,0 +1,222 @@
...
+                    target="_blank"
+                    href="https://access.redhat.com/articles/3174981"
```
karelhala: Can you make these URLs into contants?

packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costManagement.js
rvsia: @boaz0 Can you please fix this? Thanks!

packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costManagement.js
rvsia: @boaz0 It's up to you, no need to fix it.

packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costManagement.js
boaz0: Sure!

packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costManagement.js
boaz0: Will do

packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costManagement.js
boaz0: Done!

packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costManagement.js
boaz0: Done!

PR #291, opened 2019-12-02
Update to support additional adding of apps

rvsia:
Needed for https://github.com/RedHatInsights/sources-ui/pull/168

This PR updates exported components to support adding of applications to sources in UI

- Imports additional functions as helpers
- Allows to disable password requirements (in UI, we cannot pre-filled them as these values are not obtainable by API)
- Handling undefined/null values in Summary screen

TODO
- [x] disable password requirements


packages/sources/src/addSourceWizard/schemaBuilder.js
```diff
@@ -25,13 +25,24 @@ export const getAdditionalStepFields = (fields, stepKey) => fields.filter(field
 
 export const getNoStepsFields = (fields, additionalStepKeys = []) => fields.filter(field => !field.stepKey || additionalStepKeys.includes(field.stepKey));
 
-export const injectAuthFieldsInfo = (fields, type, auth, applicationName) => fields.map((field) => {
+export const removeRequiredValidator = (validate = []) =>
```
Hyperkid123: Why is this here? Is it possible to use the form-level validation and have the login there instead of doing this.

rvsia: Not sure how. I need to disable required validators on password values when editing (aka attaching application to a source which already has the right credentials) because the password is not obtainable by the API.

Hyperkid123: Can you make the field conditional?

rvsia: But user should be able to change if he wants to do it. (HelperText is also added with a description.)

It's the same flow as the adding, just with pre-filled values.

Hyperkid123:
Theoretically, you could prefill it with `null` but then you would have to check the value before posting it.

PR #296, opened 2019-12-05
Add a tag into inv table instead of div

karelhala:
### Fixes
When user right clicks on host name option to open detail in new tab is present. When users uses meta key detail is opened in new tab.

### JIRA
RHCLOUD-3577

### DIFF
--- a/packages/inventory/src/EntityTable.js
+++ b/packages/inventory/src/EntityTable.js
@@ -28,8 +28,13 @@ class EntityTable extends React.Component {
         const { match: { url }, history, onDetailSelect, loaded } = this.props;
         if (loaded) {
             const dilimeter = url.substr(-1, 1) === '/' ? '' : '/';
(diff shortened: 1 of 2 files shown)


packages/inventory/src/EntityTable.js
```diff
@@ -28,8 +28,13 @@ class EntityTable extends React.Component {
         const { match: { url }, history, onDetailSelect, loaded } = this.props;
         if (loaded) {
             const dilimeter = url.substr(-1, 1) === '/' ? '' : '/';
-            history.push(`${url}${dilimeter}${key}/${application ? application : ''}`);
-            onDetailSelect && onDetailSelect(application);
+            const isMetaKey = (event.ctrlKey || event.metaKey || event.which === 2);
```
ryelo: We should provide this as a utility in the future 😄

PR #299, opened 2019-12-09
Download button to support disabled prop

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-2883 kinda

Icon dropdowns don't support disabled flag, next best thing is to matriculate that down to the dropdown items 🤷‍♂


AllenBW:
pull request labeler says....
` ##[error]HttpError: Resource not accessible by integration
##[error]Resource not accessible by integration
##[error]Node run failed with exit code 1`

karelhala:
@AllenBW yeah labeler is broken because it does not support forks

PR #300, opened 2019-12-09
Add count of issues and systems to remediations wizard

karelhala:
### Before
Screenshot from 2019-12-09 15-57-33

### After
Screenshot from 2019-12-09 15-55-38

### JIRA
RHCLOUD-3622

### DIFF
--- a/packages/remediations/src/RemediationButton.js
+++ b/packages/remediations/src/RemediationButton.js
@@ -57,7 +57,7 @@ RemediationButton.propTypes = {
 RemediationButton.defaultProps = {
     isDisabled: false,
     onRemediationCreated: f => f,
-    children: 'Remediate with Ansible'
+    children: 'Plan remediation'
 };
 
 export default RemediationButton;
--- a/packages/remediations/src/RemediationButton.js
+++ b/packages/remediations/src/RemediationButton.js
@@ -57,7 +57,7 @@ RemediationButton.propTypes = {
 RemediationButton.defaultProps = {
     isDisabled: false,
     onRemediationCreated: f => f,
-    children: 'Plan remediation'
+    children: 'Remediate with Ansible'
 };
 
 export default RemediationButton;
--- a/packages/remediations/src/RemediationWizard.js
+++ b/packages/remediations/src/RemediationWizard.js
@@ -314,7 +314,7 @@ class RemediationWizard extends Component {
             <Wizard
                 width="initial"
                 isValidated={ this.state.valid && (this.state.nameValid || !this.state.isNewSwitch) }
-                title="Plan remediation"
+                title="Remediate with Ansible"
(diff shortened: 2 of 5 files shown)


packages/remediations/src/steps/ExistingOrNew.js
```diff
@@ -12,95 +11,136 @@ import {
...
+            <span>
+                    You have selected&nbsp;
+                <h1 className='ins-m-text__bold'>
+                    {data.systems.length} {pluralazie('system', data.systems.length)}
+                </h1>
+                    &nbsp;with&nbsp;
+                <h1 className='ins-m-text__bold'>
+                    {data.issues.length} {pluralazie('issue', data.issues.length)}
+                </h1>
+                    &nbsp;to add to a remedation plan.
```
jharting: The mocks modify terminology but that is not in effect currently. The old terminology should be used as described in https://marvelapp.com/j3b9b7e/screen/63083483
I.e., should say "Playbook" instead of "remediation plan"
Sorry for confusion.

packages/remediations/src/RemediationWizard.js
```diff
@@ -312,9 +312,9 @@ class RemediationWizard extends Component {
 
         return (
             <Wizard
-                isLarge = { true }
+                width="initial"
                 isValidated={ this.state.valid && (this.state.nameValid || !this.state.isNewSwitch) }
-                title="Remediate with Ansible"
+                title="Plan remediation"
```
jharting: The mocks modify terminology but that is not in effect currently. The old terminology should be used as described in https://marvelapp.com/j3b9b7e/screen/63083483
I.e., this title should not change. Sorry for confusion.

packages/remediations/src/RemediationButton.js
```diff
@@ -57,7 +57,7 @@ RemediationButton.propTypes = {
 RemediationButton.defaultProps = {
     isDisabled: false,
     onRemediationCreated: f => f,
-    children: 'Remediate with Ansible'
+    children: 'Plan remediation'
```
jharting: The mocks modify terminology but that is not in effect currently. The old terminology should be used as described in https://marvelapp.com/j3b9b7e/screen/63083483
I.e., this title should not change. Sorry for confusion.

packages/remediations/src/utils.js
```diff
@@ -14,3 +14,5 @@ export function getGroup () {
 export function remediationUrl (id) {
     return urijs(document.baseURI).segment(getGroup()).segment('remediations').segment(id).toString();
 }
+
+export const pluralazie = (text = '', count = 0) => `${text}${count !== 1 ? 's' : ''}`;
```
Hyperkid123: Maybe we should start using react-intl for plurals? It is in this package anyway.

PR #310, opened 2019-12-21
Hook tag filtering to real API

karelhala:
### Enhancements
* [x] hook tags fetch to real API
* [x] add filter by selected tags
### UI changes
* [x] tags typeahead
* [x] add loading state to tags filter
* [x] add no tags when no tags are present
* [x] add `${x} more tags` at the end of filter tags list when more than curr page of tags is present in DB

### UI changes
* Typeahead (I had not actual data so the string is not filtered, this is just to show interaction when user types in something. Also the loading styling is a bit off please refer to `Tags loading` point how it actually looks like)
Peek 2019-12-27 20-57

* No tags
Screenshot from 2019-12-27 21-12-27

* Tags loading
Screenshot from 2019-12-27 21-06-54

* More tags available
Screenshot from 2019-12-27 21-06-19

_Used data were placeholder data mocked based on API_

#### JIRA
RHCLOUD-3736
RHCLOUD-3746

### DIFF
--- a/packages/inventory/src/EntityTableToolbar.js
+++ b/packages/inventory/src/EntityTableToolbar.js
@@ -20,15 +20,16 @@ import flatMap from 'lodash/flatMap';
 class ContextEntityTableToolbar extends Component {
     state = {
         textFilter: '',
-        selected: {}
+        selected: {},
+        filterTagsBy: ''
     }
 
     debouncedRefresh = debounce((config) => {
         this.props.onRefreshData && this.props.onRefreshData(config);
     }, 800);
 
-    debounceGetAllTags = debounce((config) => {
-        this.props.getAllTags && this.props.getAllTags(config);
+    debounceGetAllTags = debounce((config, options) => {
+        this.props.getAllTags && this.props.getAllTags(config, options);
     }, 800);
 
     componentDidMount() {
@@ -63,19 +64,22 @@ class ContextEntityTableToolbar extends Component {
         }
 
         const refresh = debounced ? this.debouncedRefresh : onRefreshData;
+        const newFilters = [
+            ...filters,
+            { tagFilters }
+        ];
         refresh({
             page: 1,
             perPage,
-            filters: [
-                ...filters,
-                { tagFilters }
-            ]
+            filters: newFilters
         });
+
+        return newFilters;
     }
 
     createTagsFilter = () => {
-        const { allTags, allTagsLoaded, additionalTagsCount, getAllTags } = this.props;
-        const { selected } = this.state;
+        const { allTags, allTagsLoaded, additionalTagsCount, getAllTags, filters } = this.props;
+        const { selected, filterTagsBy } = this.state;
         return {
             label: 'Tags',
             value: 'tags',
@@ -83,7 +87,11 @@ class ContextEntityTableToolbar extends Component {
             placeholder: 'Filter system by tag',
             filterValues: {
                 className: 'ins-c-inventory__tags-filter',
-                onFilter: (value) => this.debounceGetAllTags(value),
+                onFilter: (value) => {
+                    this.setState({ filterTagsBy: value }, () => {
+                        this.debounceGetAllTags(value, { filters });
+                    });
+                },
                 onChange: (_e, newSelection, group, item, groupKey, itemKey) => {
                     const isSelected = newSelection[groupKey][itemKey];
                     newSelection[groupKey][itemKey] = {
@@ -93,8 +101,10 @@ class ContextEntityTableToolbar extends Component {
                     };
                     this.setState(
                         { selected: newSelection },
-                        () => this.applyTags(newSelection)
-                    );
+                        () => {
+                            const newFilter = this.applyTags(newSelection);
+                            getAllTags(filterTagsBy, { filters: newFilter });
+                        });
                 },
                 selected,
                 ...allTagsLoaded && allTags.length > 0 ? {
@@ -128,8 +138,8 @@ class ContextEntityTableToolbar extends Component {
     }
 
     constructFilters = () => {
-        const { perPage, onRefreshData, onClearFilters, activeFiltersConfig } = this.props;
-        const { selected, textFilter } = this.state;
+        const { perPage, onRefreshData, onClearFilters, activeFiltersConfig, getAllTags } = this.props;
+        const { selected, textFilter, filterTagsBy } = this.state;
         return {
             filters: [
                 ...mapGroups(selected, 'chips'),
@@ -149,6 +159,8 @@ class ContextEntityTableToolbar extends Component {
                     this.setState({
                         selected: {},
                         textFilter: ''
+                    }, () => {
+                        getAllTags(filterTagsBy, {});
                     });
                 } else {
                     if (deleted.type === TEXTUAL_CHIP) {
@@ -156,7 +168,10 @@ class ContextEntityTableToolbar extends Component {
                     } else if (deleted.type === TAG_CHIP) {
                         const deletedItem = deleted.chips[0];
                         selected[deleted.key][deletedItem.key] = false;
-                        this.setState({ selected }, () => this.applyTags(selected, false));
+                        this.setState({ selected }, () => {
+                            const newFilter = this.applyTags(selected, false);
+                            getAllTags(filterTagsBy, { filters: newFilter });
+                        });
                     }
                 }
 
@@ -297,6 +312,6 @@ function mapStateToProps(
 }
 
 export default connect(mapStateToProps, (dispatch) => ({
-    getAllTags: (search) => dispatch(fetchAllTags(search)),
+    getAllTags: (search, options) => dispatch(fetchAllTags(search, options)),
     onClearFilters: () => dispatch(clearFilters())
 }))(EntityTableToolbar);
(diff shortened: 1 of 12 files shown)


KendraMar:
Great progress, Karel (and thanks for the videos...helps a lot). Some questions:

**Issue A:**

In the first type ahead video, I would've expected to see the animation work this way:

1. user clicks in the filter input field
2. user starts to type
3. filter input field shows a drop down with the loading spinner (if necessary for performance)
4. drop down then replaces the loading spinner with a checkbox item that meets the filter type ahead criteria

What i'm seeing now is that #4 above happens right after #1 above.

Issue B:

- Are we showing a tooltip over the badge to tell users that there are 10 items in the data set (is that the entire data set or just on this page, btw?). BTW, we are filtering hte entire data set, right -- so not just the data loaded on the page?

**Issue C (more of a confirmation/question)**

The checkbox that appears in the drop down implies that the user can filter the entire data set for more than one term at a time? Cool if so!

karelhala:
>Issue A:

Yes, that is what you see. The API returns first number of tags, it feels a bit more fluid to show some values and allow typeahead if user wants to pick specific tag. But that was my personal feeling. If you'd prefer having message like `Start typing to filter tags` that would be cool as well.

> Issue B:

Good point. The number is actually to indicate how many systems has this tag. And as you drill down the number gets lower. I'll add a tooltip there that says `Applicable to ${number} of systems.`

As of the colors, the blue one is for tags that are applicable to some systems (meaning if user selects this tag something will show up in table) and gray if there will be no systems shown in the table.

> Issue C:

Yes, user can select multiple tags in order to drill down more specifically.

> Issue D:

Sure thing. I aligned it to center, because it felt kinda better - this is how it looks like left aligned 
Screenshot from 2020-01-03 15-54-36

> Issue E:

We use skeleton loader for inventory for some time. The inventory page is first loaded with spinner and once user paginate/filter there is skeleton table used.

> Issue F:

I used [grouped-checkbox-input](https://patternfly-react.surge.sh/patternfly-4/documentation/react/components/select#grouped-checkbox-input), but we can always apply custom styles if you think that it would be better.

> Issue G:

Yeah, that is to indicate there are more tags in the system, but they didn't fit the filter criteria so user should narrow the search.

------------------
Question about the filter placeholder, should it be `Filter systems by tag` or `Filter tags by namespace/key:value` to indicate how to filter tags?

KendraMar:
oh, we forgot a clear mechanism! can you include the x (white x with filled circle around it) to the left of the drop down caret?

karelhala:
@KendraMar that clear mechanism would be aweome, however PF does not support it. Do you want me to create issue on their side?

KendraMar:
@karelhala Interesting...they show a tiny X icon in the type ahead example here:

https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/filters#type-ahead-filters

karelhala:
@KendraMar oh yeah. It's even in styles doc. But missing in the React implementation, that sure as hell is bug. I'll open an issue on their side.

karelhala:
Issue has been created https://github.com/patternfly/patternfly-react/issues/3460

karelhala:
@KendraMar I've added a bit of a hack to add the clear functionality. It's not nice, but it works. Any issues with the rest of implementation? Do you want me to shoot a video to demonstrate current behavior?

KendraMar:
awesome!!

karelhala:
@KendraMar here is the video https://i.imgur.com/XSk2zQT.mp4

packages/inventory/src/EntityTableToolbar.js
```diff
@@ -66,30 +74,63 @@ class ContextEntityTableToolbar extends Component {
...
+                                isDisabled: true,
+                                className: 'ins-c-inventory__tags-more-items'
+                            }]
+                        }] : []
+                    ]
+                } : {
+                    items: [
+                        {
+                            label: !allTagsLoaded ? <Fragment>
+                                    Loading... <Spinner size="md" />
```
ryelo: Can we add this text in a span?

packages/inventory/src/InventoryList.scss
```diff
@@ -20,6 +20,29 @@
...
+
+        .ins-c-inventory__tags-no-tags {
+            margin: auto;
+        }
+
+        .pf-c-spinner {
+            margin-left: auto;
+        }
+    }
+}
```
ryelo: Do you mind doing something like this for one liners?
```
.ins-c-inventory__tags-filter ul {
    label.pf-c-check__label {
        display: flex;
        width: 100%;

        .pf-c-badge { margin-left: auto; }
    }

    .ins-c-inventory__tags-tail {
        display: flex;

        .ins-c-inventory__tags-no-tags { margin: auto; }
        .pf-c-spinner { margin-left: auto; }
    }
}
```

packages/inventory/src/api/index.js
```diff
@@ -55,8 +71,8 @@ export function getEntities(items, {
     orderBy = 'updated',
     orderDirection = 'DESC'
 }) {
-    const hostnameOrId = filters ? filters.find(filter => filter.value === 'hostname_or_id') : undefined;
-
+    console.log(TagsApi, 'sdf');
```
ryelo: Extra log message here

packages/inventory/src/api/index.js
```diff
@@ -72,14 +88,15 @@ export function getEntities(items, {
...
-            hostnameOrId && hostnameOrId.filter,
-            undefined,
+            hostnameOrId,
             undefined,
             undefined,
             perPage,
             page,
             orderBy,
             orderDirection,
+            '', //staleness
```
ryelo: Do you want to keep this comment?

packages/inventory/src/constants.js
```diff
@@ -41,8 +57,25 @@ export function constructGroups(allTags) {
...
         type: 'checkbox',
-        items: tags.map(({ tagName, tagValue }) => ({
-            label: `${tagName} - ${tagValue}`
+        items: tags.map(({ count, tag: { key: tagKey, value } }) => ({
+            label: <React.Fragment>
+                <div>{tagKey}: {value}</div>
+                <Tooltip
+                    position="right"
+                    enableFlip
+                    content={`Applicable to ${count} systems.`}
```
ryelo: Pluralize here?

packages/inventory/src/api/index.js
karelhala: Yup, those arguments are starting to get bigger so just to pin point what is what.

KendraMar:
@karelhala @ryelo The concern I have with this implementation is managing a larger (eg, larger than 9) set of tags. If a system has 20 tags, will you populate the dropdown with 20 choices before the user starts typing? the design intention was to behave the way spotlight search on the Mac OS works.

karelhala:
@KendraMar I see, sadly Patternfly does not support similirar approach. The closest one is groupped select which is used here. We can limit the number of tags user sees per each search. Currently it is limited to 25, but as you said that can be overwhelming so how about we show them 10 tags when they first open the filter (nothing typed in) and same when they filter - 10 tags that were found for the searched term.

If there is more than 10 tags we will show `X more tags available` and in future we can make this clickable. That will spawn table with all tags and checkboxes so user can select multiple tags, filter them and paginate as well. Once user clicks on `Submit tags` we will filter with them. That being said, the dialod is not partof this work.

KendraMar:
Ok, that's my misundertanding then. I thought we were implementing a new component -- type ahead search. I think Ian thought this as well. If we can't do this now, maybe we can do this in May?

@iphands

KendraMar:
If we're going with the let's show 10 tags approach, which 10 should we show? Guessing we'd go by alphabetical order.

karelhala:
@KendraMar I've limited the amount of tags to be 10, so users can see 10 tags at most. Just to clarify, if there is more than 10 tags (either filtered or not filtered) user will see `X more of tags` at the bottom. So if they type `dev` and this would result in 50 tags, they will see first 10 tags and at the bottom `40 more tags`. If they'd start typing `development` which would result in let's say 8 tags they will see only these 8 tags. Once they clear out, or change the filter the amount of tags visible and hinted (by x more tags) will change.

I think we are good to merge this and we can re-iterate on the mockups and how this is displayed - I especially like the `x more tags` to be clickable to show modal with table of all tags assigned to user and user can then filter, paginate, sort and select tags to be applied to filter.

KendraMar:
That sounds reasonable. Thanks, Karel!

packages/inventory/src/EntityTableToolbar.js
```diff
@@ -36,60 +42,107 @@ class ContextEntityTableToolbar extends Component {
...
     }
 
     applyTags = (newSelection, debounced = true) => {
-        const { allTags, page, perPage, filters, onRefreshData } = this.props;
-        const tagFilters = mapGroups(newSelection, allTags);
+        const { perPage, filters, onRefreshData } = this.props;
+        const tagFilters = mapGroups(newSelection);
         const tagFiltersIndex = filters.findIndex((value) => value.hasOwnProperty('tagFilters'));
-        filters.splice(tagFiltersIndex, 1);
+        if (tagFiltersIndex !== -1) {
```
PanSpagetka: I think it is not necessary to use `splice` here. Something like `filters.filter(value) => !value.hasOwnProperty('tagFilters'))` should be enough.

karelhala: Yeah, you are right!

PR #311, opened 2020-01-06
Update DateFormat tests in order to fix failing tests

karelhala:
In order to properly test Date creation we have to mock Date so we have control over it.

### DIFF
--- a/packages/components/src/Components/DateFormat/DateFormat.test.js
+++ b/packages/components/src/Components/DateFormat/DateFormat.test.js
@@ -4,6 +4,29 @@ import toJson from 'enzyme-to-json';
 import DateFormat from './DateFormat';
 
 describe('DateFormat component', () => {
+    const _Date = Date;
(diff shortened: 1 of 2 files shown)


packages/components/src/Components/DateFormat/DateFormat.test.js
```diff
@@ -4,6 +4,29 @@ import toJson from 'enzyme-to-json';
 import DateFormat from './DateFormat';
 
 describe('DateFormat component', () => {
+    const _Date = Date;
+    const currDate = new Date('1970');
+    beforeAll(() => {
+        /*eslint no-global-assign:off*/
+        Date = class extends Date {
```
Hyperkid123: I could imagine this mock to be useful for all components/packages. Maybe it could be added in Jest setup instead? This will override the class anyway and if you run jest in parallel it will hit other test files too.

PR #316, opened 2020-01-07
Do not select row on disabled tags

karelhala:


packages/inventory/src/InventoryList.scss
```diff
@@ -58,6 +58,10 @@
         }
     }
 
+    .ins-c-inventory__list-tags {
+        display: inline-block;
+    }
```
ryelo: Not a blocker, but we should try to keep one style CSS to one line

PR #320, opened 2020-01-09
Card select updates

rvsia:
Allow to use spacebar to select items
When there is a icon, text has set elipsis

### DIFF
--- a/packages/sources/src/sourceFormRenderer/components/CardSelect.js
+++ b/packages/sources/src/sourceFormRenderer/components/CardSelect.js
@@ -37,7 +37,8 @@ class CardSelect extends React.Component {
     }
 
     handleKeyPress = (event, value) => {
-        if (event.key === ' ') {
+        const spaceBar = 32;
+        if (event.charCode === spaceBar) {
             event.preventDefault();
             this.onClick(value);
         }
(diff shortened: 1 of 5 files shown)


packages/sources/src/sourceFormRenderer/components/CardSelect.js
```diff
@@ -37,7 +37,8 @@ class CardSelect extends React.Component {
     }
 
     handleKeyPress = (event, value) => {
-        if (event.key === 'Enter') {
+        if (event.key === ' ') {
```
Hyperkid123: Is it possible to change this to use key code? `'  '` is easily overlooked I think.

packages/sources/src/sourceFormRenderer/components/CardSelect.js
```diff
@@ -64,12 +65,12 @@ class CardSelect extends React.Component {
                     isCompact={ true }
                 >
                     <div className={ disabled ? 'disabled' : '' }>
-                        <CardHeader>
+                        <CardHeader className={Component ? 'text-elipsis' : '' }>
```
Hyperkid123: Just a tip. If you have more conditional classnames in this app, consider using library like https://www.npmjs.com/package/clsx

packages/sources/src/sourceFormRenderer/components/CardSelect.js
rvsia: Fixed

PR #326, opened 2020-01-17
Add getting started message for systems with insight_id = null

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-4063

so previously, if a system didn't have insights installed, n yah viewed it in inventory, we'd say no rule hits, which is technically incorrect, as the system hasn't been scanned, have an "install insights" call to action

#### looks like this! (ignore that it's on the rules tab)
Screen Shot 2020-01-17 at 1 53 54 PM

### DIFF
--- a/packages/inventory-insights/src/Insights.js
+++ b/packages/inventory-insights/src/Insights.js
@@ -1,14 +1,13 @@
 /* eslint-disable camelcase */
 import './insights.scss';
 
-import { AnsibeTowerIcon, CheckCircleIcon, CheckIcon, ExternalLinkAltIcon, PficonSatelliteIcon, TimesCircleIcon } from '@patternfly/react-icons';
+import { AnsibeTowerIcon, ChartSpikeIcon, CheckCircleIcon, CheckIcon, ExternalLinkAltIcon, PficonSatelliteIcon, TimesCircleIcon } from '@patternfly/react-icons';
 import { BASE_FETCH_URL, FILTER_CATEGORIES as FC } from './Constants';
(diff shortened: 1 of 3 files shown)


packages/inventory-insights/src/Insights.js
```diff
@@ -403,13 +403,33 @@ class InventoryRuleList extends Component {
                                     text={`This filter criteria matches no rules. Try changing your filter settings.`} />
                             }
                         </Fragment>
-                        :
-                        <Card>
+                        : entity.insights_id !== null ? <Card>
                             <CardBody>
-                                <MessageState icon={CheckIcon} iconStyle={{ color: global_success_color_200.value }} size='md'
+                                <MessageState icon={CheckIcon} iconClass='checkIconColor' size='md'
```
karelhala: Can you make these classes more specific?

`ins-c-insights__check`

packages/inventory-insights/src/Insights.js
```diff
@@ -369,7 +369,7 @@ class InventoryRuleList extends Component {
                             isDisabled={selectedAnsibleRules.length === 0}
                             dataProvider={() => this.processRemediation(selectedAnsibleRules)}
                             onRemediationCreated={result => addNotification(result.getNotification())} >
-                            {AnsibeTowerIcon && <AnsibeTowerIcon size='sm' color={global_BackgroundColor_100.value} />} Remediate
+                            {AnsibeTowerIcon && <AnsibeTowerIcon size='sm' className='backgroundColor100' />} Remediate
```
karelhala: How about `ins-c-background__default`?

PR #336, opened 2020-01-29
Fire only refresh when user changes pagination and refresh is set

karelhala:
When consumer passes `onRefresh` to inventory use it directly instead of `onRefreshData`. We rely on the consumer to refresh the data, if the data remains the same as previously they have to fire `onRefreshData`.

Also when using `onRefresh` fire loading so the inv table shows loading state.


karelhala:
ping @katerinapat you might be interested in this PR.

katerinapat:
@karelhala I didn't dive into the code but I tested for vuln-ui and it works as expected. thanks for all the effort :D

PR #340, opened 2020-02-03
Update build config to minimize bundle sizes.

Hyperkid123:
Build sizes should be about 60% smaller. It would be best if packages would slowly migrate to absolute import paths so we can remove the babel transform import plugin.

The build optimization comes mostly from PF. We were discussing some import changes with PF team for a long time and managed to convince them to let us implement some changes. This should result in major bundle size improvements if you use the latest versions of  PF packages.


packages/inventory-general-info/.gitignore
```diff
@@ -1,3 +1,4 @@
 /index.js
 /index.css
+/TextInputModal.css
```
karelhala: Can you add these lines to `.npmignore` as well please?
```
!/TextInputModal.css
!/TextInputModal.js
```

packages/remediations/src/RemediationButton.js
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
```
karelhala: Can you please add these lines to `.npmignore` as well
```
!/RemediationButton.css
```

packages/sources/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@redhat-cloud-services/frontend-components-sources",
-    "version": "0.0.34",
+    "version": "0.0.34-RC3",
```
karelhala: No need to include the `RC` version.

packages/utils/.gitignore
```diff
@@ -1,2 +1,3 @@
 files/
 Styles.css
+helpers.css
```
karelhala: Can you please add this to `.npmignore` as well, so we actually ship it.
```
!helpers.css
```

karelhala:
release major

karelhala:
releasemajor

karelhala:
release major

AllenBW:
nacho-bot, throwing a big celebration for the major release there...

PR #341, opened 2020-02-03
added helper function to get relative time

mkholjuraev:
Helper function to get relative time which is suggested  by PatternFly. https://www.patternfly.org/v4/design-guidelines/content/grammar-and-terminology

### DIFF
--- a/packages/utils/config/webpack.config.js
+++ b/packages/utils/config/webpack.config.js
@@ -12,6 +12,7 @@ module.exports = (env) => ({
         ReducerRegistry: './src/ReducerRegistry.js',
         helpers: './src/helpers.js',
         MiddlewareListener: './src/MiddlewareListener.js',
+        RelativeTimeHelper: './src/RelativeTimeHelper.js',
         Registry: './src/Registry.js',
         Deffered: './src/Deffered.js',
         RouterParams: './src/RouterParams.js',
--- a/packages/utils/src/RelativeTimeHelper.js
+++ b/packages/utils/src/RelativeTimeHelper.js
@@ -1,0 +1,65 @@
+class RelativeTime {
+
+    constructor(dateFrom, dateTo) {
+        this.dateFrom = dateFrom;
+        this.dateTo = dateTo;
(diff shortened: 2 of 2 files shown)


karelhala:
We already have a component for this. Look at [`DateFormat`](https://github.com/RedHatInsights/frontend-components/blob/master/packages/components/src/Components/DateFormat/DateFormat.js). It's a component that wraps this whole thing. By default it uses `relative` format so you get the date calculated from curr date. You don't have to pass the initial date in there, however if that's what you need I would prefer updating that component with new prop, something like `fromDate`. Also if you need just to calculate the date and perhaps send to server we could expose functions that does that for you.

mkholjuraev:
@karelhala I would appreciate if you can tell me how to import DateFormat into the project.

karelhala:
@mkholjuraev `npm i -S @redhat-cloud-services/frontend-components` to install frontend components. Once you install them you can use the `DateFormat` component wherever you need by importing it 
```JSX
import React from 'react';
import { DateFormat } from '@redhat-cloud-services/frontend-components'
const MyCmp = () => {
  <div>
    <DateFormat date={new Date()} />
  </div>
}
```

PanSpagetka:
You can import and use it like this:
```
import { DateFormat } from '@redhat-cloud-services/frontend-components';
...
 <DateFormat type='relative'  date='Aug 31 2019 UTC' />
```

You can provide Date object, string or number as a date.
More details here: https://cloud.redhat.com/docs/storybook?path=/story/components--dateformat

mkholjuraev:
I tried these already, but I am getting following error:

Uncaught (in promise) Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

karelhala:
What version of Frontend-components do you have? You might need to update to more recent one.

mkholjuraev:
I was using 0.0.29. I updated it. Everything is working now thank you!

PR #344, opened 2020-02-05
Export ComplianceRemediationButton and make minor changes

bastilian:
We'll use this ComplianceRemediationButton instead of writting yet another one in compliance-frontend. This exports the component and adds changes to make this possible.

### DIFF
--- a/packages/inventory-compliance/src/ComplianceRemediationButton.js
+++ b/packages/inventory-compliance/src/ComplianceRemediationButton.js
@@ -8,10 +8,6 @@ import { AnsibeTowerIcon } from '@patternfly/react-icons';
 import { global_BackgroundColor_100 as globalBackgroundColor100 } from '@patternfly/react-tokens';
 
 class ComplianceRemediationButton extends React.Component {
-    constructor(props) {
-        super(props);
-    }
-
     formatRule = ({ title, refId }, profile, system) => ({
         id: `ssg:rhel7|${profile}|${refId}`,
         description: title,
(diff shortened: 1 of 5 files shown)


packages/inventory-compliance/src/ComplianceRemediationButton.js
```diff
@@ -44,10 +40,10 @@ class ComplianceRemediationButton extends React.Component {
         const result = { systems: [], issues: [] };
         allSystems.forEach(async (system) => {
             result.systems.push(system.id);
-            if (selectedRules) {
+            if (selectedRules.length !== 0) {
                 result.issues.push(this.rulesWithRemediations(selectedRules, system.id));
             } else {
-                result.issues.push(this.rulesWithRemediations(system.rule_objects_failed, system.id));
+                result.issues.push(this.rulesWithRemediations(system.ruleObjectsFailed, system.id));
```
dLobatog: If you change this, you also have to change in SystemRulesTable: `ruleObjectsFailed: [] }] }`

packages/inventory-compliance/src/ComplianceRemediationButton.js
```diff
@@ -57,15 +53,18 @@ class ComplianceRemediationButton extends React.Component {
         });
     }
 
+    noRemediationAvailable = () => {
+        const { allSystems } = this.props;
+        return !allSystems.some((system) => system.ruleObjectsFailed.some((rule) => rule.remediationAvailable));
```
dLobatog: You need to take into account the second condition in the original conditional:

```
 selectedRules.length === 0
```

This happens on the system details page - in that page, `ruleObjectsFailed` is just `[]`, and we just select rules individually. Only `selectedRules` changes as one or more rules are selected, so right now since that's not taken into account, the button is always disabled on that page. It works great on all other pages though!

PR #348, opened 2020-02-06
Pagination dropdown at bottom of modal opens up

johnsonm325:


karelhala:
This has already been fixed in underlying component https://github.com/RedHatInsights/frontend-components/blob/master/packages/inventory/src/Pagination.js#L66 are you still experiencing this issue?

johnsonm325:
@karelhala Yep, you're right. Must have been fixed recently. Thanks!

PR #349, opened 2020-02-06
Add sortBy to skeleton and paddingColumnSize

josejulio:
- The purpose is to show the ordering (the arrow on the column)
   when loading
 - The padding had to be added to match the sort.index of the
   original table, as the collapsible and select were affecting it

Without:
without-sortby

With:
with-sortby


josejulio:
Thanks for the review @karelhala 
Addressed your comments and rebased

PR #353, opened 2020-02-11
Exports inv-insights report details card for use by other apps

AllenBW:
We're pullin CVEs out, sure would be nice if they could also use this handy dandy report rendering card 🤤🎊

### DIFF
--- a/packages/inventory-insights/package.json
+++ b/packages/inventory-insights/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@redhat-cloud-services/frontend-components-inventory-insights",
-    "version": "0.4.2",
+    "version": "1.0.1",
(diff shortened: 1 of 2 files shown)


packages/inventory-insights/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@redhat-cloud-services/frontend-components-inventory-insights",
-    "version": "0.4.2",
+    "version": "1.0.1",
```
karelhala: No need to update the version, bots should take care of it.

AllenBW: i knooooooowwwww but hate to see this field outta sync with publish field.... 
maybe we'll just can it next time 🤔

PR #357, opened 2020-02-12
Truncate component to use sanitizehtml dist

AllenBW:
Not thinkin this will help everyone... but just in the case where yah already have `sanatizehtml` around... it should reduce the size of this component

#### in this case for instance
 
Screen Shot 2020-02-12 at 11 58 35 AM


packages/components/src/Components/Truncate/Truncate.js
```diff
@@ -1,13 +1,12 @@
...
-import propTypes from 'prop-types';
-
-import classNames from 'classnames';
-
-import { Button, Stack, StackItem } from '@patternfly/react-core';
-
 import './truncate.scss';
 
-import sanitizeHtml from 'sanitize-html';
+import { Stack, StackItem } from '@patternfly/react-core/dist/js/layouts/Stack/index';
```
Hyperkid123: `/js/layouts/Stack/Stack'` index has all components from `/Stack` folder (dunno if there is more)

packages/components/src/Components/Truncate/Truncate.js
```diff
@@ -1,13 +1,12 @@
...
+
+import { Button } from '@patternfly/react-core/dist/js/components/Button/Button';
+import React from 'react';
+import classNames from 'classnames';
+import propTypes from 'prop-types';
+import sanitizeHtml from 'sanitize-html/dist/sanitize-html';
```
Hyperkid123: I am not sure if this is going to help. The module main attribute (in package.json) points to `sanitize-html/dist/sanitize-html` anyway.

This is thing is simply HUGE.

packages/components/src/Components/Truncate/Truncate.js
AllenBW: ? confused, yeah we're using the index file here?

packages/components/src/Components/Truncate/Truncate.js
AllenBW: maybe we should explicitly point to the minned one 🤔 half the size
Screen Shot 2020-02-13 at 8 52 05 AM

Hyperkid123: It should get minimized by webpack while building the bundles

packages/components/src/Components/Truncate/Truncate.js
Hyperkid123: You want to use the actual file with the component. The index is only exporting all files from the same folder so you might end up with more components in your build. Like with `Button/Button` you want to point towards `Stack/Stack`

AllenBW:
Hokay, closing for now 🤷‍♂ 🤸‍♀

PR #362, opened 2020-02-17
fix(cost-management): text and fields update

boaz0:
| # |Change | Description |
| -- | ---------  | --------------- |
| 1 | settings > sources > cost/aws flow > step 4 (cost allocations) | Replace step 1 text with the following: "1. In the AWS Billing and Cost Management console, open the Cost Allocation Tags section." |
| 2 | settings > sources > cost/aws flow | when the wizard gets to step 3, it auto-scrolls to the bottom of the step, so the top text could get cutt off |
| 3 | Settings > sources > cost/aws | the left most bullets should be numbered (e.g. 1. Specify or create... ; 2. Create a cost and usage report... ; 3. Enter the name...)|
| 4 | settings > sources > cost/azure | for step 4, after the user has entered something into the first required field, the Next button becomes enabled.  this allows the user to skip to the next step, even though there is still 3 required fields that had not been filled out.  Basically tenant id, client id, and client secret is not being treated as a required field, when it should|
| 5 | settings > sources > cost/openshift | step 5, cluster id should be a required field (show the indication)|

### 1 - Replace step 1 text
fix-aws-text-update

### 3 - The left most bullets should be numbered
fix-aws-bullets-to-number

### 4 - Next button is disabled until all fields are given
fix-azure-validation

### 5 - show the required indication
fix-ocp-step-5

// cc @rvsia @nlcwong


packages/sources/src/addSourceWizard/hardcodedComponents/aws/arn.js
```diff
@@ -11,25 +11,33 @@ import {
     ClipboardCopyVariant
 } from '@patternfly/react-core';
 
-export const UsageDescription = () => (<TextContent>
-    <Text component={ TextVariants.p }>
+export const UsageDescription = () => {
+    React.useEffect(() => {
```
rvsia: No hooks in frontend-components yet. :crying_cat_face: 

Also, I think it would be better to fix it in the Data Driven Forms package, as it is a global issue with the wizard.

packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costManagement.js
```diff
@@ -194,16 +194,17 @@ export const ConfigureUsageCollector = () => (
...
+    // Github issue: https://github.com/patternfly/patternfly-react/issues/3731
+    <React.Fragment>
+        Cluster identifier <span className="pf-c-form__label-required" aria-hidden="true">*</span>
         <Popover
             aria-label="more-information-popover"
             position="top"
             bodyContent={<Text>The cluster identifier value is stored in <b>~/.config/ocp_usage/config.json</b></Text>}
         >
-            <Button variant={ButtonVariant.plain}><QuestionCircleIcon /></Button>
+            <Button variant={ButtonVariant.plain} style={{ paddingRight: '3px', paddingLeft: '5px' }}><QuestionCircleIcon /></Button>
```
rvsia: no inline styles please

packages/sources/src/addSourceWizard/hardcodedSchemas.js
```diff
@@ -213,7 +213,10 @@ export default {
                         }, {
                             name: 'credentials.subscription_id',
                             component: componentTypes.TEXT_FIELD,
-                            label: 'Subscription ID'
+                            label: 'Subscription ID',
+                            validate: [{
```
rvsia: This is already declared here https://github.com/RedHatInsights/frontend-components/blob/854bb9a00b2d536bb3c69c712ec4f79120785d91/packages/sources/src/addSourceWizard/hardcodedSchemas.js#L153

nlcwong:
LGTM

packages/sources/src/addSourceWizard/hardcodedComponents/aws/arn.js
boaz0: Dang :sob: I really wanted to use hooks. But that's OK :smiley_cat: maybe next time.
I am not sure where I should do that in DDF. Should we put it as a `"scrollTop": true` attribute in the schema?

packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costManagement.js
boaz0: Yep, you're right. I'll fix it.

packages/sources/src/addSourceWizard/hardcodedSchemas.js
boaz0: Thanks! I missed that.

packages/sources/src/addSourceWizard/hardcodedComponents/aws/arn.js
rvsia: I think you could open a PR and fix it somewhere around here https://github.com/data-driven-forms/react-forms/blob/620a52ebf0d3513414e80dba34505a02b1185456/packages/pf4-component-mapper/src/form-fields/wizard/wizard-step.js#L14 

or open an issue and I will do it! 

This is really a nice catch!

boaz0:
Thanks @rvsia ! :bowing_man:

rvsia:
@karelhala For some reason, nacho-bot release two versions :(

karelhala:
@rvsia github is having some difficulties so it triggered the build multiple times

PR #363, opened 2020-02-18
Stop extracting css to files and inject them with component mount.

Hyperkid123:
Styles are now injected into the DOM only when the component is actually used. For instance if somebody is using only `Section`  component before they would get every CSS rule from every component. Now they will only get the `Section` specific CSS rules.

There is a breaking change. No stylesheets will be generated so if somebody is importing any CSS files from these libraries have to remove these imports from the codebase.

Styles are still working. Tested in the demo and also in Catalog APP.
Screenshot from 2020-02-18 09-54-57

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16343,54 +16343,6 @@
       "integrity": "sha512-jf84uxzwiuiIVKiOLpfYk7N46TSy8ubTonmneY9vrpHNAnp0QBt2BxWV9dO3/j+BoVAb+a5G6YDPW3M5HOdMWQ==",
       "dev": true
     },
-    "mini-css-extract-plugin": {
-      "version": "0.4.3",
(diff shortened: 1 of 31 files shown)


ryelo:
> There is a breaking change. No stylesheets will be generated so if somebody is importing any CSS files from these libraries have to remove these imports from the codebase.

I don't know if we want to do this for the utils package. I know there are apps that are using sass mixins from the utils. We can probably get teams off of it, but we should definitely mark this release as major and bump the package to (x.0.0).

Hyperkid123:
@ryelo Karel said the same thing and i agree, I will preserve the CSS files sow e don't break apps that are using css imports in all packages.

Hyperkid123:
@ryelo we had some more discussion with @karelhala  and we will probably bundle this change to the same PR when we will migrate to rollup instead of webpack. We wanted to do that for some time now and this issue should have a simpler solution with rollup than with webpack.

Closing this PR in favor of that.

PR #367, opened 2020-02-18
Allow hiding active filter label for ConditionalFilter component

bastilian:
This adds a `hideLabel` prop for the ConditionalFilter to allow hiding the label of the current active filter.

Screenshot 2020-02-18 at 18 54 27

While still showing the labels for the dropdown items available: 

Screenshot 2020-02-18 at 18 56 18

### DIFF
--- a/packages/components/src/Components/ConditionalFilter/ConditionalFilter.js
+++ b/packages/components/src/Components/ConditionalFilter/ConditionalFilter.js
@@ -56,7 +56,7 @@ class ConditionalFilter extends Component {
                                         toggle={
                                             <DropdownToggle
                                                 onToggle={ this.dropdownToggle }
-                                                className={ hideLabel ? 'no__label' : '' } >
+                                                className={ hideLabel ? 'ins-c-conditional-filter__no-label' : '' } >
                                                 <FilterIcon size="sm" />
                                                 { !hideLabel &&
                                                     <span className="ins-c-conditional-filter__value-selector">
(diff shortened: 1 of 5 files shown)


packages/components/src/Components/ConditionalFilter/ConditionalFilter.js
```diff
@@ -54,11 +54,15 @@ class ConditionalFilter extends Component {
                                         onSelect={ () => this.dropdownToggle(false) }
                                         isOpen={ isOpen }
                                         toggle={
-                                            <DropdownToggle onToggle={ this.dropdownToggle }>
+                                            <DropdownToggle
+                                                onToggle={ this.dropdownToggle }
+                                                className={ hideLabel ? 'no__label' : '' } >
```
karelhala: The className is a bit unfortunate. Can you name it perhaps `ins-c-conditional-filter__no-label`?

packages/components/src/Components/ConditionalFilter/conditional-filter.scss
```diff
@@ -5,12 +5,20 @@
...
 
-        & > svg {
-            margin-right: var(--pf-global--spacer--xs);
+            & > svg {
+                margin-right: var(--pf-global--spacer--xs);
+            }
+        }
+
+        &.no__label {
+            .pf-c-dropdown__toggle-text {
```
karelhala: Since this is effectively a single selector can you make align it on one line? `&.no__label .pf-c-dropdown__toggle-text {`

bastilian:
@karelhala Thanks for the review! The class is now renamed and the CSS better formated.

PR #370, opened 2020-02-19
Align canvas items so they can be viewed correctly in previews

karelhala:
### After
[e11469f9-b74a-4dcc-ad85-179e6e46b6c3.pdf](https://github.com/RedHatInsights/frontend-components/files/4226054/e11469f9-b74a-4dcc-ad85-179e6e46b6c3.pdf)


katerinapat:
I just checked it on Mac and it looks great!

PR #382, opened 2020-02-28
Re-enable collapsible inventory and document it

karelhala:


packages/inventory/src/EntityTable.js
```diff
@@ -150,11 +153,14 @@ class EntityTable extends React.Component {
...
     createColumns = () => {
-        const { columns, hasItems, rows } = this.props;
+        const { columns, hasItems, rows, expandable: isExpendable } = this.props;
         return columns.map(({ props, transforms, ...oneCell }) => ({
             ...oneCell,
             transforms: [
                 ...this.buildTransforms(props, transforms, hasItems, rows) || []
+            ],
+            cellFormatters: [
+                ...isExpendable ? [ expandable ] : []
```
josejulio: typo i suppose, it should be `isExpandable`

karelhala: That's actually `expandable` from `@patternfly/react-table` package.

packages/inventory/src/EntityTable.js
```diff
@@ -50,65 +52,55 @@ class EntityTable extends React.Component {
...
                 }
+            );
+        }
 
-                const [ , date, month, year, time ] = new Date(get(col, key, ' ')).toUTCString().split(' ');
-                if (date && month && year && time) {
-                    return `${date} ${month} ${year}, ${time.split(':').slice(0, 2).join(':')} UTC`;
-                }
+        if (isTime) {
+            if (DateFormat) {
```
Hyperkid123: I would probably mark the correct version of frontend components in dependencies or get rid of this. You have basically the same code here twice.

packages/inventory/src/EntityTable.js
```diff
@@ -134,11 +126,22 @@ class EntityTable extends React.Component {
             }];
         }
 
-        return rows
-        .map((oneItem) => ({
-            ...oneItem,
-            cells: this.buildCells(oneItem)
-        }));
+        return flatten(rows
```
Hyperkid123: You could use flat map here. Babel should be able to handle it I think.

karelhala: I would have to import it, because it was removed from Babel 7. And we were told, that some browsers were not fully supporting `flatMap`

Hyperkid123: I meant just `flat` that should be supported by everything now except IE.

karelhala: Oh yeah, that can work. I checked the support matrix and looks like the biggest problem was with FF, they started to support `flat` at version 62, that was released on 2018-05-07, almost 2 years. So we should be fine.

packages/inventory/src/EntityTable.js
karelhala: This component is hot loaded, meaning that the dep version in package is not used and apps are sending their version. I know it's stupid to check for it, but we don't really have other option in here.

Hyperkid123: Ok

packages/inventory/src/EntityTable.js
josejulio: it says isExp**E**ndable instead of isExp**A**ndable
Just a typo

karelhala: Oh, I thought you were refering to `expandable`, sure thing! I'll change it in the whole file.

PR #383, opened 2020-03-02
Introduce new package that will serve rule components

karelhala:


packages/rule-components/src/index.js
```diff
@@ -0,0 +1 @@
+console.log('rule-components');
```
PanSpagetka: console.log

packages/rule-components/rollup.config.js
```diff
@@ -0,0 +1,92 @@
...
+    if (last === 'src') {
+        return {
+            index: item
+        };
+    }
+
+    return {
+        [last]: item
+    };
+}).reduce((acc, curr) => ({ ...acc, ...curr }), {});
```
PanSpagetka: I think you can do it in one reduce, something like this should work:
```
const globMapper = (mapper) => glob.sync(mapper).reduce((acc, item) => {
    const [ path ] = item.split('/index.js');
    const last = path.split('/').pop();
    return {
      ...acc,
      ...last === 'src' ? { index: item } : { [last]: item }
    };
}, {});
```

packages/rule-components/rollup.config.js
```diff
@@ -0,0 +1,92 @@
...
+            name,
+            globals
+        },
+        external,
+        plugins
+    })),
+    ...Object.entries(globMapper('src/**/index.js')).map(([ key, input ]) => ({
+        input,
+        output: {
+            file: `./dist/umd/${key}.js`,
```
Hyperkid123: I am not sure if it's a good idea to code a split UMD package. You can't read from any UMD/IIFE package so you will not avoid bundling all the necessary resources into a UMD package. You risk having the same code multiple times in multiple outputs.

karelhala: That's actually what is needed from umd output - multiple files with similiar content. If you end up using only 1/3 of the code you don't really need the rest of it.

packages/rule-components/src/index.js
karelhala: That's actually needed, because we don't have the component ready, yet.

packages/rule-components/rollup.config.js
Hyperkid123: I was thinking more about you can end up with 10x the same code if you use 10 different files that all use the same common function for instance. The end result might be much worse than just having it all in one file. UMD is meant to be self-contained.

Unless the files/components themselves are supposed to be self-contained and wont reference each other in the builds.

PR #384, opened 2020-03-02
Rule component

karelhala:


packages/rule-components/src/RuleTable/RuleTable.js
```diff
@@ -0,0 +1,227 @@
...
+        }
+    }
+
+    onCollapse = (_e, key, isOpen, { ruleId }) => {
+        const { expanded } = this.state;
+        if (isOpen) {
+            this.setState({
+                expanded: [
+                    ...expanded,
+                    ruleId || key
```
bond95: I think better to avoid this construction and define variable that will have needed value, for example:
`const expandedId = ruleId || key`

packages/rule-components/src/RuleTable/RuleTable.js
```diff
@@ -0,0 +1,227 @@
...
+                        key
+                    )).filter(Boolean),
+                    onDelete: this.onFilterDelete
+                }}
+                pagination={pagination}
+                { ...toolbarProps }
+            />
+            <Table
+                {...props}
+                {...!isLoading && { actions } }
```
bond95: I suggest to avoid this contstructions as well, because it's really hard to read, it can be easily transformed into:
`actions={!isLoading && actions}`

packages/rule-components/src/RuleTable/RuleTable.js
```diff
@@ -0,0 +1,227 @@
...
+    isLoading: PropTypes.bool,
+    columns: PropTypes.arrayOf(PropTypes.shape({
+        title: PropTypes.node,
+        selector: PropTypes.oneOfType([ PropTypes.string, PropTypes.arrayOf(PropTypes.string), PropTypes.func ]),
+        disableSort: PropTypes.bool
+    })),
+    sort: PropTypes.string,
+    defaultFilters: PropTypes.arrayOf(PropTypes.shape({})),
+    fetchData: PropTypes.func,
+    detail: PropTypes.oneOfType([ PropTypes.func, PropTypes.node ]),
```
bond95: According to what I see in README it's should be a function, but it's not. So maybe here it should be `PropTypes.func`?

packages/rule-components/README.md
```diff
@@ -6,24 +6,119 @@ If you want to use rules table component please pass entire dataset from server
 import React, { Component } from 'react';
 import ReactDOM from 'react-dom';
 import './index.scss';
-import RuleTable, { severityLabels } from '../../rule-components/src/RuleTable';
+import RuleTable, { severityLabels } from '@redhat-cloud-services/rule-components/dist/cjs/RuleTable';
```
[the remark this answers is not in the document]
Hyperkid123: @bond95 having absolute exports is useful for bundle size optimizations. If you import some function, component, constant, etc, from index.js in the non-ESM environment, tree shaking won't be applied and your bundle will end up will every single exported module instead of the one you are using.

packages/rule-components/README.md
```diff
@@ -6,24 +6,119 @@ If you want to use rules table component please pass entire dataset from server
...
+    descriptionFilter,
+    totalRiskFilter,
+    ansibleSupportFilter,
+    categoryFilter,
+    incidentRulesFilter,
+    likelihoodFilter,
+    riskOfChangeFilter,
+    ruleStatusFilter,
+    impactFilter
+} from '@redhat-cloud-services/rule-components/dist/cjs/RuleFilters';
```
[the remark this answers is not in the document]
Hyperkid123: ^^

packages/rule-components/src/RuleFilters/riskOfChange.js
```diff
@@ -0,0 +1,23 @@
...
+
+export default ({ onChange, value, ...props } = { onChange: () => undefined }) => ({
+    ...props,
+    label: 'Risk of change',
+    value: 'risk_of_change',
+    type: 'checkbox',
+    filterValues: {
+        value,
+        onChange,
+        items: [ ...new Array(riskOfChangeValues.length) ].map((_item, key) => ({
```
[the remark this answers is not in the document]
Hyperkid123: `[...new Array(5)]` will create an array with 5 items of `undefined` items. Then you can map over it and assing whatever you need.

packages/rule-components/src/RuleTable/RuleTable.js
Hyperkid123: defining extra variable will result in more memory usage (i know its minimal size). If you don't use a value like this multiple times, it's more efficient to do it inline so it will end up on the stack.

packages/rule-components/README.md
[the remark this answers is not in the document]
bond95: I understand that, but point of this example is not to reduce bundle size but show how to use this component, newbies will not just understand this import. So my suggestion is use typical import, also some remark about reducing size can be added in README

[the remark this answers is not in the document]
Hyperkid123: Actually "newbies" should peferably use absolute paths because they probably don't know how to optimize the bundle sizes.

packages/rule-components/src/RuleFilters/riskOfChange.js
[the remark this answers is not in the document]
bond95: Oh, ok, thank you

packages/rule-components/src/RuleTable/RuleTable.js
bond95: But it's value is used more then one time.

packages/rule-components/README.md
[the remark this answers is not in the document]
bond95: I'm agree, but the problem that newbies are don't know what is this and how and why to use it. So, I said, better add to README some additional suggestion like:
```Markdown
If you don't need to use whole package, please, import separate component like this:
 - RuleTable - @redhat-cloud-services/rule-components/dist/cjs/RuleTable
 - RuleFilters - @redhat-cloud-services/rule-components/dist/cjs/RuleFilters
```

packages/rule-components/src/RuleFilters/riskOfChange.js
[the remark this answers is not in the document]
bond95: But why not do like that?
```JavaScript
items: riskOfChangeValues.map((_item, key) => ({
```
Maybe I missed something, but I think this variant will be more faster. WDYT?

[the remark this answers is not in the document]
Hyperkid123: Yup

packages/rule-components/src/RuleTable/RuleTable.js
Hyperkid123: Yup

packages/rule-components/src/RuleTable/RuleTable.js
karelhala: Patternfly table has problems with some attributes even if they have value set to `false` I usually try to play safe and do this instead of assign with check. But looks like the actions can be false and plays nicely, I'll change it to suggested approach.

packages/rule-components/README.md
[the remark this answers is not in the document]
karelhala: As @Hyperkid123 said, this is to teach programmers to use direct imports as they are good with tree shaking. I'm going to add note that you can use package import, but be cautious and check if your bundle does bloat or not.

packages/rule-components/src/RuleFilters/riskOfChange.js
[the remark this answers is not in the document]
karelhala: Yeah, this is a leftover from previous implementation where I used numbers instead of constants. Good point.

packages/rule-components/src/RuleTable/RuleTable.js
karelhala: It can be either function or react nore, I'll change it in README.

packages/rule-components/README.md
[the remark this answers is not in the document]
bond95: I know, but my point was that in case of using direct import somewhere at least should be described why to use import in this way and what’s alternatives.

packages/rule-components/src/RuleTable/RuleTable.js
bond95: Unfortunately it can’t, because it’s already used as function. But react node is text for example, or number, or anything that react can render. So I suggest to leave it only as function.

packages/rule-components/src/RuleFilters/ansibleSupport.js
```diff
@@ -0,0 +1,17 @@
...
+    label: 'Ansible support',
+    value: 'ansible_support',
+    type: 'checkbox',
+    filterValues: {
+        value,
+        onChange,
+        items: ansibleSupportValues.map((_item, key) => ({
+            label: ansibleSupportLabels[key],
+            textual: ansibleSupportLabels[key],
+            value: ansibleSupportValues[key]
```
bond95: Maybe better to use here `_item` instead of `ansibleSupportValues[key]` and in couple of other files too?

packages/rule-components/src/RuleTable/RuleTable.js
```diff
@@ -0,0 +1,228 @@
...
+        }, () => this.onUpdate(true));
+    }
+
+    onSort = (_e, index, direction) => {
+        const { fetchData, rules, columns, detail } = this.props;
+        const { filterValues } = this.state;
+        fetchData({
+            meta: calculateMeta(rules.meta),
+            filterValues,
+            sortBy: { index, direction, column: columns[index - Boolean(detail)] }
```
bond95: Why to use here `index - Boolean(detail)`? Maybe better `index - Number(!!detail)`?

bond95:
Could you, please, add mechanism to manipulate with filters? For example we need to have risk filter outside of table, we have button by click on which table should be filtered by `Total risk` value

bond95:
@karelhala I got a few errors, first is:
image
Second is appear when I trying to import package with `import { RuleTable, severityLabels } from '@redhat-cloud-services/rule-components';
`
image

karelhala:
@bond95 looks like the build is not well optimized for linked package (the first error), you can import the component as `import { RuleTable } from '@redhat-cloud-services/rule-components/dist/umd';` in order to load the component properly.

bond95:
> @bond95 looks like the build is not well optimized for linked package (the first error), you can import the component as `import { RuleTable } from '@redhat-cloud-services/rule-components/dist/umd';` in order to load the component properly.

So I think you should update readme because I copied it from there 😃

karelhala:
@bond95 not really, the env (`esm`, `cjs` or `umd`) should be picked by your build tool. If you are experiencing some problems you can use `UMD` as fallback.

packages/rule-components/src/RuleTable/RuleTable.js
```diff
@@ -0,0 +1,249 @@
...
+        expanded: []
+    }
+
+    debounceFetchData = debounce(options => {
+        const { fetchData } = this.props;
+        fetchData(options);
+    }, 500)
+
+    componentDidUpdate(prevProps) {
+        if (JSON.stringify(this.props.filterValues) !== JSON.stringify(prevProps.filterValues)) {
```
bond95: I understand that this solution will work, but this is not right way to comparing two objects, because it very sensitive on attribute order, so, please, I strongly suggest to rewrite it properly to some deepEqual function. Or use simple comparing: `this.props.filterValues !== prevProps.filterValues` because if nothing changed object pointer should be the same.

packages/rule-components/src/RuleTable/RuleTable.js
```diff
@@ -0,0 +1,249 @@
...
+            this.setState({
+                ...this.state,
```
bond95: I'm not sure I think this line is useless, because setState automatically merge your new object and state object.

packages/rule-components/src/RuleTable/RuleTable.js
```diff
@@ -0,0 +1,229 @@
...
+            meta: calculateMeta(rules.meta),
+            filterValues,
+            sortBy: { ...sortBy, column: columns[sortBy.index - Boolean(detail)] }
+        });
+    }
+
+    onFilterDelete = (_e, [ deleteItem ], isAll) => {
+        const { filterValues } = this.state;
+        if (isAll) {
+            this.setState({ filterValues: {} }, () => this.onUpdate());
```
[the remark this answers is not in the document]
karelhala: I tried that, but for some reason that wasn't working.

packages/rule-components/src/RuleTable/RuleTable.js
karelhala: I'll probably use `isEqual` from lodash, good point.

packages/rule-components/src/RuleTable/RuleTable.js
karelhala: Yeah, I had a bit more complicated state setter here so I kept the spread in here, but it is better to update just the filterValues.

packages/rule-components/src/RuleTable/RuleTable.js
Sergey1011010: What do you think about using `const offset = detail ? -1 : 0;`?

packages/rule-components/src/RuleTable/RuleTable.js
```diff
@@ -0,0 +1,249 @@
...
+                cells={columns.map(({ transforms, ...column }) => ({
+                    ...column,
+                    transforms: [
+                        ...transforms || [],
+                        ...!isLoading && sortBy && !column.disableSort ? [ sortable ] : []
+                    ]
+                }))}
+                rows={
+                    isLoading ?
+                        [ ...new Array(5) ].map(() => ({
```
Sergey1011010: shouldn't this constant be moved to top of the file or something?

karelhala: Not really, if the data are not loading this is not invoked at all. If it was moved to the top of component it would mean that whenever component is used it will be invoked.

Sergey1011010: What do you mean? I'm saying that having hardcoded constants is a bad practice.

bond95: I suggest to use here method `fill`: `Array(5).fill()`. It will be more readable and you don't need to copy array.

packages/rule-components/src/RuleTable/RuleTable.js
```diff
@@ -0,0 +1,249 @@
...
+    rules: {
+        data: []
+    },
+    columns: [],
+    ariaLabel: 'Rules table',
+    isLoading: false,
+    sortBy: {},
+    filters: {},
+    filterValues: {},
+    fetchData: () => undefined
```
bond95: What do you think about `noop`, because I see same pattern in various places in code?

karelhala: I prefer this over importing noop because well I'd have to import a constant and use it. I have a macro that writes this automatically. And it is imediatelly visible, you know what the function is doing whereas with noop you'd have to trust that noop is doing the same thing.

packages/rule-components/src/RuleTable/RuleTable.js
karelhala: For me it's easier to read spread than fill, sorry.

I've added new prop so the number of loading bars/rows can be changed.

Sergey1011010: @karelhala, great, thanks

PR #392, opened 2020-03-06
RHICOMPL-502 - Add tailoring and selected filter to rules table

dLobatog:
Screenshot from 2020-03-06 14-01-27]

Still a draft as I want to polish the code duplication and write tests.

### DIFF
--- a/packages/inventory-compliance/src/SystemRulesTable.js
+++ b/packages/inventory-compliance/src/SystemRulesTable.js
@@ -38,6 +38,7 @@ class SystemRulesTable extends React.Component {
         itemsPerPage: this.props.itemsPerPage,
         rows: [],
         hidePassed: this.props.hidePassed,
+        selectedFilter: this.props.selectedFilter,
         severity: [],
         policy: [],
         currentRows: [],
@@ -58,7 +59,7 @@ class SystemRulesTable extends React.Component {
     }
 
     setInitialCurrentRows() {
-        const { hidePassed, itemsPerPage, severity, policy } = this.state;
+        const { hidePassed, itemsPerPage, severity, policy, selectedFilter } = this.state;
         let { profileRules } = this.props;
         const rowsRefIds = this.rulesToRows(profileRules);
 
@@ -69,8 +70,8 @@ class SystemRulesTable extends React.Component {
             profiles: rowsRefIds.profiles,
             refIds: rowsRefIds.refIds
         }, () => {
-            if (hidePassed) {
-                this.updateFilter(hidePassed, severity, policy);
+            if (hidePassed || selectedFilter) {
+                this.updateFilter(hidePassed, severity, policy, selectedFilter);
             }
         });
     }
@@ -167,6 +168,7 @@ class SystemRulesTable extends React.Component {
     }
 
     onSelect = (_event, selected, key) => {
+        const { remediationsEnabled, tailoringEnabled } = this.props;
         const { columnIndices, page, itemsPerPage } = this.state;
         let { currentRows, rows } = this.state;
         if (key === -1) {
@@ -175,7 +177,10 @@ class SystemRulesTable extends React.Component {
         } else {
             // One rule was selected
             const index = ((page - 1) * itemsPerPage * 2) + Number(key);
-            if (currentRows[index].cells[columnIndices.REMEDIATIONS_COLUMN].original) {
+            if (remediationsEnabled && currentRows[index].cells[columnIndices.REMEDIATIONS_COLUMN].original) {
+                rows[index].selected = selected;
+                currentRows[key].selected = selected;
+            } else if (tailoringEnabled) {
                 rows[index].selected = selected;
                 currentRows[key].selected = selected;
             }
@@ -227,9 +232,11 @@ class SystemRulesTable extends React.Component {
 
     calculateParent = ({ profile }, rule) => {
         const { columns } = this.state;
+        const { selectedRefIds, tailoringEnabled } = this.props;
 
         return {
             isOpen: false,
+            selected: tailoringEnabled ? selectedRefIds.includes(rule.refId) : undefined,
             cells: this.columnsToTableHeaders(columns, profile, rule).filter(e => e)
         };
     }
@@ -361,10 +368,10 @@ class SystemRulesTable extends React.Component {
     }
 
     handleSearch = debounce(searchTerm => {
-        const { hidePassed, severity, policy } = this.state;
+        const { hidePassed, severity, policy, selectedFilter } = this.state;
         return this.setState({
             searchTerm
-        }, () => this.updateFilter(hidePassed, severity, policy));
+        }, () => this.updateFilter(hidePassed, severity, policy, selectedFilter));
     }, 500)
 
     onSort = (_event, index, direction) => {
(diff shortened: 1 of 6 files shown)


packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -454,39 +468,49 @@ class SystemRulesTable extends React.Component {
             result = passedRows;
         }
 
-        if (policy.length > 0 && result.length > 0) {
```
dLobatog: This has been generalized into `appendToFilteredRows`

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -175,13 +177,13 @@ class SystemRulesTable extends React.Component {
         } else {
             // One rule was selected
             const index = ((page - 1) * itemsPerPage * 2) + Number(key);
-            if (currentRows[index].cells[columnIndices.REMEDIATIONS_COLUMN].original) {
+            if (tailoringEnabled || remediationsEnabled && currentRows[index].cells[columnIndices.REMEDIATIONS_COLUMN].original) {
```
bastilian: I'm a bit confused. `tailoringEnabled || remediationsEnabled` also happens down here as well: https://github.com/RedHatInsights/frontend-components/pull/392/files#diff-49f1e6a71fee60aaf0a07a8a1f8511aaR598

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -38,6 +38,8 @@ class SystemRulesTable extends React.Component {
         itemsPerPage: this.props.itemsPerPage,
         rows: [],
         hidePassed: this.props.hidePassed,
+        selectedFilter: this.props.selectedFilter,
+        selectedRefIds: this.props.selectedRefIds,
```
dLobatog: There are *many occasions* where the `rows` or `currentRows` will be empty arrays or will just not contain all of the rules. For example, if you search for something that doesn't exist, or if you select some severity filter.

Because of that, we cannot keep the state for the selected rows in neither `currentRows`, or `rows`. Instead, I'm keeping an array of refIds that are selected, which is fairly easy to update, and also to send to redux-form.

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -58,19 +60,19 @@ class SystemRulesTable extends React.Component {
...
             originalRows: rowsRefIds.rows,
-            rows: rowsRefIds.rows,
+            rows: this.markSelectedRows(rowsRefIds.rows),
             profiles: rowsRefIds.profiles,
             refIds: rowsRefIds.refIds
         }, () => {
-            if (hidePassed) {
-                this.updateFilter(hidePassed, severity, policy);
+            if (hidePassed || selectedFilter) {
+                this.updateFilter(hidePassed, severity, policy, selectedFilter);
```
dLobatog: If we pass the `selectedFilter` enabled from the create policy wizard, all of the pre-selected rules will show up as selected

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -166,22 +168,46 @@ class SystemRulesTable extends React.Component {
...
     onSelect = (_event, selected, key) => {
+        const { remediationsEnabled, tailoringEnabled, handleSelect } = this.props;
         const { columnIndices, page, itemsPerPage } = this.state;
-        let { currentRows, rows } = this.state;
+        let { currentRows, rows, selectedRefIds } = this.state;
+
         if (key === -1) {
             rows = this.selectAll(rows, selected);
             currentRows = this.selectAll(currentRows, selected);
+            if (selected) {
```
dLobatog: This is just marking all rules as selected or unselected

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -166,22 +168,46 @@ class SystemRulesTable extends React.Component {
...
+            }
         } else {
             // One rule was selected
             const index = ((page - 1) * itemsPerPage * 2) + Number(key);
-            if (currentRows[index].cells[columnIndices.REMEDIATIONS_COLUMN].original) {
+            if (tailoringEnabled || remediationsEnabled && currentRows[index].cells[columnIndices.REMEDIATIONS_COLUMN].original) {
                 rows[index].selected = selected;
                 currentRows[key].selected = selected;
+
+                const currentRefId = this.rowRefId(rows[index]);
```
dLobatog: In here, I try to find the refId of the rule that was selected to add it to `selectedRefIds` or remove it, depending on whether `selected` is true or false.

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -166,22 +168,46 @@ class SystemRulesTable extends React.Component {
...
+                }
             }
         }
 
-        this.setState({ rows, currentRows });
+        this.setState({
+            rows,
+            currentRows,
+            selectedRefIds
+        }, () => handleSelect(selectedRefIds));
```
dLobatog: This `handleSelect` is a callback which sends this list to redux-form through https://github.com/RedHatInsights/compliance-frontend/pull/413 - it could serve for any other purpose in the future though

packages/inventory-compliance/src/SystemRulesTable.js
dLobatog: You're right - onSelect won't even be called if those were not true. However, if I remove this check here, it would be something like 
`if (currentRows[index].cells[columnIndices.REMEDIATIONS_COLUMN].original)`
which would not allow me to select rules without remediations during tailoring.

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -324,10 +350,10 @@ class SystemRulesTable extends React.Component {
     }
 
     selectedRules = () => {
-        const { currentRows, profiles, refIds, columnIndices } = this.state;
+        const { currentRows, profiles, columnIndices } = this.state;
         return currentRows.filter(row => row.selected && row.cells[columnIndices.REMEDIATIONS_COLUMN].original).map(row => ({
             // We want to match this response with a similar response from GraphQL
-            refId: refIds[row.cells[columnIndices.TITLE_COLUMN].original],
+            refId: this.rowRefId(row),
```
dLobatog: Just a convenience

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -473,25 +501,54 @@ class SystemRulesTable extends React.Component {
...
         const policyRows = this.filterBy(policy, originalRows, columnIndices.POLICY_COLUMN);
         const searchRows = this.searchBy(searchTerm, originalRows);
-        const filteredRows = this.filteredRows(passedRows, severityRows, policyRows, searchRows,
-            hidePassed, severity, policy, searchTerm);
+        const filteredRows = this.filteredRows(passedRows, severityRows, policyRows, searchRows, selectedRows,
+            hidePassed, severity, policy, searchTerm, selectedFilter);
 
         return this.setState({
-            currentRows: this.currentRows(1, itemsPerPage, { rows: filteredRows, profiles, refIds }),
+            currentRows: this.markSelectedRows(this.currentRows(1, itemsPerPage, { rows: filteredRows, profiles, refIds })),
```
dLobatog: Before displaying selected rows, let's mark them as selected based on the `selectedRefIds` array

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -595,6 +654,10 @@ SystemRulesTable.propTypes = {
     system: propTypes.object,
     itemsPerPage: propTypes.number,
     remediationsEnabled: propTypes.bool,
+    tailoringEnabled: propTypes.bool,
+    selectedRefIds: propTypes.array,
+    selectedFilter: propTypes.bool,
+    handleSelect: propTypes.handleSelect,
```
bastilian: `propTypes.handleSelect` should probably be `propTypes.func`

PR #396, opened 2020-03-11
[RFR] Only Display Checkboxes for Remediable Rules in Inventory-Insights

LI0131:
Aha! @AllenBW this PR (RHCLOUD-4801) will conditionally add the checkboxes and manage the onSelect manually within the table component. This should give us the functionality that we would like after I get the onSelect method working within the rows. We may also need to amend the number of selected systems during a bulk select.


AllenBW:
Nice work @LI0131 ! will review in the mor, bulk select already ignores ones without remediation 👍

LI0131:
Screen Shot 2020-03-19 at 11 02 11 AM

PR #398, opened 2020-03-12
Remove count numbers in chart legend

katerinapat:
for readability

before
Screenshot from 2020-03-12 13-40-31

after 
Screenshot from 2020-03-12 13-40-11


katerinapat:
updated the description with the screenshots

PR #400, opened 2020-03-12
Fixes warning message about rows[0] being invalid to Table

josejulio:
- Cells needs to be within the "cell" element, contents were on
   root of the rows object

warning:
```
react_devtools_backend.js:6 Warning: Failed prop type: Invalid prop `rows[0]` supplied to `Table`.
    in Table
```


josejulio:
@karelhala could you review? thanks!

PR #401, opened 2020-03-16
refactored like dislike buttons

Sergey1011010:
Added a callback for user's feedback, changed color and spacing between buttons

image
image
image
image

### DIFF
--- a/packages/rule-components/src/ReportDetails/RuleFeedback.js
+++ b/packages/rule-components/src/ReportDetails/RuleFeedback.js
@@ -14,9 +14,7 @@ const RuleFeedback = ({ ruleId, onFeedbackChanged }) => (
             className="ins-c-rule__rule-feedback-like-button"
             variant="plain"
             aria-label="Rule is helpful"
-            onClick={ () => {
-                onFeedbackChanged(ruleId, 1, '');
-            } }
+            onClick={ () => onFeedbackChanged(ruleId, 1, '') }
         >
             <ThumbsUpIcon/>
         </Button>
@@ -24,9 +22,7 @@ const RuleFeedback = ({ ruleId, onFeedbackChanged }) => (
             className="ins-c-rule__rule-feedback-like-button"
             variant="plain"
             aria-label="Rule is not helpful"
-            onClick={ () => {
-                onFeedbackChanged(ruleId, -1, '');
-            } }
+            onClick={ () => onFeedbackChanged(ruleId, -1, '') }
         >
             <ThumbsDownIcon/>
         </Button>
@@ -34,8 +30,8 @@ const RuleFeedback = ({ ruleId, onFeedbackChanged }) => (
 );
 
 RuleFeedback.propTypes = {
-    ruleId: PropTypes.number,
-    onFeedbackChanged: PropTypes.func
+    ruleId: PropTypes.string.isRequired,
+    onFeedbackChanged: PropTypes.func.isRequired
 };
 
 export default RuleFeedback;
--- a/packages/rule-components/src/ReportDetails/RuleFeedback.js
+++ b/packages/rule-components/src/ReportDetails/RuleFeedback.js
@@ -2,19 +2,40 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import ThumbsUpIcon from '@patternfly/react-icons/dist/js/icons/outlined-thumbs-up-icon';
 import ThumbsDownIcon from '@patternfly/react-icons/dist/js/icons/outlined-thumbs-down-icon';
+import { Button } from '@patternfly/react-core';
 
-const RuleFeedback = ({ ruleId }) => (
+// ruleId - is id of current rule
+// onFeedbackSent(ruleId, vote, text) is a callback which is called when feedback is changed
+// where ruleId is id of the rule, vote is either -1, 0 or 1 and text is a string
+const RuleFeedback = ({ ruleId, onFeedbackChanged }) => (
     <div>
-        <div>
-            <span>Is this rule helpful?</span>
+        <span>Is this rule helpful?</span>
+        <Button
+            className="ins-c-rule__rule-feedback-like-button"
+            variant="plain"
+            aria-label="Rule is helpful"
+            onClick={ () => {
+                onFeedbackChanged(ruleId, 1, '');
+            } }
+        >
             <ThumbsUpIcon/>
+        </Button>
+        <Button
+            className="ins-c-rule__rule-feedback-like-button"
+            variant="plain"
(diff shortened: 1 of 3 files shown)


packages/rule-components/src/ReportDetails/RuleFeedback.js
```diff
@@ -2,19 +2,40 @@ import React from 'react';
...
             <ThumbsDownIcon/>
-        </div>
+        </Button>
     </div>
 );
 
 RuleFeedback.propTypes = {
-    ruleId: PropTypes.number
+    ruleId: PropTypes.number,
+    onFeedbackChanged: PropTypes.func
```
bond95: I think here should be
```
ruleId: PropTypes.number.isRequired,
onFeedbackChanged: PropTypes.func.isRequired,
```

packages/rule-components/src/ReportDetails/RuleFeedback.js
```diff
@@ -2,19 +2,40 @@ import React from 'react';
...
     <div>
-        <div>
-            <span>Is this rule helpful?</span>
+        <span>Is this rule helpful?</span>
+        <Button
+            className="ins-c-rule__rule-feedback-like-button"
+            variant="plain"
+            aria-label="Rule is helpful"
+            onClick={ () => {
+                onFeedbackChanged(ruleId, 1, '');
```
bond95: Why do you have here 3 arguments, while use only 2?

packages/rule-components/src/ReportDetails/RuleFeedback.js
```diff
@@ -2,19 +2,40 @@ import React from 'react';
...
+            } }
```
bond95: It can be slightly simplified `onClick={() => onFeedbackChanged(ruleId, 1)}`

packages/rule-components/src/ReportDetails/index.js
```diff
@@ -38,7 +39,7 @@ const ReportDetails = (
                     </div>
                 </StackItem>
                 <StackItem>
-                    <RuleFeedback ruleId={ ruleId }/>
+                    <RuleFeedback ruleId={ ruleId } onFeedbackChanged={ onFeedbackChanged }/>
```
bond95: I think it will be nice to hide this item in case if `onFeedbackChanged ` is `undefined`

packages/rule-components/src/ReportDetails/RuleFeedback.js
Sergey1011010: text feedback my be added at some point

bond95: I think feedback should be done not in this component, this component is only two buttons, and some text, nothing more

Sergey1011010: Why not? It's called RuleFeedback and I think it's the best place for text feedback as well. We can though remove third arg now because it's not used.

Sergey1011010:
@karelhala

packages/rule-components/src/ReportDetails/index.scss
```diff
@@ -9,5 +9,9 @@
   .ins-c-rule__risk-detail {
     margin-bottom: var(--pf-global--spacer--sm);
   }
+
+  .ins-c-rule__rule-feedback-like-button, .ins-c-rule__rule-feedback-dislike-button {
+    padding: 0 5px;
```
karelhala: Shouldn't this be `padding: 0 var(--pf-global--spacer--xs)`?

karelhala:
ping @doruskova

packages/rule-components/src/ReportDetails/index.scss
Sergey1011010: fixed

doruskova:
Thank you for updating this :) Just a few things that I noticed:

- "Total risk" and "Risk of change" title should have normal `font-weight: 400` with a specific color e.g. `#F0AB00`
- The status of a Total risk and Risk of change should begin with uppercase. 
- The description below these titles should have 14px with text color `#737679`
- The risk of change is using a different icon.  
- The external link on the left should always begin below description and there should be a 24px spacer between description and the KCS link. The same for KCS link and thumbs up and down icons. 
- There should be a bigger gap between the "thumbs up" and down" icons. It should be wrapped in a button, so the icon itself has 16x16, but in <button> it takes 32x40. 
- The status "Important" in a row should be aligned with the text in other columns. (see picture below)
76770243-e227b180-679d-11ea-9241-c22fba8354f0

bond95:
> @Sergey1011010 Thank you for updating this :) Just a few things that I noticed:
> 
> * "Total risk" and "Risk of change" title should have normal `font-weight: 400` with a specific color e.g. `#F0AB00`
> * The status of a Total risk and Risk of change should begin with uppercase.
> * The description below these titles should have 14px with text color `#737679`
> * The risk of change is using a different icon.
> * The external link on the left should always begin below description and there should be a 24px spacer between description and the KCS link. The same for KCS link and thumbs up and down icons.
> * There should be a bigger gap between the "thumbs up" and down" icons. It should be wrapped in a button, so the icon itself has 16x16, but in  it takes 32x40.
> * The status "Important" in a row should be aligned with the text in other columns. (see picture below)
>   76770243-e227b180-679d-11ea-9241-c22fba8354f0

@doruskova I think most of this points are not related to likes and dislikes, so, please, write them to main PR on gitlab, thank you 🙂

doruskova:
> > @Sergey1011010 Thank you for updating this :) Just a few things that I noticed:
> > 
> > * "Total risk" and "Risk of change" title should have normal `font-weight: 400` with a specific color e.g. `#F0AB00`
> > * The status of a Total risk and Risk of change should begin with uppercase.
> > * The description below these titles should have 14px with text color `#737679`
> > * The risk of change is using a different icon.
> > * The external link on the left should always begin below description and there should be a 24px spacer between description and the KCS link. The same for KCS link and thumbs up and down icons.
> > * There should be a bigger gap between the "thumbs up" and down" icons. It should be wrapped in a button, so the icon itself has 16x16, but in  it takes 32x40.
> > * The status "Important" in a row should be aligned with the text in other columns. (see picture below)
> >   76770243-e227b180-679d-11ea-9241-c22fba8354f0
> 
> @doruskova I think most of this points are not related to likes and dislikes, so, please, write them to main PR on gitlab, thank you 🙂

@bond95 ok, it's there now too:)

Sergey1011010:
@doruskova, something like this?(see updated images above)

doruskova:
> @doruskova, something like this?(see updated images above)

yes! this is ok :) Thanks!

PR #403, opened 2020-03-16
Add delete option to EntityDetail

fhlavac:
- added a delete option to EntityDetail

@karelhala

### DIFF
--- a/packages/inventory/src/DeleteModal.js
+++ b/packages/inventory/src/DeleteModal.js
@@ -13,14 +13,14 @@ import {
 } from '@patternfly/react-core';
 import { ExclamationTriangleIcon } from  '@patternfly/react-icons';
 
-const DeleteModal = ({ handleModalToggle, isModalOpen, currentSytem, onConfirm }) => {
+const DeleteModal = ({ handleModalToggle, isModalOpen, currentSytems, onConfirm }) => {
     let systemToRemove;
     let systemLabel = 'system';
-    if (Array.isArray(currentSytem)) {
-        systemToRemove = currentSytem.length === 1 ? currentSytem[0].display_name : `${currentSytem.length} systems`;
-        systemLabel = currentSytem.length === 1 ? systemLabel : 'systems';
+    if (Array.isArray(currentSytems)) {
+        systemToRemove = currentSytems.length === 1 ? currentSytems[0].display_name : `${currentSytems.length} systems`;
+        systemLabel = currentSytems.length === 1 ? systemLabel : 'systems';
     } else {
-        systemToRemove = currentSytem.display_name;
+        systemToRemove = currentSytems.display_name;
     }
 
     return <Modal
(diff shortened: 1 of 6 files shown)


karelhala:
No need to commit `package-lock.json` file. Looks like the problem with failing tests is related to it.

packages/inventory/src/EntityDetail.js
```diff
@@ -82,6 +93,38 @@ class EntityDetails extends Component {
...
                 }
+                { this.state.isModalOpen && (
+                    <DeleteModal
+                        handleModalToggle={this.handleModalToggle}
+                        isModalOpen={this.state.isModalOpen}
+                        currentSytem={entity}
+                        onConfirm={() => {
+                            let displayName;
+                            let removeSystems;
+                            if (Array.isArray(entity)) {
```
karelhala: No need to check if entity is array, this will always be single entity in detail, you can leave just the else path.

packages/notifications/src/redux/action-types.js
```diff
@@ -2,3 +2,4 @@ const notificationsPrefix = '@@INSIGHTS-CORE/NOTIFICATIONS/';
 export const ADD_NOTIFICATION = `${notificationsPrefix}ADD_NOTIFICATION`;
 export const REMOVE_NOTIFICATION = `${notificationsPrefix}REMOVE_NOTIFICATION`;
 export const CLEAR_NOTIFICATIONS = `${notificationsPrefix}CLEAR_NOTIFICATIONS`;
+export const REMOVE_ENTITY = `${notificationsPrefix}REMOVE_ENTITY`;
```
karelhala: Wrong file and wrong type. It should be in `/packages/inventory/src/redux/action-types` in the list of `asyncInventory` as new array element.

packages/inventory/src/EntityDetail.js
```diff
@@ -46,9 +50,16 @@ class EntityDetails extends Component {
...
+        this.setState({
+            isModalOpen: !this.state.isModalOpen
+        });
+    };
+
     generateTop = () => {
-        const { entity, loaded, actions } = this.props;
+        const { entity, loaded, actions, deleteEntity, addNotification } = this.props;
         const { isOpen } = this.state;
+        const inventoryActions = [{ onClick: this.handleModalToggle, title: 'Delete' }, ... actions || [] ];
```
rvsia: Should not the button be red as in Sources to be consistent?

Hyperkid123: Nope PF now says that destructive actions in dropdowns should have regular color.

rvsia: Ahhh....

packages/inventory/src/DeleteModal.js
```diff
@@ -0,0 +1,81 @@
...
+    LevelItem,
+    Button,
+    ClipboardCopy
+} from '@patternfly/react-core';
+import { ExclamationTriangleIcon } from  '@patternfly/react-icons';
+
+const DeleteModal = ({ handleModalToggle, isModalOpen, currentSytem, onConfirm }) => {
+    let systemToRemove;
+    let systemLabel = 'system';
+    if (Array.isArray(currentSytem)) {
```
PanSpagetka: According to @karelhala comment below, it should be enough to also leave only the else path

karelhala: Not really, here we want to allow multiple systems for deletion, so we can reuse this component on inventory table.

fhlavac: Discussed it with @karelhala and if I understand, DeleteModal component should be also used in the inventory table, where the isArray option will be needed

packages/inventory/src/index.js
```diff
@@ -1,3 +1,4 @@
 export { default as Inventory, inventoryConnector } from './Inventory';
 export * from './redux/reducers';
 export { INVENTORY_ACTION_TYPES, SELECT_ENTITY } from './redux/action-types';
+export { default } from './DeleteModal';
```
karelhala:
```suggestion
export { default as DeleteModal } from './DeleteModal';
```

PR #412, opened 2020-03-19
Add source filter to inventory

karelhala:
Screenshot from 2020-03-19 17-06-20
Screenshot from 2020-03-19 17-06-28


karelhala:
ping @bclarhk the API currently supports only insights source.

PR #415, opened 2020-03-20
[RFR] Set row collapse to not change on filtering/sorting

LI0131:
This PR is in response to https://projects.engineering.redhat.com/browse/RHCLOUD-5246 (on the last page of the attached google doc). This sets the rows to be open by default and then maintains the collapsed setting through filtering and sorting.


packages/inventory-insights/src/Insights.js
```diff
@@ -135,11 +135,14 @@ class InventoryRuleList extends Component {
                 return rowKey % 2 === 0 && rowVal.rule.rule_id === rule.rule_id && rowVal.selected;
             });
             selected = selected.length ? true : false;
+            const isOpen = rows.filter((rowVal, rowKey) => {
```
AllenBW: so we do this same thing on  line 134, the map through rows... smells like an opportunity to consolidate the.... 🤔

PR #416, opened 2020-03-23
[RFR] Add Label to Battery in Insights-Inventory

LI0131:
@AllenBW This is in response to RHCLOUD-5246 on page 2 of the google doc. We're adding the label to the battery icon.

### DIFF
--- a/packages/inventory-insights/src/Insights.js
+++ b/packages/inventory-insights/src/Insights.js
@@ -1,7 +1,7 @@
 /* eslint-disable camelcase */
 import './insights.scss';
 
-import { BASE_FETCH_URL, FILTER_CATEGORIES as FC } from './Constants';
+import { BASE_FETCH_URL, FILTER_CATEGORIES as FC, RISK_TO_STRING } from './Constants';
 import React, { Component, Fragment } from 'react';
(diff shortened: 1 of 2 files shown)


packages/inventory-insights/src/Insights.js
```diff
@@ -128,6 +128,12 @@ class InventoryRuleList extends Component {
 
     buildRows = (activeReports, kbaDetails, filters, rows, searchValue = '', kbaLoading = false) => {
         const builtRows = flatten(activeReports.map((value, key) => {
+            const riskToString = {
```
AllenBW: @LI0131 while there's not many other places we might be usin this, lets throw this in `constants.js` just cuz its the kinda thing that should be in there <3

packages/inventory-insights/src/Insights.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable camelcase */
 import './insights.scss';
 
-import { BASE_FETCH_URL, FILTER_CATEGORIES as FC } from './Constants';
+import { BASE_FETCH_URL, FILTER_CATEGORIES as FC, RISK_TO_STRING } from './Constants';
```
AllenBW: love this name. 😆 🎀

PR #419, opened 2020-03-25
Add rhsm detail information

karelhala:
### UI improvements
This will add new values to general information tab populated from rhsm facts (if they are found)
Screenshot from 2020-03-25 11-05-11

### Babel improvemens
I've included `@babel/plugin-proposal-optional-chaining` in order to easier map nested properties in objects that could be undefined and updated `@babel/*` packages.


packages/inventory-general-info/src/OperatingSystemCard.test.js
```diff
@@ -41,6 +47,21 @@ describe('OperatingSystemCard', () => {
...
                 }
+            }, entityDetails: {
+                entity: {}
+            }
+        });
+        const wrapper = render(<OperatingSystemCard store={ store } />);
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+
+    it.only('should render correctly with rhsm facts', () => {
```
rvsia: only :(

packages/inventory-general-info/src/SystemCard.test.js
```diff
@@ -50,6 +53,19 @@ describe('SystemCard', () => {
         expect(toJson(wrapper)).toMatchSnapshot();
     });
 
+    it.only('should render correctly with rhsm facts', () => {
```
rvsia: only :(

packages/inventory-general-info/src/BiosCard.js
```diff
@@ -4,14 +4,18 @@ import { connect } from 'react-redux';
...
+import { DateFormat } from '@redhat-cloud-services/frontend-components';
 
 const BiosCard = ({ bios, detailLoaded, handleClick }) => (<LoadingCard
     title="BIOS"
     isLoading={ !detailLoaded }
     items={ [
         { title: 'Vendor', value: bios.vendor },
         { title: 'Version', value: bios.version },
-        { title: 'Release date', value: bios.releaseDate },
+        { title: 'Release date', value: (DateFormat ?
```
rvsia: Just curious: is this check done because the application using this package does not need to have the frontend-components package?

babel.config.js
```diff
@@ -61,6 +61,7 @@ module.exports = {
         'babel-plugin-lodash',
         '@babel/plugin-transform-react-display-name',
         '@babel/plugin-proposal-class-properties',
+        '@babel/plugin-proposal-optional-chaining',
```
Hyperkid123: You just need to bump babel-core and babel preset env and babel runtime (maybe other non-plugin babel packages). The plugin is included in babel version >=7.8.4

package.json
```diff
@@ -67,15 +67,16 @@
...
-        "@babel/plugin-syntax-dynamic-import": "^7.2.0",
-        "@babel/plugin-transform-react-display-name": "^7.2.0",
-        "@babel/plugin-transform-runtime": "^7.4.3",
-        "@babel/preset-env": "^7.4.3",
-        "@babel/preset-react": "^7.0.0",
-        "@babel/runtime": "^7.4.3",
+        "@babel/plugin-proposal-class-properties": "^7.8.3",
+        "@babel/plugin-proposal-decorators": "^7.8.3",
+        "@babel/plugin-proposal-object-rest-spread": "^7.9.0",
+        "@babel/plugin-proposal-optional-chaining": "^7.9.0",
```
Hyperkid123: ^^

packages/inventory-general-info/src/BiosCard.js
Hyperkid123: Not exactly. The date format component is not in an older version of frontend-components package and it could just blow up.

rvsia: Should not it be set as a peer dependency then?

karelhala: Not really, it is set as dependency, so it'll be installed all the times. However if the app has older version of it this could blow up.

babel.config.js
karelhala: Yeah, good point. I guess we'll keep it as is for now and remove this in future. What are your thoughts?

Hyperkid123: If you want we can keep it. But I think it's not doing anything at this point.

PR #420, opened 2020-03-25
Disable remediate button if user has incorrect permissions

karelhala:
Screenshot from 2020-03-25 14-36-10


karelhala:
release

ryelo:
@karelhala we may need to make updates to this, because a user may have:
`remediations:remediation:write`
`remediations:*:*`
`remediations:*:write`

PR #425, opened 2020-03-28
Use PrimaryToolbar for SystemRulesTable

bastilian:
This replaces the custom toolbar for the `SystemRulesTable` with the `PrimaryToolbar` component in order to add filters with chips:

Screenshot 2020-03-28 at 21 01 56


bastilian:
@akofink The sorting by severity will now sort High to Unknown and the name filter wil now fire on typing. However on the SystemsTable in compliance-frontend searching onchange is disabled and will still require pressing enter, as it causes issues with filters and chips getting out of sync and breaking their deletion. We are only searching objects on the Rules table, while on the SystemsTable we also fetch from our API as well as the inventory API. I'll investigate how we can also search on typing there as well, but for now will let it as is atm.

I tried to reproduce the filter checkboxes not updating, but this works for me as expected:

Kapture 2020-04-01 at 13 17 44

bastilian:
@akofink Thank you for testing it again! This and the accompanying PR on compliance-frontend is ready for another review. The issues are addressed and tested thoroughly.

The pages where this is used in compliance are:

* Policies Table on policy overview
* Rules Table in Create Policy Wizard
* Rules Table on policy page
* Systems Table on policy page
* Systems Table on systems page
* Rules Table on system page
* Systems Table 0n Reports page

packages/inventory-compliance/src/Utilities/FilterConfigBuilder.js
```diff
@@ -0,0 +1,191 @@
...
+        return initialState;
+    }
+
+    categoryLabelForValue = (value) => {
+        const category = this.config.filter((category) => (
+            category.items ? category.items.map((item) => item.value).includes(value) : false
+        ))[0];
+
+        return category ? category.label : value;
+    };
```
akofink, line 119: I'm not sure this is the problem area, but `category` is some object rather than a string. It's causing this warning and probably the behavior I'm seeing:

```
Warning: Encountered two children with the same key, `.$group_Severity_chip_[object Object]`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

bastilian: I know about this, there is no way I've seen so far to override the keys that the filterships generated.

bastilian:
@akofink I'm not sure why, but it looks like you are using an older version of the PrimaryToolbar/ConditionalFilter/FilterChips components from the `frontend-components` package. You shouldn't see the label for each filter when the select is closed, nor should the "Clear all filters" linger around.

PR #426, opened 2020-03-31
Use isEqual instead of JSON.stringify

PanSpagetka:
Uselodash  isEqual instead of JSON.stringify

### DIFF
--- a/packages/inventory/src/InventoryList.js
+++ b/packages/inventory/src/InventoryList.js
@@ -7,6 +7,7 @@ import PropTypes from 'prop-types';
 import './InventoryList.scss';
 import { InventoryContext } from './Inventory';
 import { CancelToken } from 'axios';
+import isEqual from 'lodash/isEqual';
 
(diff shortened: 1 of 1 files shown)


packages/inventory/src/InventoryList.js
```diff
@@ -52,9 +54,9 @@ class ContextInventoryList extends React.Component {
 
     componentDidUpdate(prevProps) {
         const { items, hasItems, sortBy } = this.props;
-        if (hasItems && (JSON.stringify(items) !== JSON.stringify(prevProps.items))) {
+        if (hasItems && (isEqual(items, prevProps.items))) {
```
karelhala: These should be negated
```
if (hasItems && (!isEqual(items, prevProps.items))) {
```
} else if (!hasItems && !isEqual(prevProps.sortBy, sortBy)) {
```

```

PanSpagetka: Oh...good catch!

PR #429, opened 2020-04-01
[RFR] Add Paragraph component to pdf-generator

LI0131:
@AllenBW This component allows the user to add paragraph-like text blocks to their pdfs.


packages/pdf-generator/src/components/Paragraph.js
```diff
@@ -0,0 +1,45 @@
...
+
+const appliedStyles = styles();
+
+const Paragraph = ({
+    style,
+    children,
+    paragraphProps,
+    ...props
+}) => {
+    return (
```
AllenBW, line 17: ah can simplify this to just return the html block

`Paragraph = ({
    style,
    children,
    paragraphProps,
    ...props
}) => <View`

PR #430, opened 2020-04-02
Use inventory page from parent if used in different app

karelhala:
This PR fixes problem, when if used in application and user changes page and after that updates per page count the page number remains the same even though the data are clearly rom first page.


karelhala:
@Razorlazer this fixes the issue in vulnerability

PR #434, opened 2020-04-02
Use score from backend in SystemPolicyCard

akofink:
This should be the last place where score was still being calculated from passed/failed rules.

We are for sure going to get bugs reported against this now that the math isn't so simple to figure out. Check out what looks like a discrepancy between the score and pass/fail rule count.

```
18 of 51 rules passed (78%)
43 of 141 rules passed (63%)
```

image

I'm not sure why I needed to update snapshots not related to what I changed, but they were failing. Any guidance would be appreciated.

### DIFF
--- a/packages/inventory-compliance/src/Compliance.js
+++ b/packages/inventory-compliance/src/Compliance.js
@@ -39,6 +39,7 @@ query System($systemId: String!){
             rulesFailed
             rulesPassed
             lastScanned
+            score
             rules {
                 title
                 severity
--- a/packages/inventory-compliance/src/SystemPolicyCard.js
+++ b/packages/inventory-compliance/src/SystemPolicyCard.js
@@ -52,9 +52,9 @@ class SystemPolicyCard extends React.Component {
     }
 
     render() {
-        const { rulesPassed, rulesFailed, compliant, lastScanned } = this.state.policy;
+        const { rulesPassed, rulesFailed, compliant, lastScanned, score } = this.state.policy;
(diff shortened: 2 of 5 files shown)


packages/inventory-compliance/src/Compliance.js
```diff
@@ -39,6 +39,7 @@ query System($systemId: String!){
             rulesFailed
             rulesPassed
             lastScanned
+            score
```
akofink, line 42: From https://github.com/RedHatInsights/compliance-backend/pull/375 (merged)

packages/components/src/Components/CullingInfo/__snapshots__/CullingInformation.test.js.snap
```diff
@@ -54,7 +54,7 @@ exports[`CullingInformation should render with dates - danger 1`] = `
     aria="describedby"
     boundary="window"
     className=""
-    content="This system is scheduled for removal from inventory in 0 days (05 Feb 2020 00:00:00 UTC)"
+    content="This system is scheduled for removal from inventory in 0 days (05 Feb 2020 05:00:00 UTC)"
```
dLobatog, line 57: Tests are failing due to these changes - https://travis-ci.org/github/RedHatInsights/frontend-components/builds/670212631?utm_source=github_status&utm_medium=notification - you can run `npm run test packages/inventory-compliance/src -- -u` if you just want to update 1 dir

akofink:
Updated to remove the rogue snap change

PR #435, opened 2020-04-02
Add RBAC helpers to utilities

boaz0:
Hi @karelhala @ryelo @rvsia 

I thought maybe I can share this here as other projects are going to use or are already using RBAC.
Let me know what you think.

Thanks.


rvsia:
Can you please write a minimal example how to use it properly? :pray: Thanks!

packages/utils/src/RBAC.js
```diff
@@ -0,0 +1,93 @@
...
+    ) {
+        const user = await insights.chrome.auth.getUser();
+        const permissions = await insights.chrome.getUserPermissions(applicationName);
+        return {
+            isOrgAdmin: user.identity.user.is_org_admin,
+            permissions
+        };
+    }
+
+    return { isOrgAdmin: false, permissions: null };
```
karelhala, line 19: This can be even more simplified
```JS
export async function getRBAC(applicationName = '') {
  const user = await insights?.chrome?.auth?.getUser()
  return {
    isOrgAdmin: user?.identity?.user?.is_org_admin || false
    permissions: (await insights?.chrome?.getUserPermissions(applicationName)) || null
  }
}
```

packages/utils/src/RBAC.js
```diff
@@ -0,0 +1,93 @@
...
+}
+
+export function doesHavePermissions(RBACResults, permissionList) {
+    const { isOrgAdmin, permissions } = RBACResults;
+    if (isOrgAdmin) {
```
karelhala, line 24: This is not quite correct, if user is Org admin they might not have access to something based on the permission.

packages/utils/src/RBAC.js
```diff
@@ -0,0 +1,93 @@
...
+        return {
+            isOrgAdmin: user.identity.user.is_org_admin,
+            permissions
+        };
+    }
+
+    return { isOrgAdmin: false, permissions: null };
+}
+
+export function doesHavePermissions(RBACResults, permissionList) {
```
karelhala, line 22: I wouldn't mix orgAdmin and permission together when checking for permission access. User can have access to some parts if he's not org admin but has permissions nad vie versa.

```JS
export function doesHavePermissions(userPermissions, permissionList) {
  return userPermissions.some((access) => permissionList.includes(access.permission));
}
```

packages/utils/src/RBAC.js
```diff
@@ -0,0 +1,93 @@
...
+    if ([ null, undefined ].includes(permissions)) {
+        return false;
+    }
+
+    return permissions.some(function(access) {
+        return permissionList.includes(access.permission);
+    });
+}
+
+export const RBAC_ACTION_TYPES = { PENDING: 'RBAC_PENDING', SUCCESS: 'RBAC_SUCCESS', FAILURE: 'RBAC_FAILURE' };
```
karelhala, line 37: Can you move the redux stuff to specific file? This way we won't import them if not needed. Also I would prefer hooks in here rather than redux.

```JS
import { useState, useEffect } from 'react';
import { getRBAC, doesHavePermissions } from './RBAC';

export function usePermissions(appName, permissionsList) {
  const [permissions, setPermissions] = useState({});
  useEffect(() => {
    (async () => {
        const { isOrgAdmin, permissions: userPermissions } = await getRBAC(appName);
        setPermissions({
           isOrgAdmin,
           permissions: userPermissions,
           hasAccess: doesHavePermissions(userPermissions, permissionsList)
        });
    })()
  }, [ appName ])
  return permissions;
}
```
And usage would be
```JSX
import React from 'react'
import { usePermissions } from '@redhat-cloud-services/frontend-components-utilities/files/permissions';

export const () => {
  const { hasAccess } = usePermissions('inventory', [ 'inventory:detail:read' ]);
  return <div>{hasAccess && <span>User has access</span>}</div>;
}
```

packages/utils/src/RBAC.test.js
```diff
@@ -0,0 +1,116 @@
+import { getRBAC, RBACAction, RBACStatusSelector, RBACDataSelector, RBACErrorSelector, RBAC_KEY, doesHavePermissions, RBACReducer, RBAC_ACTION_TYPES } from './RBAC';
+
+function setUpInsights(getUser = null) {
+    const value = getUser === null ? null : {
```
karelhala, line 4: There's `setupTests` file that can be used to mock these functions.

packages/utils/src/RBAC.js
```diff
@@ -0,0 +1,21 @@
+
+export async function getRBAC(applicationName = '') {
+    const insights = window.insights;
+    const user = await insights?.chrome?.auth?.getUser();
+    const r = {
```
karelhala, line 5: No need to assign to a value here.

boaz0:
Thanks @karelhala 
@rvsia after these changes, do you still think a readme file is required?

karelhala:
@boaz0 you also need to add
```
RBAC: './src/RBAC.js',
RBACHook: './src/RBACHook.js',
```
to `packages/utils/config/webpack.config.js` in `entry` field in order to properly build these files.

PR #438, opened 2020-04-03
Made few small fixes for rule-components

bond95:
Fixed sortBy prop in RuleTable
Hide Risk of Change if value not in range (1,4)

Fixes: https://github.com/RedHatInsights/insights-results-aggregator/issues/466

image

### DIFF
--- a/packages/rule-components/src/RuleTable/RuleTable.js
+++ b/packages/rule-components/src/RuleTable/RuleTable.js
@@ -143,15 +143,9 @@ class RuleTable extends Component {
             })
         }));
 
-        const sortProps = sortBy ? { sortBy, onSort: this.onSort } : undefined;
-
         return <section className="rhcs-c-rules-table">
             <PrimaryToolbar
-                {
-                    ...filterItems &&
-                    filterItems.length > 0 &&
-                    { filterConfig: { items: filterItems } }
-                }
+                { ...filterItems && filterItems.length > 0 && { filterConfig: { items: filterItems } }}
                 activeFiltersConfig={{
                     filters: Object.entries(filterValues).map(([ key, value ]) => calculateActiveFilters(
                         filterItems.find(({ filterKey }) => filterKey === key),
@@ -165,7 +159,7 @@ class RuleTable extends Component {
             />
             <Table
                 {...props}
-                {...sortProps}
+                {...sortBy && { sortBy, onSort: this.onSort }}
                 actions={!isLoading && actions}
                 aria-label={ariaLabel}
                 cells={columns.map(({ transforms, ...column }) => ({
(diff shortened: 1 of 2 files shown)


packages/rule-components/src/RuleTable/RuleTable.js
```diff
@@ -143,9 +143,15 @@ class RuleTable extends Component {
             })
         }));
 
+        const sortProps = sortBy ? { sortBy, onSort: this.onSort } : undefined;
+
         return <section className="rhcs-c-rules-table">
             <PrimaryToolbar
-                { ...filterItems && filterItems.length > 0 && { filterConfig: { items: filterItems } }}
```
karelhala, line 148: No need to align this line.

packages/rule-components/src/RuleTable/RuleTable.js
```diff
@@ -159,7 +165,7 @@ class RuleTable extends Component {
             />
             <Table
                 {...props}
-                {...sortBy && { onSort: this.onSort }}
+                {...sortProps}
```
karelhala, line 168: No need to assign to a constant
```JSX
<Table
  { ...sortBy && { ...sortBy, onSort: this.onSort } }
```

bond95: I think it's better to leave it as it is, because otherwise it's hard to understand.

packages/rule-components/src/RuleTable/RuleTable.js
bond95: fixed

packages/rule-components/src/RuleTable/RuleTable.js
karelhala: That's not hard to understand, it is used on many other places and this way we are not using constant that is used on one place.

bond95: Well, I’m not sure about that, because when I tried to find what was broken, I find out that `onSort` just didn’t insert into RuleTable props, so it was not so obvious construction.

karelhala: anybody from @RedHatInsights/ui-reviewers any thoughts on this?

rvsia: I am understanding what bond95 means but Karel's construction is really simple - the difference in readability is really small in this case, so I would go with that as it is a suggestion from a maintainer, who will take over this code.

>  I find out that onSort just didn’t insert into RuleTable props,

Actually I would say this would be faster to find in the Karel's solution. Using the separate constant, you have to go up and look, what it is. Otherwise the code is basically the same.

bond95: @rvsia thank you for your response. Ok, 2 vs 1, so I'll return it as it was.

bond95: @karelhala @rvsia done

PR #440, opened 2020-04-05
Add pagination, filter and select to tags modal

karelhala:
### Pagination and filter in system's tags
Screenshot from 2020-04-07 14-49-46

### No tags found for tag's filter
Screenshot from 2020-04-07 14-49-55

### All tags in inventory table, selecting and confirming will apply these tags to filter
Screenshot from 2020-04-07 14-50-06


packages/components/src/Components/TagModal/TagModal.js
```diff
@@ -1,15 +1,47 @@
...
+import { SkeletonTable } from '../SkeletonTable';
+
+const calculateChecked = (rows = [], selected) => (
+    rows.every(({ id }) => selected && selected.find(({ id: selectedId }) => selectedId === id))
+        ? rows.length > 0
+        : rows.some(({ id }) => selected && selected.find(({ id: selectedId }) => selectedId === id)) && null
+);
+
+const unique = (arr) => {
+    return arr.filter(({ id }, index, arr) => arr.findIndex(({ id: currId }) => currId === id) === index);
```
Hyperkid123, line 33: no need for the extra block

packages/components/src/Components/TagModal/TagModal.js
```diff
@@ -1,15 +1,47 @@
...
+        ? rows.length > 0
+        : rows.some(({ id }) => selected && selected.find(({ id: selectedId }) => selectedId === id)) && null
+);
+
+const unique = (arr) => (
+    arr.filter(({ id }, index, arr) => arr.findIndex(({ id: currId }) => currId === id) === index)
+);
 
 export default class TagModal extends React.Component {
+    onSelect = (_event, isSelected, rowId) => {
```
AllenBW, line 37: if not using _event descrust as follows 🤔 ? `onSelect = ({isSelected, rowId}) => {`

packages/components/src/Components/TagModal/TagModal.js
```diff
@@ -57,16 +180,27 @@ TagModal.propTypes = {
...
+    pagination: PropTypes.shape({
+        count: PropTypes.number,
+        page: PropTypes.number,
+        perPage: PropTypes.number
+    }),
+    selected: PropTypes.array
 };
 
 TagModal.defaultProps = {
+    loaded: true,
```
AllenBW, line 195: Not sure, double checking, should this be false?

packages/inventory/src/EntityTableToolbar.js
```diff
@@ -111,77 +105,47 @@ class ContextEntityTableToolbar extends Component {
...
+            allTagsLoaded,
+            allTags,
+            additionalTagsCount > 0 ? [{
+                items: [{
+                    label: `${additionalTagsCount} more tags available`,
+                    onClick: () => toggleTagModal(),
+                    className: 'ins-c-inventory__tags-more-items'
+                }]
+            }] : [],
+            <span> Loading... <Spinner size="md" /> </span>
```
AllenBW, line 136: can we get away with only a spinner here? (sans text)

ryelo: We should be able to remove the text here, the Spinner has hidden text for screen readers

packages/components/src/Components/TagModal/TagModal.js
karelhala: Yeah, that's correct!

PR #441, opened 2020-04-06
Add progress bar to sources wizard

rvsia:
TODO
- [x] tests

loading

### DIFF
--- a/packages/sources/src/addSourceWizard/FinalWizard.js
+++ b/packages/sources/src/addSourceWizard/FinalWizard.js
@@ -16,8 +16,10 @@ const FinalWizard = ({
     successfulMessage,
     hideSourcesButton,
     returnButtonTitle,
-    errorMessage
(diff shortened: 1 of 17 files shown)


rvsia:
@Hyperkid123 That's done by the component.

Hyperkid123:
> @Hyperkid123 That's done by the component.

Can you override it?

rvsia:
I will try

rvsia:
@Hyperkid123 The progress inherits text-align from the empty-state :face_with_thermometer: 

Fixed with `text-align: initial`

loading

(Yeah, this is actually how it should look)

Hyperkid123:
@rvsia i would definitely use text-align: right. This just looks wrong to me :smile:

rvsia:
Yeah, you are right, looks much better

load

(_Maybe it's the right time for maffernfly_)

PR #446, opened 2020-04-08
Lock version of react dom and react to 12 in order to correctly render charts

karelhala:


AllenBW:
for the curious https://github.com/facebook/react/blob/master/CHANGELOG.md#16130-february-26-2020

Hyperkid123:
@karelhala maybe we could try to open an issue in the packages repository. But I was checking it out and the development on that thing seems to be slowing down pretty quickly

PR #447, opened 2020-04-08
Sometimes primary toolbar placeholders need to be capitalized...

AllenBW:
So now, we'll capitalize the labels, but for the filter placeholder to be sentence case, the label will have to be lowercase, BUT this will allow a label of say `Ansible`to remain `Ansible` n make everyone happy

fixes https://projects.engineering.redhat.com/browse/RHCLOUD-5702


AllenBW:
you'll notice snapshot output doesn't change (only the  input values as a result of makin labels lowercase)

terezanovotna:
everyone's happy ✨

PR #450, opened 2020-04-09
When editing authentication do not show empty password

rvsia:
When adding application using existing authentication, users do not have to enter password and the value is not stored in the form state. In these cases, the summary was showing `-`, now it shows `Not edited`

Before
image

After
image

### DIFF
--- a/packages/sources/src/sourceFormRenderer/components/SourceWizardSummary.js
+++ b/packages/sources/src/sourceFormRenderer/components/SourceWizardSummary.js
@@ -32,6 +32,10 @@ export const createItem = (formField, values, stepKeys) => {
         value = value ? 'Yes' : 'No';
     }
 
+    if (!value && formField.name === 'authentication.password' && get(values, 'authentication.id')) {
+        value = '●●●●●●●●●●●●';
+    }
+
     return ({ label: formField['aria-label'] || formField.label,  value: value || '-' });
 };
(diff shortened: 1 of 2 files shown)


Hyperkid123:
@rvsia so if you change your password you will see it in the summary?

rvsia:
@Hyperkid123 Not exactly, If you change the password you will see dots, because you should not be able to see passwords. :D

Hyperkid123:
@rvsia and is the password required field?

rvsia:
@Hyperkid123 when you are using already created authentication, then no.

Hyperkid123:
Hm, I was thinking why not just put the dots here as well. I am not feeling the "Not edited" term.

rvsia:
@Hyperkid123 Maybe better word.. 'Unchanged'? It would be nice to show users that the password is still the same.

Hyperkid123:
I do get that, I am all for it. The "not edited" just does not feel natural.

rvsia:
or we can do both `●●●●●●●●●●●● (unchanged)` ?

Hyperkid123:
I think that the dots are good enough honestly. Everybody knows that password is not going to be visible no matter what. Other fields do not say unchanged as well.

rvsia:
@Hyperkid123 done! :+1:

PR #452, opened 2020-04-11
Add custom form components to be used in DDF

karelhala:
These components will later be used in settings and user preferences. We want to share similiar components between multiple repositories.


karelhala:
ping @Hyperkid123 @rvsia

karelhala:
@Hyperkid123 migration to new version of DDF can happen seperately. I'd prefer releasing these components in older version as well, so we have some fallback. Perform upgrade in all places where we use DDF (user preferences and settings currently) and upgrade these (FEC) components after that.

Hyperkid123:
@karelhala :+1: but you should keep it in mind that the migration should happen sooner rather than later. We still support v1, but we will stop eventually.

Hyperkid123:
@karelhala this does not seem to be relevant anymore. Re-open if you disagree.

PR #453, opened 2020-04-13
Align battery SVG

LI0131:
This is a PR in response to RHCLOUD-5244 (page one of attached google doc). This code moves the battery svg inline with its label. See below.

Before:
Screen Shot 2020-04-13 at 12 20 27 PM

After:
Screen Shot 2020-04-13 at 12 18 56 PM

### DIFF
--- a/packages/components/src/Components/Battery/battery.scss
+++ b/packages/components/src/Components/Battery/battery.scss
@@ -10,6 +10,7 @@ i.ins-battery {
   line-height: 0;
   svg {
     position: relative;
+    top: var(--pf-global--spacer--sm);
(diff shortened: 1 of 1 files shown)


packages/components/src/Components/Battery/battery.scss
```diff
@@ -10,6 +10,7 @@ i.ins-battery {
   line-height: 0;
   svg {
     position: relative;
+    top: 0.5em;
```
karelhala, line 13: Please use global CSS variable `--pf-global--spacer--sm` instead of direct number.

PR #457, opened 2020-04-14
Inventory Advisor cleanups n changes 👨‍🚒🍇🌌

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-5873

#### looks like this
Screen Shot 2020-04-14 at 11 24 11 AM
Screen Shot 2020-04-14 at 11 23 53 AM

### DIFF
--- a/packages/inventory-insights/src/Api.js
+++ b/packages/inventory-insights/src/Api.js
@@ -1,0 +1,25 @@
+import axios from 'axios';
+
+export default {
+    get(url, headers = {}, params = {}) {
+        return axios.get(url, {
+            headers,
+            params
+        });
+    },
+    put(url, data = {}, headers = {}) {
+        return axios.put(url, data, {
+            headers
+        });
+    },
+    post(url, headers = {}, data = {}) {
+        return axios.post(url, data, {
+            headers
+        });
+    },
+    delete(url, data = {}, headers = {}) {
+        return axios.delete(url, data, {
+            headers
+        });
+    }
+};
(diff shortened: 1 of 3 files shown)


packages/inventory-insights/src/Api.js
```diff
@@ -0,0 +1,25 @@
+import axios from 'axios';
+
+export default {
```
AllenBW, line 3: no we don't use all of these, but 🤷 doesn't hurt to have em around

karelhala: You should consider using [@redhat-cloud-services/insights-client](https://www.npmjs.com/package/@redhat-cloud-services/insights-client) at some point. No need to call URLs directly.

packages/inventory-insights/src/Insights.js
```diff
@@ -404,7 +404,7 @@ class InventoryRuleList extends Component {
                     actionsConfig={{ actions }}
                     bulkSelect={bulkSelect}
                     filterConfig={{ items: filterConfigItems }}
-                    pagination={<React.Fragment> {results === 1 ? `${results} recommendation` : `${results} recommendations`} </React.Fragment>}
+                    pagination={<React.Fragment> {results === 1 ? `${results} recommendation` : `${results} Recommendations`} </React.Fragment>}
```
karelhala, line 407: Plural has capital R and singular has lower r, shouldn't this be the same?

AllenBW: yup good call

AllenBW:
OOOO DIDN'T KNOW ABOUT @redhat-cloud-services/insights-client thats hella neat 🤔 not here... maybe again in da... futureeeeeee 🤗

PR #460, opened 2020-04-15
Kba row attribute ensures response of all queried node_id

AllenBW:
oughta fix https://projects.engineering.redhat.com/browse/RHCLOUD-4927

#### we request 23 rows, receive 23 rows
Screen Shot 2020-04-15 at 11 11 26 AM

### DIFF
--- a/packages/inventory-insights/src/Insights.js
+++ b/packages/inventory-insights/src/Insights.js
@@ -95,10 +95,17 @@ class InventoryRuleList extends Component {
 
     fetchKbaDetails = async (reportsData) => {
         const { filters, searchValue, rows } = this.state;
-        const kbaIds = reportsData.map(report => report.rule.node_id).filter(x => x).join(` OR `);
+        const kbaIds = reportsData.map(({ rule }) => rule.node_id).filter(x => x);
         try {
-            const kbaDetailsFetch = (await API.get(`https://access.redhat.com/hydra/rest/search/kcs?q=id:(${kbaIds})&fl=view_uri,id,publishedTitle`,
-                {}, { credentials: 'include' })).data.response.docs;
+            const kbaDetailsFetch = (await API.get(
+                `https://access.redhat.com/hydra/rest/search/kcs?q=id:(${
+                    kbaIds.join(` OR `)
+                })&fl=view_uri,id,publishedTitle&rows=${
+                    kbaIds.length
+                }`,
+                {},
+                { credentials: 'include' }
+            )).data.response.docs;
             this.setState({
                 kbaDetailsData: kbaDetailsFetch,
                 rows: this.buildRows(reportsData, kbaDetailsFetch, filters, rows, true, searchValue)


packages/inventory-insights/src/Insights.js
```diff
@@ -95,10 +95,14 @@ class InventoryRuleList extends Component {
...
+        let nodeIdCount = 0;
+        const kbaIds = reportsData.map(report => {
+            report.rule.node_id && nodeIdCount++;
+            return report.rule.node_id;
+        }).filter(x => x).join(` OR `);
         try {
-            const kbaDetailsFetch = (await API.get(`https://access.redhat.com/hydra/rest/search/kcs?q=id:(${kbaIds})&fl=view_uri,id,publishedTitle`,
-                {}, { credentials: 'include' })).data.response.docs;
+            // eslint-disable-next-line max-len
+            const kbaDetailsFetch = (await API.get(`https://access.redhat.com/hydra/rest/search/kcs?q=id:(${kbaIds})&fl=view_uri,id,publishedTitle&rows=${nodeIdCount}`, {}, { credentials: 'include' })).data.response.docs;
```
karelhala, line 105: This can be styled into multiple rows
```JS
const kbaDetailsFetch = (await API.get(
  `https://access.redhat.com/hydra/rest/search/kcs?q=id:(${
    kbaIds
  })&fl=view_uri,id,publishedTitle&rows=${
    nodeIdCount
  }`,
  {},
  { credentials: 'include' }
 )).data.response.docs;
```

packages/inventory-insights/src/Insights.js
```diff
@@ -95,10 +95,14 @@ class InventoryRuleList extends Component {
 
     fetchKbaDetails = async (reportsData) => {
         const { filters, searchValue, rows } = this.state;
-        const kbaIds = reportsData.map(report => report.rule.node_id).filter(x => x).join(` OR `);
+        let nodeIdCount = 0;
+        const kbaIds = reportsData.map(report => {
+            report.rule.node_id && nodeIdCount++;
```
karelhala, line 100: This will include reports that will be filtered later on is that correct. I think that this can be slightly changes
```JS
const kbaIds = reportsData.map(({ rule }) => rule.node_id).filter(x => x);
const kbaDetailsFetch = (await API.get(
  `https://access.redhat.com/hydra/rest/search/kcs?q=id:(${
    kbaIds.join(` OR `)
  })&fl=view_uri,id,publishedTitle&rows=${
    kbaIds.length
  }`,
  {},
  { credentials: 'include' }
 )).data.response.docs;
```

packages/inventory-insights/src/Insights.js
AllenBW: yipes, thats ah lotta rows, but okkkkk

packages/inventory-insights/src/Insights.js
AllenBW: All we do with the filter there is kill blank entries for a clean join, but dig this, like the idea of no let, thanks for the 👁️ 🤗

AllenBW: works out okie dokie with no ids and 1 id so force push incoming!

PR #461, opened 2020-04-15
Update cost management OpenShift wizard with new operator steps.

chambridge:
* Update OpenShift for cost management with reduced steps with new operator

Source UI - Cost Mgmt Operator

### DIFF
--- a/packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costManagement.js
+++ b/packages/sources/src/addSourceWizard/hardcodedComponents/openshift/costManagement.js
@@ -1,12 +1,9 @@
-import { TextContent, Text, TextVariants, TextListItem, TextList, Button, ClipboardCopy, Popover, ButtonVariant } from '@patternfly/react-core';
+import { TextContent, Text, TextVariants, TextListItem, TextList, Button, Popover, ButtonVariant } from '@patternfly/react-core';
 import { QuestionCircleIcon } from '@patternfly/react-icons';
 import React from 'react';
 import { HCCM_DOCS_PREFIX } from '../../../utilities/stringConstants';
 
-const USAGE_COLLECTOR_URL = 'https://github.com/project-koku/korekuta/archive/master.zip';
 const INSTALL_PREREQUISITE = `${HCCM_DOCS_PREFIX}/html/getting_started_with_cost_management/assembly_adding_sources_cost#installing_ocp_prerequisites`;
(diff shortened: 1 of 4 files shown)


boaz0:
//cc @nlcwong PTAL 
You will probably have to update the mocks

dayleparker:
Looks very nice @chambridge :)

One thing to fix - we mention that you must install the Cost Management Operator in the prerequisites step (screen 3), then again as a step in screen 4. Perhaps we could remove the prereqs step entirely and mention in a different screen OCP 4.3+ is needed? (Or maybe this could go in the cost-mgmt-operator doc?) Other ideas?

Or for now, we could just remove "Cost Management Operator" from the prereqs list until we figure it out.

nlcwong:
I agree, we should remove the prerequisites step and change the wording on the configuration step to the following text:

The Cost Management Operator collects the data required for Cost Management.  This is supported for clusters that are OpenShift Container Platform version 4.3 or later.

The Cost Management Operator from OperatorHub on your cluster
1. Install and configure the Cost Management Operator from OperatorHub on your cluster.
2. When configuration is complete, enter the cluster identifier from the confirmation message below:

abaiken:
Looks good @chambridge - the only thing i would note is that on the last step - the confirmation no longer provides the cluster identifier so we might want to tell them where to find that.

dayleparker:
We have a popover in the old OCP sources flow that said "The cluster identifier value is stored in ~/.config/ocp_usage/config.json". Is that still true? @abaiken @chambridge

chambridge:
@dayleparker The pop over now says:
The cluster identifier can be found in the **Help > About** screen.

dayleparker:
> @dayleparker The pop over now says:
> The cluster identifier can be found in the **Help > About** screen.

Awesome, thank you! I'll add that to the doc.

chambridge:
@nlcwong @dayleparker @abaiken 

Here is the updated version:
Sources-OCP-operator

nlcwong:
@rvsia I noticed in the animated gif, that in the review details section, the cluster identifier's has the required and question mark icon on it.  That doesn't look right, it should just be a straight field label correct?

chambridge:
@nlcwong Its not there in the final version:
image

You can play with it in CI now.

nlcwong:
@chambridge given the comment that @abaiken made about the cluster identifier not being in the confirmation after configuration, can you change the second step in the Configure operator step to say:

2. When configuration is complete, enter the cluster identifier below.  The cluster identifier can be found in the cluster's **Help > About** screen.

That's assuming that we are on the cluster console?

And if you change that text, then you can just get rid of the ? icon and pop over.

PR #464, opened 2020-04-16
Remove stale information from stale component

karelhala:
animation

### DIFF
--- a/packages/components/src/Components/CullingInfo/CullingInformation.js
+++ b/packages/components/src/Components/CullingInfo/CullingInformation.js
@@ -38,15 +38,16 @@ const CullingInformation = ({ culled, className, staleWarning, stale, currDate, 
         }) : children;
     }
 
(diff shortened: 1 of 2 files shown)


packages/components/src/Components/CullingInfo/CullingInformation.js
```diff
@@ -18,35 +18,37 @@ const calculateTooltip = (culled, warning, currDate) => {
...
     return {
         isWarn: true,
-        msg: `This system will be removed from the inventory in ${removeIn} days (${exact(culledDate)})`
+        msg
     };
 };
 
 const CullingInformation = ({ culled, className, staleWarning, stale, currDate, children, render, ...props }) => {
-    if ((new Date(currDate) - new Date(stale)) < 0) {
+    // TODO: remove comments once culling is fine
```
rvsia, line 36: What about to remove the comments and let git history keep it?

karelhala: We'd have to go look trough the history, this is easier.

PR #470, opened 2020-04-20
Apply all filters when fetching all tags

karelhala:
* Allow passing filterBy paramter to Group filter (to show active filter if it changed from outside)
* When fetching inventory data refresh tags as well, this will apply newly selected filters to all tags
* When manipulating with filters clear previous search value for fetching all tags
* When feching all tags apply filters only if no pagination is given, when in modal selection mode you will get all tags regardless on staleness etc.

animation

### DIFF
--- a/packages/components/src/Components/ConditionalFilter/Group.js
+++ b/packages/components/src/Components/ConditionalFilter/Group.js
@@ -23,11 +23,17 @@ class Group extends Component {
         });
     };
 
-    componentDidUpdate({ selected: prevSelected }) {
-        const { selected } = this.props;
+    componentDidUpdate({ selected: prevSelected, filterBy: prevFilterBy }) {
+        const { selected, filterBy } = this.props;
         if (!isEqual(prevSelected, selected)) {
             this.setState({
                 selected
+            });
+        }
+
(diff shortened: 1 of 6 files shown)


packages/components/src/Components/ConditionalFilter/Group.js
```diff
@@ -23,13 +23,19 @@ class Group extends Component {
...
+    componentDidUpdate({ selected: prevSelected, filterBy: prevFilterBy }) {
+        const { selected, filterBy } = this.props;
         if (!isEqual(prevSelected, selected)) {
             this.setState({
                 selected
             });
         }
+
+        if (filterBy !== undefined && prevFilterBy !== filterBy) {
+            this.setState({
```
Hyperkid123, line 35: is it possible to only use the filterBy from props?

karelhala: That would require larger update to this component, currently `filterBy` is stored in state and changes whenever user types something in.

Hyperkid123:
@karelhala just update the snapshot, please.

karelhala:
@Hyperkid123 updated!

PR #473, opened 2020-04-20
Pass hide inv link from parent component

karelhala:


packages/inventory/src/InventoryDetail.js
```diff
@@ -24,10 +24,10 @@ class InventoryDetail extends React.Component {
     }
 
     render() {
-        const { root, match: { params }, useCard, hideBack, actions, showTags } = this.props;
+        const { root, match: { params }, useCard, hideBack, actions, showTags, hideInvLink } = this.props;
         return (
             <React.Fragment>
-                <Entitydetail useCard={ useCard } actions={ actions } showTags={ showTags } />
+                <Entitydetail useCard={ useCard } actions={ actions } showTags={ showTags } hideInvLink={ hideInvLink } />
```
rvsia, line 30: What about to do `{...rest}` ?

karelhala: We don't want send rest props, this is just hot fix to allow hiding inventory app link.

PR #476, opened 2020-04-22
Remove count default to 0 and add isDisabled to BulkSelect

johnsonm325:
If isDisabled is passed in as true, then the dropdown will be disabled.
The default prop value of 0 for count has been removed.


johnsonm325:
How are all of these linting errors showing up?

johnsonm325:
@karelhala tests are passing.

PR #478, opened 2020-04-23
Added markdown templates support to rule-components

bond95:


packages/rule-components/src/ReportDetails/index.js
```diff
@@ -12,26 +12,35 @@ import ReactMarkdown from 'react-markdown';
 import { riskOfChangeMeta, totalRiskMeta } from './constants';
 import LinkInDetails from './LinkInDetails';
 import RiskDescription from './RiskDescription';
+import doT from 'dot';
 
 export { default as style } from './index.scss';
 
+const templateProcessor = (template, definitions) => {
+    const DOT_SETTINGS = { ...doT.templateSettings, varname: [ 'pydata' ], strip: false };
```
karelhala, line 20: No need to assign to a const
```JS
const templateProcessor = (template, definitions) =>(
  definitions ? doT.template(template, {
    ...doT.templateSettings,
    varname: [ 'pydata' ],
    strip: false
})(definitions) : template
);
```

packages/rule-components/src/ReportDetails/index.js
```diff
@@ -12,26 +12,35 @@ import ReactMarkdown from 'react-markdown';
...
         totalRisk,
         riskOfChange,
         showRiskDescription,
+        definitions,
         onFeedbackChanged
-    }) => (
+    }) => {
 
-    <Grid gutter="md" className="ins-c-rule__report-detail">
+    const compiledMd = templateProcessor(details, definitions);
```
karelhala, line 36: No need to assign to a const
```JSX
source={ templateProcessor(details, definitions)}
```

packages/rule-components/src/ReportDetails/index.js
bond95: I like from time to time use constants, to make code more understandable, but ok, we need this changes as soon as possible

PR #480, opened 2020-04-24
RHICOMPL-629 - Pass profile name properly to Remediations

dLobatog:
Before this PR, we were NOT sending properly the profiles to remediations, causing a failure.


packages/inventory-compliance/src/ComplianceRemediationButton.js
```diff
@@ -9,17 +9,13 @@ import { global_BackgroundColor_100 as globalBackgroundColor100 } from '@pattern
 
 class ComplianceRemediationButton extends React.Component {
     formatRule = ({ title, refId }, profile, system) => ({
-        id: `ssg:rhel7|${profile}|${refId}`,
+        id: `ssg:rhel7|${profile.split('xccdf_org.ssgproject.')[1]}|${refId}`,
```
dLobatog, line 12: When this has some garbled up profile refId, it will return undefined - which is fine, because Remediations returns a default profile for that

packages/inventory-compliance/src/ComplianceRemediationButton.js
```diff
@@ -9,17 +9,13 @@ import { global_BackgroundColor_100 as globalBackgroundColor100 } from '@pattern
...
         description: title,
         systems: [
             system
         ]
     })
 
-    findRule = (rules, refId) => {
```
dLobatog, line 19: This is just unused

packages/inventory-compliance/src/ComplianceRemediationButton.js
```diff
@@ -29,9 +25,13 @@ class ComplianceRemediationButton extends React.Component {
...
     }
 
+    ruleProfile = (rule, system) => (
+        system.profiles.find(profile => profile.rules.find(profileRule => rule.refId === profileRule.refId))
+    )
+
     rulesWithRemediations = (rules, system) => {
         return rules.filter(rule => rule.remediationAvailable).map(
-            rule => this.formatRule(rule, system.profiles[0].refId, system.id)
+            rule => this.formatRule(rule, this.ruleProfile(rule, system).refId, system.id)
```
dLobatog, line 34: :warning: Here's one of the keys of the PR. In this commit (https://github.com/RedHatInsights/frontend-components/commit/a5b43f91adb8dbce0c149b02f369425a526d6f9a#diff-2339b617769a4b932bc7ed3b63b7e41dR34) , I changed the old code, which was *always* returning undefined, to a code that aimed to find the profile. However, it was not accurate - as profiles[0] might not be the right one. We should find it in the list of profiles instead

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -330,7 +330,7 @@ class SystemRulesTable extends React.Component {
                     { remediationsEnabled &&
                         <DataToolbarItem>
                             <ComplianceRemediationButton
-                                allSystems={ [{ id: system.id, ruleObjectsFailed: [] }] }
+                                allSystems={ [{ id: system.id, profiles: system.profiles, ruleObjectsFailed: [] }] }
```
dLobatog, line 333: :warning:  And here's the other key of the PR - we need to pass the profiles to the ComplianceRemediationButton to find it and pass it

packages/inventory-compliance/src/PresentationalComponents/ConditionalLink.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 
 const ConditionalLink = ({ children, href, ...additionalProps }) => (
-    href && <a href={ href } { ...additionalProps }>{ children }</a> || children
+    href && <a href={ href } { ...additionalProps }>{ children }</a> || children || ''
```
dLobatog, line 4: I noticed that if you have some references/identifiers in the compliance db that are returning nil, this component isn't able to render - and causes the whole page to fail. This is just a side-note and should hardly be called (it just returns an empty string if needed)

bastilian:
Looks good to me! It would be good to add fixtures or updated the test data to also return a profile in the snapshot to verify it does and keeps doing the right thing. 

Another thing that would be nice to have woudl be a prop validation for `allSystems` with `propType.arrayOf`. However, this will require to provide a custom validation, so for now it would be nice if there was a comment next to the prop that describes what a "system" has to provide.

PR #481, opened 2020-04-27
Left Align Remediations Wizard

LI0131:
@AllenBW This PR is in response to RHCLOUD-5247 (page 2 of the attached doc). This change makes the footer of the Wizard appear on the left-hand side. This _should_ be the only change that is necessary for this to function, however, I haven't been able to correctly setup a local chrome env in order to throughly test. The wizard is loaded through the `chrome.experimental` window var.

### DIFF
--- a/packages/components/src/Components/Wizard/Wizard.js
+++ b/packages/components/src/Components/Wizard/Wizard.js
@@ -50,12 +50,6 @@ class Wizard extends Component {
         );
 
         const renderModalActions =  [
-            <Button key="cancel" action="cancel" variant="secondary" onClick={ () => this.handleOnClose(false) }>
-            Cancel
-            </Button>,
-            // Conditionally render 'previous' button if not on first page
-            this.state.currentStep !== 0 &&
-                <Button key="back" action="back" variant="secondary" onClick={ this.handlePreviousModalStep }> Back </Button>,
             // Conditionally render 'confirm' button if on last page
             this.state.currentStep < this.props.content.length - 1
                 ? <Button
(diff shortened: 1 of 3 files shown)


LI0131:
So like I said in my previous comment... I've been struggling to setup a local chrome env to test this code throughly 😞 but I think I have a logically equivalent test which proves the functionality.

I am using the following function to test.
```
function fakeLoader () {
    //eslint-ignore-next-line
    const loaded = getLoader()({
        react: React,
        reactCore,
        pfReactTable
    });
    return Promise.all([
        import('../index.js'),
        import('../index.css')
    ]).then(([remediations]) => remediations);
}
```
This loads the dependencies using the loadRemediations function, and then imports `@redhat-cloud-services/frontend-components-remediations` and `@redhat-cloud-services/frontend-components-remediations/index.css` (the functionality of the loadRemediations chrome function) from the local dir.

Here is the output from before and after.

Before:
Screen Shot 2020-04-29 at 10 03 55 AM

After:
Screen Shot 2020-04-29 at 10 03 26 AM

karelhala:
@LI0131 yeah, running local chrome is hard. If you want ping me on slack and I'll help you. This change request comes from @terezanovotna? @jharting are you awere of it? I know it's just a button shift, just wanted to overcommunicate instead of showing this to users without rest of the folks aware of this change.

LI0131:
@terezanovotna the buttons should now appear in the opposite order 🥳 

Screen Shot 2020-05-01 at 1 24 15 PM

Screen Shot 2020-05-01 at 1 24 24 PM

PR #483, opened 2020-04-27
Add Tooltip to Inventory-Insights added column

LI0131:
This PR is in response to RHCLOUD 5247 (page 2 of the attached doc). It adds a tooltip to the Inventory table over the `added` column showing the time in raw UTC.

Before:
Screen Shot 2020-04-27 at 1 53 38 PM

After:
Screen Shot 2020-04-27 at 1 54 56 PM

### DIFF
--- a/packages/inventory-insights/src/Insights.js
+++ b/packages/inventory-insights/src/Insights.js
@@ -19,6 +19,7 @@ import ChartSpikeIcon from '@patternfly/react-icons/dist/js/icons/chartSpike-ico
 import CheckCircleIcon from '@patternfly/react-icons/dist/js/icons/check-circle-icon';
 import CheckIcon from '@patternfly/react-icons/dist/js/icons/check-icon';
 import { ClipboardCopy } from '@patternfly/react-core/dist/js/components/ClipboardCopy/ClipboardCopy';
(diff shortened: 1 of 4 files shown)


terezanovotna:
looks good!

packages/inventory-insights/src/Insights.js
```diff
@@ -163,7 +164,11 @@ class InventoryRuleList extends Component {
                         { title: <div> {rule.description}</div> },
                         {
                             title: <div key={key}>
-                                {moment(rule.publish_date).fromNow()}
+                                <Tooltip key={key} position={TooltipPosition.bottom} content={
```
karelhala, line 167: Please do not use moment.js as it bring a ton of unnecessary code. You don't really need it since `DateFormat` does everything for you
```JSX
title: <div key={key}>
  <DateFormat date={rule.publish_date} />
</div>
```

LI0131:
@karelhala @terezanovotna currently `DateFormat` provides a ToolTip by default for `type=relative` so I added the ability to pass custom props to ToolTip. The functionality should be the same as in my original comment.

karelhala:
@LI0131 looking good! Thank you!

PR #489, opened 2020-04-29
Fix async validator - now it has to throw errors

rvsia:
Fix asyncValidator according to changes in v2


packages/sources/src/addSourceWizard/SourceAddSchema.js
```diff
@@ -10,11 +10,11 @@ import ValidatorReset from './ValidatorReset';
 
 export const asyncValidator = (value, sourceId = undefined) => findSource(value).then(({ data: { sources } }) => {
     if (sources.find(({ id }) => id !== sourceId)) {
-        return 'Name has already been taken';
+        throw 'Name has already been taken';
     }
 
     if (value === '' || value === undefined) {
-        return 'Required';
+        throw 'Required';
```
Hyperkid123, line 17: You can now combine async validator with others. So we might want to use the required validator instead in the schema.

rvsia: fixed!

PR #490, opened 2020-04-29
Update steps styling according to mockups

rvsia:
Icon color should be #3E8635
32px padding between icon and title
16px padding between title and text
32px padding between text and button

~~Entire empty state should be centered (vertically and horizontally) within the whitespace~~ skipped because that's not how PF works


rvsia:
@Hyperkid123 don't ask me :D

PR #491, opened 2020-04-29
Display the request-id in the error window

fhlavac:
- Added request id (header "x-rh-insights-request-id") displaying in error window for every call coming.
- Updated tests

@karelhala

### DIFF
--- a/packages/utils/src/interceptors.js
+++ b/packages/utils/src/interceptors.js
@@ -42,8 +42,10 @@ export function interceptor500(error) {
 
 export function errorInterceptor(err) {
     if (!axios.isCancel(err)) {
+        let requestId;
         try {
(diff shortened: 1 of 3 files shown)


packages/utils/src/interceptors.js
```diff
@@ -53,7 +55,7 @@ export function errorInterceptor(err) {
         catch (customError) {
             const sentryId = Sentry.captureException(customError);
             customError.sentryId = sentryId;
-            throw customError;
+            throw { ...customError, requestId };
```
karelhala, line 58: This is not correct, spreading is not going to work on this example
```JS
customError.sentryId = sentryId;
customError.requestId = requestId;
throw customError;
```

PR #495, opened 2020-04-30
Fixes the case of the inv-adv missing ul list stying 🕵🏿‍♀️

AllenBW:
cuz pf-c-list is no longer available on chrome


AllenBW:
@karelhala chrome overrides default styles for some html elements, in the past this was fixable by just applying the pf class, but after some global pf class cleanup looks like its no longer available... no component here, this is for the rule render engine

PR #496, opened 2020-04-30
Added Remediating modal and changed markdown rendered engine

bond95:
image
image
image


packages/rule-components/src/ReportDetails/RemediatingModal.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Modal, Button, Tabs, Tab } from '@patternfly/react-core';
+import Markdown from './Markdown';
+
+class RemediatingModal extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
```
karelhala, line 9: You can write the state directly to the class.
```
class RemediatingModal extends React.Component {
  state = {
    isModalOpen: false,
    activeTabKey: 0
  }
}
```

packages/rule-components/src/ReportDetails/RemediatingModal.js
```diff
@@ -0,0 +1,71 @@
...
+          activeTabKey: tabIndex
+      });
+  };
+
+  render() {
+      const { reason, resolution, definitions } = this.props;
+      const { isModalOpen, activeTabKey } = this.state;
+
+      return (
+      <>
```
karelhala, line 32: I don't really like this usage, it is super easy to overlook it. Can you add `Fragment` instead? You can import it with `Component` as well so you don't have to write `React.Fragment`

packages/rule-components/src/ReportDetails/Markdown.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import propTypes from 'prop-types';
+
+import doT from 'dot';
+import marked from 'marked';
+import sanitizeHtml from 'sanitize-html';
+
+marked.Renderer.prototype.link = (href, title, text) => {
```
karelhala, line 8: Nice usage of prototype.

packages/rule-components/src/ReportDetails/Markdown.js
```diff
@@ -0,0 +1,58 @@
...
+    const externalLinkIcon = '<i class="fas fa-external-link-alt"></i>';
+
+    try {
+        const compiledDot = definitions ? doT.template(template, DOT_SETTINGS)(definitions) : template;
+        const compiledMd = marked(sanitizeHtml(compiledDot, sanitizeOptions));
+
+        return <div dangerouslySetInnerHTML={{
+            __html: compiledMd
+            .replace(/<ul>/gim, `<ul class="pf-c-list" style="font-size: inherit">`)
+            .replace(/<a>/gim, `<a rel="noopener noreferrer" target="_blank">`)
```
karelhala, line 41: There's no href in here. Is that intentional?

doruskova:
@bond95 The remediation modal looks good and can you fix the spacing between links (How to remediate and KCS?)
Snímek obrazovky 2020-05-04 v 8 49 31

Btw. good work!

packages/rule-components/src/ReportDetails/Markdown.js
bond95: I copied it from `inventory-insights`, so I think it should be like this 😃

bond95:
@doruskova unfortunately, I cannot change order, because KB link is part of markdown, so I can add remediation link only after it.

doruskova:
@bond95 And is it possible to have a gap between the description and KCS with Remediation? 
Like this? 
Snímek obrazovky 2020-05-04 v 12 50 21

bond95:
@doruskova ?
image

bond95:
@karelhala updated, please, take a look

PR #504, opened 2020-05-05
Pass isChecked in order to properly calculate 3rd state of checkbox

karelhala:


mmenestr:
Do you mind adding a screenshot of it :D

packages/components/src/Components/BulkSelect/BulkSelect.js
```diff
@@ -49,6 +49,7 @@ visible unless you update it.');
                                                 aria-label="Select all"
                                                 onChange={ onSelect }
                                                 checked={ checked }
```
PanSpagetka, line 51: Do we need both `checked` and `isChecked` ?

karelhala:
@mmenestr sure thing!

Screenshot from 2020-05-06 12-04-00

packages/components/src/Components/BulkSelect/BulkSelect.js
karelhala: Yeah, I wanted to keep the checked for consistency reasons, but it works same way without it.

mmenestr:
Looks good to me! Thanks, Karel!

PR #510, opened 2020-05-08
Add NotAuthorized component to handle 403s

akofink:
Stolen (and modified a bit) directly from https://github.com/RedHatInsights/insights-rbac-ui/blob/master/src/presentational-components/states/DeniedState.js

### DIFF
--- a/packages/components/src/Components/NotAuthorized/NotAuthorized.js
+++ b/packages/components/src/Components/NotAuthorized/NotAuthorized.js
@@ -1,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import {
+    Title,
+    Button,
+    EmptyState,
+    EmptyStateVariant,
+    EmptyStateIcon,
+    EmptyStateBody
+} from '@patternfly/react-core';
+
+import { LockIcon } from '@patternfly/react-icons';
+
+import './NotAuthorized.scss';
+
+const NotAuthorized = ({ serviceName }) => {
+    return (
+        <React.Fragment>
+            <EmptyState variant={ EmptyStateVariant.full } className='ins-c-not-authorized'>
+                <EmptyStateIcon icon={ LockIcon } />
+                <Title headingLevel="h5" size="lg"> You do not have access to { serviceName }</Title>
+                <EmptyStateBody>
+                    Contact your organization administrator(s) for more information.
+                </EmptyStateBody>
+                {
+                    document.referrer ?
+                        <Button variant="primary" onClick={ () => history.back() }>Return to previous page</Button> :
+                        <Button variant="primary" component="a" href=".">Go to landing page</Button>
+                }
+            </EmptyState>
+        </React.Fragment>
+    );
+};
+
+NotAuthorized.propTypes = {
+    serviceName: PropTypes.node
+};
+
+export default NotAuthorized;
--- a/packages/components/src/Components/NotAuthorized/NotAuthorized.scss
+++ b/packages/components/src/Components/NotAuthorized/NotAuthorized.scss
@@ -1,0 +1,7 @@
+.ins-c-not-authorized {
+    .pf-c-title {
+        max-width: 540px;
+        margin-left: auto;
+        margin-right: auto;
+    }
+}
--- a/packages/components/src/Components/NotAuthorized/NotAuthorized.test.js
+++ b/packages/components/src/Components/NotAuthorized/NotAuthorized.test.js
@@ -1,0 +1,11 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import toJson from 'enzyme-to-json';
+import NotAuthorized from './NotAuthorized';
+
+describe('NotAuthorized component', () => {
+    it('should render', () => {
+        const wrapper = shallow(<NotAuthorized serviceName='Foo' />);
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+});
--- a/packages/components/src/Components/NotAuthorized/__snapshots__/NotAuthorized.test.js.snap
+++ b/packages/components/src/Components/NotAuthorized/__snapshots__/NotAuthorized.test.js.snap
@@ -1,0 +1,31 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`NotAuthorized component should render 1`] = `
+<Fragment>
+  <EmptyState
+    className="ins-c-not-authorized"
+    variant="full"
+  >
+    <EmptyStateIcon
+      icon={[Function]}
+    />
+    <Title
+      headingLevel="h5"
+      size="lg"
+    >
+       You do not have access to 
+      Foo
+    </Title>
+    <EmptyStateBody>
+      Contact your organization administrator(s) for more information.
+    </EmptyStateBody>
+    <Component
+      component="a"
+      href="."
+      variant="primary"
+    >
+      Go to landing page
+    </Component>
+  </EmptyState>
+</Fragment>
+`;
--- a/packages/components/src/Components/NotAuthorized/index.js
+++ b/packages/components/src/Components/NotAuthorized/index.js
@@ -1,0 +1,2 @@
+export { default as NotAuthorized } from './NotAuthorized';
+import './NotAuthorized.scss';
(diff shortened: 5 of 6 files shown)


packages/components/src/Components/NotAuthorized/NotAuthorized.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon403 from './icon-403';
+import {
+    Title,
+    Button
+} from '@patternfly/react-core';
+import { withRouter } from 'react-router-dom';
```
dLobatog, line 8: I think that you could do this without props with `useHistory` (https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/hooks.md#usehistory) - you'd need to update react-router-dom to 5.x, I don't know if that's OK though.

dLobatog: Is react-router-dom needed at all here actually? You can call `history.back()` directly from the browser without importing anything

bastilian:
@akofink Looks good to me! When linking (not npm linking) `react-router-dom` in compliance to the package under node_modules in frontend-components the `NotAuthorized` component renders normally in the compliance app and does not warn about `You should not use <Route> or withRouter() outside a <Router>`. It should also render normaly when the `frontend-components` package is release with this component and normally installed via `npm`.

I'm not sure why that is. Maybe someone else in @RedHatInsights/ui-reviewers has seen this issue before and can shed some light on this.

bastilian:
Ps.: I did try to upgrade `react-router-dom` packages to the latest and the same version in all related packages, but that didn't do the trick. It had to be exacatly the same package. aparently. 🤷

packages/components/src/Components/NotAuthorized/Icon403.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
```
karelhala, line 1: Please don't use this component, it is already exported by PF for us
```JSX
import { LockIcon } from '@patternfly/react-icons`;
```

you can check the full list of icons at https://patternfly-react.surge.sh/icons/

packages/components/src/Components/NotAuthorized/NotAuthorized.js
```diff
@@ -0,0 +1,42 @@
...
+const NotAuthorized = ({ serviceName, ...props }) => {
+    let history = useHistory();
+    let button;
+
+    if(history.length) {
+        button = <Button onClick={ history.goBack }>
+            Return to previous page
+        </Button>;
+    } else {
+        button = <Button onClick={ history.push('/') }>
```
karelhala, line 19: This will operate only at app level. Meaning if I go from `insights/compliance/systems` to `insights/compliance/systems/UUID` I can go back to `insights/compliance/systems` if you'll use `history.push('/')` it'll redirect you to `insights/compliance/`.

We've had a consensus with UX folks to redirect users to root URL `window.location.href = './'`

packages/components/src/Components/NotAuthorized/NotAuthorized.js
```diff
@@ -0,0 +1,42 @@
...
+import PropTypes from 'prop-types';
+import Icon403 from './Icon403';
+import {
+    Title,
+    Button
+} from '@patternfly/react-core';
+import { useHistory } from 'react-router-dom';
+
+const NotAuthorized = ({ serviceName, ...props }) => {
+    let history = useHistory();
```
karelhala, line 11: I don't know how to behave with hooks, we should check if every app is using newer version of React and if they are using it we should update the `react-router-dom` and `react` to newer version. But I'd rather use `withRouter` in here so we don't have any problems.

packages/components/src/Components/NotAuthorized/NotAuthorized.js
```diff
@@ -0,0 +1,42 @@
...
+            <Icon403/>
+            <Title size='3xl'>You do not have access to { serviceName }</Title>
+            <Title headingLevel="h2" size='xl'>Contact your organization administrator(s) for more information.</Title>
+            { button }
+        </section>
+    );
+};
+
+NotAuthorized.propTypes = {
+    serviceName: PropTypes.string
```
karelhala, line 35: I'd use `PropTypes.node` in here.

packages/components/src/Components/NotAuthorized/NotAuthorized.js
```diff
@@ -0,0 +1,42 @@
...
+};
+
+NotAuthorized.defaultProps = {
+    serviceName: undefined
```
karelhala, line 39: No need to add defaultProps to `undefined` it is by default `undefined` if you want to have a value in here you could use empty string or null for instance.

packages/components/src/Components/NotAuthorized/notAuthorized.scss
```diff
@@ -0,0 +1,19 @@
...
+.ins-c-component_not-authorized {
+    display: flex;
+    justify-content: center;
+    flex-direction: column;
+    align-items: center;
+    text-align: center;
+
+    .ins-c-text__sorry { max-width: 70%; }
+
+    h1, svg, h2 { @include rem('margin-bottom', 30px); }
```
karelhala, line 12: Either use `var(--pf-global--spacer--lg)` or `var(--pf-global--spacer--xl)` depends on what fits it better.

akofink:
Thanks for the review, @karelhala! I think I've addressed all your comments. How do I get around this error? I can't even see the page locally to make sure it matches the mocks.

`Error: Invariant failed: You should not use <withRouter(NotAuthorized) /> outside a <Router>`

packages/components/src/Components/NotAuthorized/NotAuthorized.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { LockIcon } from '@patternfly/react-icons';
+import {
+    Title,
+    Button
+} from '@patternfly/react-core';
+import { withRouter } from "react-router";
```
karelhala, line 8: Single quotes should be used here.

karelhala:
@akofink I tried to run test locally and it is working fine, no error thrown. I'll try to include it in app as well and see the outcome.

karelhala:
@akofink this will be problem with different context since newer versions of react-router-dom started to use context it won't connect properly to it. The best way here will be by adding `history` prop that is passed from parent component. If no `history` is passed redirect users to root. This way you don't have to use `withRouter` or `useHistory` in this component, parent component can import history with `useHistory` and pass it down to this component, sounds much cleaner to be honest.

akofink:
I've updated this PR to match the DeniedState from the rbac-ui https://github.com/RedHatInsights/insights-rbac-ui/blob/master/src/presentational-components/states/DeniedState.js. It's a shame they didn't add it here first, since it's needed for all apps.

packages/components/src/Components/NotAuthorized/NotAuthorized.js
```diff
@@ -0,0 +1,41 @@
...
+    <React.Fragment>
+      <EmptyState variant={ EmptyStateVariant.full } className='ins-c-not-authorized'>
+        <EmptyStateIcon icon={ LockIcon } />
+        <Title headingLevel="h5" size="lg"> You do not have access to { serviceName }</Title>
+        <EmptyStateBody>
+          Contact your organization administrator(s) for more information.
+        </EmptyStateBody>
+        {
+          document.referrer ?
+            <Button variant="primary" onClick={ () => history.back() }>Return to previous page</Button> :
```
dLobatog, line 29: Just a nitpick, I think what @karelhala mentioned before was that it'd be good to have a `history` prop you could pass to this component to go back in history within the app without having to reload

akofink: @dLobatog This is what you originally requested (using the browser's `history`), and that's what rbac-ui does. It seems to work pretty well - try to navigate to edit user roles in CI with your account, and you'll see this work as written here in the rbac-ui. https://github.com/RedHatInsights/insights-rbac-ui/blob/master/src/presentational-components/states/DeniedState.js#L32 No React required! What do you think?

PR #514, opened 2020-05-15
Added action feedback to RuleFeedback

bond95:
image
image
image

### DIFF
--- a/packages/rule-components/src/ReportDetails/RuleFeedback.js
+++ b/packages/rule-components/src/ReportDetails/RuleFeedback.js
@@ -1,36 +1,59 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import ThumbsUpIcon from '@patternfly/react-icons/dist/js/icons/outlined-thumbs-up-icon';
-import ThumbsDownIcon from '@patternfly/react-icons/dist/js/icons/outlined-thumbs-down-icon';
+import ThumbsUpIcon from '@patternfly/react-icons/dist/js/icons/thumbs-up-icon';
+import OutlinedThumbsUpIcon from '@patternfly/react-icons/dist/js/icons/outlined-thumbs-up-icon';
+import ThumbsDownIcon from '@patternfly/react-icons/dist/js/icons/thumbs-down-icon';
+import OutlinedThumbsDownIcon from '@patternfly/react-icons/dist/js/icons/outlined-thumbs-down-icon';
 import { Button } from '@patternfly/react-core';
 
+export const feedback = { negative: -1, neutral: 0, positive: 1 };
 // ruleId - is id of current rule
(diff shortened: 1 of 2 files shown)


packages/rule-components/src/ReportDetails/RuleFeedback.js
```diff
@@ -1,36 +1,58 @@
...
+                variant="plain"
+                aria-label="Rule is helpful"
+                onClick={ () => {
+                    this.setState({ feedbackSaved: true });
+                    onFeedbackChanged(ruleId, 1);
+                }}
+            >
+                {
+                    userVote === 1
+                        ? <ThumbsUpIcon color="#3C6C37"/>
```
karelhala, line 29: Please don't use direct colors, if we'll change color pallete in future this will be hard to fix. You should be good with `-pf-global--palette--green-500`

packages/rule-components/src/ReportDetails/RuleFeedback.js
```diff
@@ -1,36 +1,58 @@
...
+                variant="plain"
+                aria-label="Rule is not helpful"
+                onClick={ () => {
+                    this.setState({ feedbackSaved: true });
+                    onFeedbackChanged(ruleId, -1);
+                }}
+            >
+                {
+                    userVote === -1
+                        ? <ThumbsDownIcon color="#0066CC"/>
```
karelhala, line 44: Same as above, here you should use `--pf-global--palette--light-blue-500` but that's Marie's call on this.

packages/rule-components/src/ReportDetails/index.js
```diff
@@ -110,6 +115,7 @@ ReportDetails.propTypes = {
     ruleId: PropTypes.string,
     totalRisk: PropTypes.number,
     riskOfChange: PropTypes.number,
+    userVote: PropTypes.oneOf([ -1, 0, 1 ]),
```
karelhala, line 118: I guess that this comes from server, right? Can you make it into constants?

doruskova:
@bond95 looks good to me.

packages/rule-components/src/ReportDetails/RuleFeedback.js
bond95: I think we should ask @doruskova if she is fine with that color.

So you want to move color to CSS?

doruskova: @bond95 it's var(--pf-global--success-color--200)

packages/rule-components/src/ReportDetails/RuleFeedback.js
doruskova: @karelhala @bond95 it's var(--pf-global--primary-color--100)

bond95: Thank you

PR #522, opened 2020-05-26
Add rollup to enable esm and cjs builds

karelhala:
* [x] notifications
* [x] components
* [x] utils
* [x] charts
* [x] inventory-compliance
* [x] inventory-insights
* [x] inventory
* [x] inventory-general-info
* [x] remediations
* [x] sources
* [x] translations


karelhala:
This PR is ready to review! Once we merge this, we'll switch master to v2 branch and release v2 of each package.

README.md
```diff
@@ -17,6 +17,80 @@ Monorepo of Red Hat Cloud services Components for applications in a React.js env
...
+## Treeshaking PF with babel plugin
+
+Patternfly packages require some ehancements to be done in order to properly treeshake your bundles. You can either use direct imports or plugin that does that for you, there are actually 2 plugins to do this
+* [babel-plugin-import](https://www.npmjs.com/package/babel-plugin-import) - easy setup, however not that extensible
+* [transform-imports](https://www.npmjs.com/package/transform-imports) - harder to setup, but allows custom rules
+
+Since Patternfly requires a bit of custom settings you should use `transform-imports`. Change your babel to be JS file `babel.config.js` and add these changes to it
+
+```JS
+require.extensions['.css'] = () => undefined;
```
Hyperkid123, line 29: I would add a note that is required only in the case of PF. And also that `require.extensions` are deprecated in the latest versions of node. We need to find a different solution fo the future.

packages/charts/package.json
```diff
@@ -3,12 +3,13 @@
     "version": "1.0.0",
     "description": "Charts for RedHat Cloud Services project.",
     "main": "dist/index.js",
```
Hyperkid123, line 5: You should update this to point at `/dist/cjs`

packages/components/config/rollup.config.js
```diff
@@ -0,0 +1,112 @@
...
+import babel from 'rollup-plugin-babel';
+import nodeGlobals from 'rollup-plugin-node-globals';
+import { terser } from 'rollup-plugin-terser';
+import postcss from 'rollup-plugin-postcss';
+import json from '@rollup/plugin-json';
+import { dependencies, peerDependencies, name } from '../package.json';
+import { createFilter } from '@rollup/pluginutils';
+import glob from 'glob';
+
+const globMapper = (mapper) => glob.sync(mapper).reduce((acc, item) => {
```
Hyperkid123, line 13: I think that there can be some common place for this script.

packages/components/config/rollup.config.js
```diff
@@ -0,0 +1,112 @@
...
+    const [ path ] = item.split('/index.js');
+    const last = path.split('/').pop();
+
+    return {
+        ...acc,
+        [last === 'src' ? 'index' : last]: item
+    };
+}, {});
+
+const externalDeps = Object.keys({ ...dependencies, ...peerDependencies }).map(item =>
```
Hyperkid123, line 23: Same here, this pattern is probably going to be exactly the same for the rest of the packages

packages/components/src/index.js
```diff
@@ -29,7 +29,7 @@ export * from './Components/BulkSelect';
 export * from './Components/ConditionalFilter';
 export * from './Components/TagCount';
 export * from './Components/TagModal';
-export * from './Components/PrimaryToolbar';
+// export * from './Components/PrimaryToolbar';
```
Hyperkid123, line 32: :question:

packages/config/src/config.js
```diff
@@ -16,12 +17,12 @@ module.exports = ({
         optimization: {
             minimize: (process.env.NODE_ENV || mode) === 'production',
             splitChunks: {
+                chunks: 'all',
+                maxInitialRequests: Infinity,
```
Hyperkid123, line 21: I think that there should be some upper limit for the entry point requests. Usually is set around 20-40

packages/inventory-compliance/package.json
```diff
@@ -3,12 +3,13 @@
     "version": "1.0.12",
     "description": "Detail part of compliance for RedHat Cloud Services.",
     "main": "index.js",
```
Hyperkid123, line 5: again missing the new path

packages/inventory-general-info/package.json
```diff
@@ -3,12 +3,13 @@
     "version": "1.0.5",
     "description": "System info detail for RedHat Cloud Services project.",
     "main": "index.js",
```
Hyperkid123, line 5: ^^

Hyperkid123: Oh now it hit me. We might even need the "files" property of you want to publish more that cjs/index and esm/index. That would be probably for the umd build which is not listed anywhere and might get ignored.

packages/inventory-insights/package.json
```diff
@@ -3,12 +3,13 @@
     "version": "1.0.17",
     "description": "Rules detail page for RedHat Cloud Services project.",
     "main": "index.js",
```
Hyperkid123, line 5: ^^

packages/inventory/package.json
```diff
@@ -1,14 +1,15 @@
 {
     "name": "@redhat-cloud-services/frontend-components-inventory",
-    "version": "1.0.29",
+    "version": "2.0.0-beta2",
     "description": "Inventory components for RedHat Cloud Services project.",
     "main": "index.js",
```
Hyperkid123, line 5: ^^

README.md
karelhala: Good point, I'll add it there.

packages/charts/package.json
karelhala: Not really, this deliberately points to UMD since it's the fallback way, less breaking changes. If folks want to use `cjs` they can do that by importing said file.

packages/components/config/rollup.config.js
karelhala: Mhh, yeah. I'll write some common config, since a lot of these things just repeat.

packages/components/src/index.js
karelhala: Whoops, forgot to re-add it when testing some comments.

packages/inventory-general-info/package.json
karelhala: The `files` field is just alternative to `.npmignore` https://docs.npmjs.com/files/package.json#files they work in different way

> Files included with the “package.json#files” field cannot be excluded through .npmignore or .gitignore.

packages/charts/package.json
Hyperkid123: Ok then we are definitely missing the files attribute to include the cjs files..

Hyperkid123: I have noticed that you have changed the main in one of the packages though. Was that deliberate or only for testing purposes?

Hyperkid123: Also, if we are releasing v2, I would be in favor using the cjs over UMD anyway. Current users will still be able to use UMD but we will point them in the right direction for the future.

karelhala: Yeah, that was deliberate, because it was poiting to wrong file.

packages/translations/config/rollup.config.js
```diff
@@ -76,29 +54,41 @@ const plugins = [
...
+        [ '@patternfly', '@redhat-cloud-services', 'react-intl' ]
+    )),
+    plugins,
+    globals,
+    name,
+    [ entryMapper ],
+    './'
+);
+
+// export default [
```
Hyperkid123, line 69: Sneaky comment

packages/config/src/config.js
karelhala: I went with example of how to do it from this [blog](https://medium.com/hackernoon/the-100-correct-way-to-split-your-chunks-with-webpack-f8a9df5b7758). I might add a few more `cacheGroups` though. So we have at least PF, FEC and react dependencies in seperate files.

karelhala:
I've applied the config changes directly to rbac, this is the outcome (total filesize remained the same).

## Before
Screenshot from 2020-06-01 14-00-16

## After
Screenshot from 2020-06-01 14-00-40

Hyperkid123:
Now we can go through the whole internet and fix the lodash usage :smile:

PR #525, opened 2020-05-29
Big sources refactoring

rvsia:
todo:
- [x] remove class components
  - [x] add source button
  - [x] add source wizard
  - [x] add source modal
  - [x] cardSelect
- [x] export steps to consume them in sources UI
- [x] fix rejected promises in handleError test


Hyperkid123:
@rvsia make sure this will end up in the v2 branch aswell.

PR #527, opened 2020-06-01
Enhance collection information in Inventory detail

fhlavac:
Enhanced "Collection information" in Inventory detail
* added Insights id
* added Reporter
* added Rhel machine id

* updated snapshots

@karelhala

### DIFF
--- a/packages/inventory-general-info/src/__snapshots__/GeneralInformation.test.js.snap
+++ b/packages/inventory-general-info/src/__snapshots__/GeneralInformation.test.js.snap
@@ -739,6 +739,54 @@ exports[`GeneralInformation should render correctly - no data 1`] = `
                  
               </div>
             </dd>
+            <dt
+              class=""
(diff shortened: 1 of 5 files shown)


packages/inventory-general-info/src/__snapshots__/GeneralInformation.test.js.snap
```diff
@@ -1136,7 +1184,7 @@ exports[`GeneralInformation should render correctly 1`] = `
               class=""
               data-pf-content="true"
             >
-              13 Aug 2019
+              12 Aug 2019
```
karelhala, line 1187: This should actually be `13 Aug 2019` because of timezones and such.

fhlavac: oh, I missed that one, thanks!

PR #529, opened 2020-06-02
RHICOMPL-646 - Sync empty states in Reports/SCAP policies

dLobatog:
The empty state should also give some info to the user about existing
policies, as the current message is a bit confusing. This shows how many
policies exist (but are not reporting) if any - if there are no policies, it just shows the same message as before.

Screenshot from 2020-06-02 09-58-31
Screenshot from 2020-06-02 09-57-31
Screenshot from 2020-06-02 09-51-29
Screenshot from 2020-06-02 09-51-19
Screenshot from 2020-06-02 09-51-09

### DIFF
--- a/packages/inventory-compliance/package.json
+++ b/packages/inventory-compliance/package.json
@@ -40,6 +40,7 @@
         "@redhat-cloud-services/frontend-components-notifications": "*",
         "@redhat-cloud-services/frontend-components-remediations": "*",
         "@redhat-cloud-services/frontend-components-utilities": "*",
+        "@apollo/react-hooks": "^3.1.5",
         "apollo-boost": "^0.1.23",
         "graphql-tag": "^2.10.0",
         "react-apollo": "^2.3.3",
--- a/packages/inventory-compliance/src/Compliance.js
+++ b/packages/inventory-compliance/src/Compliance.js
@@ -8,13 +8,8 @@ import gql from 'graphql-tag';
 import { ApolloProvider } from 'react-apollo';
 import { ApolloClient, HttpLink, InMemoryCache } from 'apollo-boost';
 import { columns } from './defaultColumns';
-import {
-    Card,
-    CardBody,
-    CardHeader
-} from '@patternfly/react-core';
-import { NotEqualIcon } from '@patternfly/react-icons';
 import './compliance.scss';
+import { ErrorCard } from './PresentationalComponents';
 
 const COMPLIANCE_API_ROOT = '/api/compliance';
 
@@ -79,15 +74,8 @@ class SystemDetails extends Component {
     renderError = (error) => {
         const errorMsg = `Oops! Error loading System data: ${error}`;
         return (error.networkError && error.networkError.statusCode === 404) ?
-            <ComplianceEmptyState title='No policies are reporting' /> :
-            <Card className="ins-error-card">
-                <CardHeader>
-                    <NotEqualIcon />
-                </CardHeader>
-                <CardBody>
-                    <div>{ errorMsg }</div>
-                </CardBody>
-            </Card>;
+            <ComplianceEmptyState title='No policies are reporting for this system' /> :
+            <ErrorCard message={errorMsg} />;
     }
 
     render() {
--- a/packages/inventory-compliance/src/ComplianceEmptyState.js
+++ b/packages/inventory-compliance/src/ComplianceEmptyState.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import propTypes from 'prop-types';
+import { ApolloClient, HttpLink, InMemoryCache } from 'apollo-boost';
 import {
     Title,
     TextContent,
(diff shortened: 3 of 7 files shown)


packages/inventory-compliance/package.json
```diff
@@ -39,6 +39,7 @@
         "@redhat-cloud-services/frontend-components-notifications": "*",
         "@redhat-cloud-services/frontend-components-remediations": "*",
         "@redhat-cloud-services/frontend-components-utilities": "*",
+        "@apollo/react-hooks": "^3.1.5",
```
dLobatog, line 42: This is so I can `useQuery` in the EmptyState

packages/inventory-compliance/src/Compliance.js
```diff
@@ -79,15 +74,8 @@ class SystemDetails extends Component {
...
-            <Card className="ins-error-card">
-                <CardHeader>
-                    <NotEqualIcon />
-                </CardHeader>
-                <CardBody>
-                    <div>{ errorMsg }</div>
-                </CardBody>
-            </Card>;
+            <ComplianceEmptyState title='No policies are reporting for this system' /> :
+            <ErrorCard message={errorMsg} />;
```
dLobatog, line 78: Moved the error to its own component because it is now used in the EmptyState

packages/inventory-compliance/src/ComplianceEmptyState.js
```diff
@@ -11,16 +12,48 @@ import {
...
                 size="xl" title="Compliance" icon={CloudSecurityIcon} />
             <br/>
             <Title size="lg">{ title }</Title>
             <br/>
             <EmptyStateBody>
+                { policiesCount > 0 ?
+                    <TextContent>
+                        <a href='insights/compliance/scappolicies'>{policiesCount} {policyWord}</a> {haveWord} been created but {haveWord} no reports.
+                    </TextContent> :
+                    '' }
```
dLobatog, line 56: This is really the main change that's visible

packages/inventory-compliance/src/ComplianceEmptyState.js
```diff
@@ -54,7 +88,14 @@ ComplianceEmptyState.defaultProps = {
...
         Create new policy
-    </Button>
+    </Button>,
+    client: new ApolloClient({
+        link: new HttpLink({
+            uri: COMPLIANCE_API_ROOT + '/graphql',
+            credentials: 'include'
+        }),
+        cache: new InMemoryCache()
+    })
```
dLobatog, line 98: This might be surprising, because it looks like if you do:
```
import ComplianceEmptyState from ...;

<ApolloProvider client={client}>
  <ComplianceEmptyState/>
</ApolloProvider> 
```

it does not work. Apparently components that are imported as a dependency don't inherit the Apollo client from the context, so it has to have its own. In the case of the SystemDetails page, this would be unnecessary, but I kept this as the default so that the component is "batteries-included" and works without having to provide anything new. 

https://github.com/RedHatInsights/compliance-frontend/search?q=complianceemptystate&unscoped_q=complianceemptystate are the places where we use this and the client is not inherited from the state

packages/inventory-compliance/src/ComplianceEmptyState.js
```diff
@@ -11,16 +12,48 @@ import {
...
+const QUERY = gql`
+{
+    profiles(search: "external = false") {
+        totalCount
+    }
+}
+`;
+
+const ComplianceEmptyState = ({ title, mainButton, client }) => {
+    let { data, error, loading } = useQuery(QUERY, { fetchPolicy: 'network-only', client: client });
```
bastilian, line 30: You could also just pass down the count of policies from `SystemQuery` and avoid another query, can you not?

dLobatog: Unfortunately not - in SystemQuery we get the list of *policies associated to a host*, but in here I want the list of *all internal policies*.

Also, this component is used from `compliance-frontend`, in the CompliancePolicies component I could do that, but in the Reports component, we're only fetching the list of policies with `has_test_results = true`. So I'd have to change 2 out of 3 components to make an additional query to fetch the count of internal policies always when I only care if I want to load the empty state.

packages/inventory-compliance/src/ComplianceEmptyState.js
```diff
@@ -11,16 +12,48 @@ import {
...
+
+    return <Bullseye>
         <EmptyState>
             <EmptyStateIcon style={{ fontWeight: '500', color: 'var(--pf-global--primary-color--100)' }}
                 size="xl" title="Compliance" icon={CloudSecurityIcon} />
             <br/>
             <Title size="lg">{ title }</Title>
             <br/>
             <EmptyStateBody>
+                { policiesCount > 0 ?
```
bastilian, line 52: You can use `policiesCount > 0 && ...` here and not need the empty string.

packages/inventory-compliance/src/ComplianceEmptyState.test.js
```diff
@@ -2,9 +2,37 @@ import React from 'react';
...
 import ComplianceEmptyState from './ComplianceEmptyState';
+import { useQuery } from '@apollo/react-hooks';
+jest.mock('@apollo/react-hooks');
+jest.mock('apollo-boost');
 
 describe('ComplianceEmptyState', () => {
-    it('expect to render without error', () => {
+    it('expect to render without error if no policies exist', () => {
+        useQuery.mockImplementation(() => ({
+            data: { profiles: { totalCount: 0 } }, error: false, loading: false
```
bastilian, line 12: To stay in line with the what the `useQuery` API returns, I would suggets to use `undefined` instead of `false`.

packages/inventory-compliance/src/ComplianceEmptyState.js
bastilian: We could (and should at some point) also migrate to **not** use the `react-apollo` package.
The `@apollo/react-components` package provides a `Query` to use in the `SystemDetails` component.

packages/inventory-compliance/src/ComplianceEmptyState.test.js
```diff
@@ -2,11 +2,39 @@ import React from 'react';
...
 import ComplianceEmptyState from './ComplianceEmptyState';
+import { useQuery } from '@apollo/react-hooks';
+jest.mock('@apollo/react-hooks');
+jest.mock('apollo-boost');
 
 describe('ComplianceEmptyState', () => {
-    it('expect to render without error', () => {
+    it('expect to render without error if no policies exist', () => {
+        useQuery.mockImplementation(() => ({
+            data: { profiles: { totalCount: 0 } }, error: undefined, loading: false
```
bastilian, line 12: `useQuery` returns `undefined` for both `error` and `loading`

PR #534, opened 2020-06-02
RHICOMPL-572 - Prevent rule overflow in Compliance Rules table

dLobatog:
The Stack Patternfly component is display: flex and therefore it's
height is set as 0. This makes it impossible for the Table component to
understand how tall is each of the rows.

Instead, I'm switching the rows to use a simple TextContent component
which the Table can effortlessly use to figure out the correct height.

Before, note the overlapping identifiers on the 'removexattr' rows:  
Screenshot from 2020-06-02 14-01-13

After, note how each row text is vertically centered too: 
Screenshot from 2020-06-02 14-00-03

### DIFF
--- a/packages/inventory-compliance/src/PresentationalComponents/RuleTitle.js
+++ b/packages/inventory-compliance/src/PresentationalComponents/RuleTitle.js
@@ -1,23 +1,16 @@
 import React from 'react';
 import propTypes from 'prop-types';
 import {
-    Stack,
-    StackItem,
     Text,
     TextVariants,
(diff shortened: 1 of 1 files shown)


katierik:
Ack - Looks good and the rows are behaving as designed now.

dLobatog:
Travis never returned anything but it was :green_apple:  https://travis-ci.org/github/RedHatInsights/frontend-components/builds/694648395?utm_source=github_status&utm_medium=notification

PR #537, opened 2020-06-03
Increase version of react-intl and remove locale data

karelhala:


Hyperkid123:
@karelhala linter :skull:

PR #538, opened 2020-06-03
Shuffle components around, introduce wrapper component to properly load inv components

karelhala:
This is basically inventory rewrite. I know it's a lot of code to go trough, I can open 2 PRs one for table changes and the other one for detail changes if it'll help.

### New react-redux dependency
Since we introduced new name for react-redux dependency we should merge this after each app updates to the new dependencies
* [x] inventory https://github.com/RedHatInsights/insights-inventory-frontend/pull/611
* [x] vulnerability https://github.com/RedHatInsights/vulnerability-ui/pull/427
* [x] patch https://github.com/RedHatInsights/patchman-ui/pull/179
* [x] remediations https://github.com/RedHatInsights/insights-remediations-frontend/pull/181
* [x] advisor/insights https://github.com/RedHatInsights/insights-advisor-frontend/pull/688
* [x] compliance https://github.com/RedHatInsights/compliance-frontend/pull/638
* [x] drift https://github.com/RedHatInsights/drift-frontend/pull/328

### TODO
* [x] write tests


packages/inventory/config/webpack.constants.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
     externals: [
-        /^@patternfly\/.*/,
+        /^@patternfly\/react-core.*/,
+        /^@patternfly\/react-icons.*/,
```
Hyperkid123, line 4: i feel like this `/^@patternfly\/react-icons.*/` should be like this `/^@patternfly\/react-icons/*`

packages/inventory/src/components/detail/AppInfo.js
```diff
@@ -1,16 +1,18 @@
...
     render () {
-        const { activeApps, active, loaded } = this.props;
+        const { activeApps, active, loaded, componentsMapper, store } = this.props;
         const activeApp = activeApps.find(item => item.name === active.appName) || activeApps[0];
+        const Cmp = activeApp && activeApp.name ? componentsMapper?.[activeApp.name] || activeApp?.component : 'missing component';
         return (
             <Fragment>
                 { activeApp && <div className={ `ins-active-app-${activeApp.name}` }>
-                    { activeApp.component ? <activeApp.component /> : 'missing component' }
+                    { Cmp ? <Cmp store={store} /> : 'missing component'}
```
Hyperkid123, line 15: I see a `missing component` node in the `Cmp` declaration. Feels like the condition is not necessary

packages/inventory/src/index.js
```diff
@@ -1,7 +1,8 @@
...
 export * from './redux/reducers';
 export * from './redux/action-types';
 export * from './redux/actions';
-export { default as DeleteModal } from './DeleteModal';
-import './InventoryDetail.scss';
-import './InventoryList.scss';
+export { default as DeleteModal } from './shared/DeleteModal';
+export * from './components/filters';
+// import './InventoryDetail.scss';
+// import './InventoryList.scss';
```
Hyperkid123, line 8: Is this supposed to be a comment?

packages/inventory/src/redux/actions.js
```diff
@@ -4,32 +4,39 @@ import {
...
+    const itemIds = items.reduce((acc, curr) => (
+        [
+            ...acc,
+            curr && typeof curr === 'string' ? curr : curr.id
+        ]
+    ), []).filter(Boolean);
+    return {
+        type: ACTION_TYPES.LOAD_ENTITIES,
+        payload: getEntities(itemIds, config, showTags).then(({ results, ...data }) => {
+            return ({
```
Hyperkid123, line 26: extra return

packages/inventory/src/shared/Wrapper.js
```diff
@@ -0,0 +1,26 @@
+import React, { PureComponent, createRef } from 'react';
+import PropTypes from 'prop-types';
+
+class RenderWrapper extends PureComponent {
```
Hyperkid123, line 4: Is there any specific reason for this to be a class?

packages/inventory/src/shared/constants.js
```diff
@@ -193,3 +184,56 @@ export const staleness = [
 export const registered = [
     { label: 'Insights', value: 'insights' }
 ];
+
+export const InventoryContext = createContext({});
+
+export const loadSystems = (options, showTags) => {
+
+    console.log(options, 'Am I here?');
```
Hyperkid123, line 192: console

packages/inventory/src/components/detail/AppInfo.js
karelhala: Good point. I've change the `Cmp` declaration.

packages/inventory/src/api/api.js
```diff
@@ -0,0 +1,162 @@
...
+    ...oneResult,
+    rawFacts: facts,
+    facts: {
+        ...facts.reduce((acc, curr) => ({ ...acc, [curr.namespace]: curr.facts }), {}),
+        ...flatMap(facts, (oneFact => Object.values(oneFact)))
+        .map(item => typeof item !== 'string' ? ({
+            ...item,
+            // eslint-disable-next-line camelcase
+            os_release: item.os_release || item.release,
+            // eslint-disable-next-line camelcase
```
Hyperkid123, line 23: You have both global and eslint override for the same rule.

packages/inventory/src/components/table/EntityTableToolbar.js
```diff
@@ -0,0 +1,267 @@
...
+    const debounceGetAllTags = useCallback(debounce((config, options) => {
+        if (showTags && !hasItems) {
+            dispatch(fetchAllTags(config, options));
+        }
+    }, 800), []);
+
+    useEffect(() => {
+        const { textFilter, tagFilters, staleFilter, registeredWithFilter } = reduceFilters(filters);
+        debouncedRefresh();
+        ReactDOM.unstable_batchedUpdates(() => {
```
Hyperkid123, line 87: Wouldn't it be safer to switch to `useReducer` than using experimental features? This can break with any patch/minor update of React. I know that the react devs stated that this should not change but they have also stated that any unstable method should be used with caution :man_shrugging: 

Considering that react-dom is an external dependency it can cause issues with applications with an older package version.

karelhala: The problem is that each hook is using custom state, however this can be removed as it was just to prevent muliple re-renders (at one point was firing multiple GET requests, but I've fixed that in other places).

Hyperkid123: I think that this is a weak point and can cause unpredictable behavior in consumer apps. I would even prefer doing multiple re-renders over it.

karelhala: Yeah, good point. I am working on an enhancement to use reducer wih multiple hooks.

Hyperkid123:
@karelhala any word from the drift app team?

packages/inventory/src/components/filters/useTagsFilter.js
```diff
@@ -0,0 +1,85 @@
...
+        tagsFilter: payload
+    }
+});
+
+export const useTagsFilter = (
+    allTags = [],
+    loaded = false,
+    additionalTagsCount = 0,
+    onShowMoreClick,
+    [ globalState, dispatch ] = [ tagsFilterState ]
```
PanSpagetka, line 18: I feel a bit unsure about this line, could you explain it to me?

karelhala: Sure thing! It's pretty easy, it destructs array into 2 items and if no array is passed it will use defaul state so nothing breaks.

The call can look something like
```JS
const allTags = ['something'];
const allTagsLoaded = false;
const additionalTagsCount = 0;
const onToggleClicked = () => console.log('I was clicked!');
const state = {
  some: 'value'
};
const stateDispatch = () => console.log('dispatch was called!');
useTagsFilter(allTags, allTagsLoaded, additionalTagsCount, () => dispatch(toggleTagModal(true)), [ state, stateDispatch ]);
```

packages/inventory/src/components/table/EntityTableToolbar.js
```diff
@@ -0,0 +1,327 @@
...
+        tagsFilterReducer
+    ]), {
+        ...textFilterState,
+        ...stalenessFilterState,
+        ...registeredWithFilterState,
+        ...tagsFilterState
+    });
+    const filters = useSelector(({ entities: { activeFilters } }) => activeFilters || []);
+    const loaded = useSelector(({ entities: { loaded } }) => hasItems && isLoaded !== undefined ? (isLoaded && loaded) : loaded);
+    const allTagsLoaded = useSelector(({ entities: { allTagsLoaded } }) => allTagsLoaded);
```
Hyperkid123, line 76: ^^ all of these selectors with primitive selector type could use the shallow equal hook.

packages/inventory/src/components/table/InventoryTable.js
```diff
@@ -0,0 +1,85 @@
...
+    total,
+    page,
+    perPage,
+    filters,
+    showTags,
+    sortBy: propsSortBy,
+    ...props
+}, ref) => {
+    const hasItems = Boolean(items);
+    let pagination = {
```
Hyperkid123, line 29: should be const. I would be also a bit happier if the selectors would be used outside of the property declaration. They will be called on each render like this since the variable is always instantiated.

packages/inventory/src/components/filters/useTextFilter.js
```diff
@@ -0,0 +1,39 @@
...
+});
+
+export const useTextFilter = ([ state, dispatch ] = [ textFilterState ]) => {
+    let value;
+    let setValue;
+    if (dispatch) {
+        value = state.textFilter;
+        setValue = (newValue) => dispatch({ type: TEXT_FILTER, payload: newValue });
+    } else {
+        [ value, setValue ] = useState('');
```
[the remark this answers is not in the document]
karelhala: Addressd, what do you think about the new way?

packages/inventory/src/components/table/EntityTable.js
```diff
@@ -0,0 +1,151 @@
...
+    onSort,
+    expandable: isExpandable,
+    onRowClick,
+    noDetail
+}) => {
+    const dispatch = useDispatch();
+    const history = useHistory();
+    const loaded = useSelector(({ entities: { loaded } }) => (
+        hasItems && isLoaded !== undefined ? (isLoaded && loaded) : loaded
+    ));
```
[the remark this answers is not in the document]
karelhala: It's a boolean so I think that this one could be safe. But I'd add it there to be sure

packages/inventory/src/components/filters/useTextFilter.js
```diff
@@ -0,0 +1,36 @@
...
+export const textFilterReducer = (_state, { type, payload }) => ({
+    ...type === TEXT_FILTER && {
+        textFilter: payload
+    }
+});
+
+export const useTextFilter = ([ state, dispatch ] = [ textFilterState ]) => {
+    let [ value, setValue ] = useState('');
+    if (dispatch) {
+        value = state.textFilter;
```
Hyperkid123, line 15: what about initializing in the `useState` hook. Again this is probably something that might break the rules of hook:
```jsx
  let [ value, setValue ] = useState(dispatch ? state.textFilter : '');
```

packages/inventory/src/components/filters/useTextFilter.js
```diff
@@ -0,0 +1,36 @@
...
+        setValue = (newValue) => dispatch({ type: TEXT_FILTER, payload: newValue });
```
Hyperkid123, line 16: I would probably make a custom handler instead of overriding the `useState` callback
```jsx
handleSomething = newValue => dispatch ? dispatch(...) : setValue(...)
```

packages/inventory/src/components/filters/useTextFilter.js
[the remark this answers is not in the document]
Hyperkid123: I have one more comment about the variable overrides. I think rewriting the `useState` values might lead to some confusion. And potentially to memory leaks.

Hyperkid123:
@karelhala we could potentially create a decorator for these and just past the value and dispatch as the function argument and make the decision about the state mutators and values inside. So it can be re-used.

packages/inventory/src/components/filters/useTextFilter.js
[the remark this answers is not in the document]
karelhala: How about now? I wanted to avoid ternary in assign, but as you pointed this coul lead to confusion.

karelhala:
@Hyperkid123 naaah let's keep it this way. With custom decorator it would be even harder to understand, I think that ternary assign migh be the easiest to understand.

Hyperkid123:
> @Hyperkid123 naaah let's keep it this way. With custom decorator it would be even harder to understand, I think that ternary assign migh be the easiest to understand.

Allright :+1:

PR #539, opened 2020-06-03
Fixed missing file extension in babel config.

Hyperkid123:
The missing file extension can cause issues when resolving external dependencies. Files might not get imported and the module is undefined.

@karelhala this will probably require a new release of all packages. Is there a label for that?

cc @rvsia


rvsia:
:+1: This fixes the `timesCircleIcon is not defined` issue for cjs/sources.

PR #540, opened 2020-06-04
Release all packages to apply the babel config fix.

Hyperkid123:
Use the magical :mage_man:  power of new lines to release and apply the fix from #539 in all packages.

@rvsia this should release the fixed version of sources.


karelhala:
Prepare for the comment storm!

Hyperkid123:
@karelhala are we there yet?

PR #542, opened 2020-06-04
Fix sources v2 build

rvsia:
Sources build contains all PF CSS, some components, DDF components... this should get rid of it.

### DIFF
--- a/packages/sources/config/rollup.config.js
+++ b/packages/sources/config/rollup.config.js
@@ -6,7 +6,7 @@ import nodeGlobals from 'rollup-plugin-node-globals';
 import { terser } from 'rollup-plugin-terser';
 import postcss from 'rollup-plugin-postcss';
 import json from '@rollup/plugin-json';
-import { dependencies, peerDependencies, name } from '../package.json';
+import { dependencies, peerDependencies, devDependencies, name } from '../package.json';
 import {
     rollupConfig,
     externalDeps,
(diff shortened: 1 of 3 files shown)


packages/sources/config/rollup.config.js
```diff
@@ -44,15 +44,30 @@ const plugins = [
...
+        extract: true,
+        minimize: true
     }),
     json()
 ];
 
 export default rollupConfig(
-    external(externalDeps({ ...dependencies, ...peerDependencies })),
+    external([
+        ...externalDeps({ ...dependencies, ...peerDependencies, ...devDependencies }),
```
karelhala, line 55: No need to list all external deps with stars and such, you can do that in `externalDeps` with easier config. 
```
external(externalDeps(
  { ...dependencies, ...peerDependencies, ...devDependencies },
  [ '@data-driven-forms', 'lodash' ]
))
```

rvsia: But this will override the default externals, so shouldn't it be
```jsx
external(externalDeps(
  { ...dependencies, ...peerDependencies, ...devDependencies },
  ['@patternfly', '@redhat-cloud-services',  '@data-driven-forms', 'lodash' ]
))
```
?

karelhala: Oh yeah! You are right

packages/sources/config/rollup.config.js
```diff
@@ -46,15 +46,25 @@ const plugins = [
...
+        minimize: true
     }),
     json()
 ];
 
 export default rollupConfig(
-    external(externalDeps({ ...dependencies, ...peerDependencies })),
+    external(externalDeps(
+        { ...dependencies, ...peerDependencies, ...devDependencies },
+        [ '@patternfly', '@redhat-cloud-services',  '@data-driven-forms', 'lodash' ]
```
Hyperkid123, line 58: does the `externalDeps` add wildcards? (`@patternfly/**`)

rvsia: it transform deps to ``${item}/**``

Hyperkid123: I assume it keeps the original as well

PR #543, opened 2020-06-04
Use direct imports for PDF generator and do not lock react-dom version

karelhala:
When importing PDF generator, it adds 1.7 MB of redundant dependencies. This PR uses direct imports for PF react-tokens (removes about 800 KB) and uses caret version of react-dom (removes about 900 KB).

### DIFF
--- a/packages/pdf-generator/package.json
+++ b/packages/pdf-generator/package.json
@@ -34,8 +34,8 @@
         "@patternfly/react-styles": "^3.6.26",
         "@patternfly/react-tokens": "^2.7.25",
         "@react-pdf/renderer": "^1.6.8",
-        "react": "16.12.0",
-        "react-dom": "16.12.0",
+        "react": "^16.12.0",
+        "react-dom": "^16.12.0",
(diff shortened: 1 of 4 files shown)


packages/pdf-generator/package.json
```diff
@@ -34,8 +34,8 @@
         "@patternfly/react-styles": "^3.6.26",
         "@patternfly/react-tokens": "^2.7.25",
         "@react-pdf/renderer": "^1.6.8",
-        "react": "16.12.0",
-        "react-dom": "16.12.0",
+        "react": "^16.12.0",
```
Hyperkid123, line 37: Wasn't there some issue with a higher version if react/react-dom that made you lock them? I vaguely remember that there was some problem with async rendering.

karelhala: Oh, yeah. Mhh gotta use the locked version... It'll bload the bundle by 900 KB unfortunatelly.

Hyperkid123: Yeah :cry: sorry about reminding you.

PR #545, opened 2020-06-04
Add minified version of build

karelhala:


rvsia:
Is this also going to minify CSS files?

Hyperkid123:
@karelhala wasn't the terser plugin supposed to minify the code?

karelhala:
@Hyperkid123 it is, but react-dom and other react deps require `NODE_ENV` to be set in order to load correct files. Without this change when `react-dom` was required i loaded `react-dom.development.js` which is like 500 KB.

karelhala:
@rvsia since this will release new version of all packages I've included the minimize option to all rollup configs as well. Good call!

PR #547, opened 2020-06-05
Export sass in bundles and use them in insights

karelhala:


packages/components/config/rollup.config.js
```diff
@@ -58,7 +58,9 @@ const plugins = [
     copy({
         targets: [
             { src: 'components/index.css', dest: './' },
-            { src: 'components/index.js', dest: './' }
+            { src: 'components/index.js', dest: './' },
+            { src: 'src/**/*.scss', dest: './components/esm' },
```
Hyperkid123, line 62: dest can be an array

packages/inventory-insights/src/Insights.js
```diff
@@ -10,7 +10,7 @@ import { flatten, sortBy } from 'lodash';
 
 import API from './Api';
 import AnsibeTowerIcon from '@patternfly/react-icons/dist/js/icons/ansibeTower-icon';
-import { Battery } from '@redhat-cloud-services/frontend-components/components/Battery';
+import { Battery } from '@redhat-cloud-services/frontend-components/components/esm/Battery';
```
Hyperkid123, line 13: I think this should depend in the target env. This will duplicate certain assets in CJS builds

PR #548, opened 2020-06-05
If no kba article, hide kba section (gated off node_id param)

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-6584

Now that all the kba api funkiness is sorted out, if yah have ah node id, we know we'll have an article, which makes this ask real simple to fulfill

#### looks like this!
Screen Shot 2020-06-05 at 8 41 53 AM
Screen Shot 2020-06-05 at 8 41 33 AM

### DIFF
--- a/packages/inventory-insights/src/ReportDetails.js
+++ b/packages/inventory-insights/src/ReportDetails.js
@@ -86,21 +86,19 @@ const ReportDetails = ({ report, kbaDetail, kbaLoading }) => {
                         </CardBody>
                     </Card>
                 </StackItem>
-                <StackItem>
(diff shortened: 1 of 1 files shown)


AllenBW:
Yeah word for above is, if we have nothing show nothing

"
This enhancement was identified during in-depth rule review. When a recommendation does not have any knowledgebase article associated to it, we wish to hide this section from being shown in the UI to streamline the user consumption of the recommendation content.
"

AllenBW:
awww 💩 travis is green but wont report 😭

PR #549, opened 2020-06-05
Use sass imports for insights and rule component

karelhala:


karelhala:
Travis finished, but refuses to send report....

PR #554, opened 2020-06-08
Add inventory hook for easier usage

karelhala:


packages/utils/doc/useInventory.md
```diff
@@ -0,0 +1,62 @@
...
+```JSX
+import React, { useEffect, useState, Fragment } from 'react';
+import { useStore } from 'react-redux';
+import { useInventory } from '@redhat-cloud-services/frontend-components-utilities/files/useInventory';
+import { getRegistry } from '@redhat-cloud-services/frontend-components-utilities/files/Registry';
+
+const MyCmp = () => {
+    const [ConnectedInventory, setInventory] = useState();
+    const { InventoryTable, newReducers } = useInventory({
+        tableReducer: entitiesReducer,
```
AllenBW, line 14: Might want to specify from where `entitiesReducer` originates 🤔

packages/utils/doc/useInventory.md
```diff
@@ -0,0 +1,62 @@
...
+    useEffect(() => {
+        if (InventoryTable && newReducers) {
+            getRegistry().register(newReducers);
+            setInventory(() => InventoryTable);
+        }
+    }, [InventoryTable, newReducers]);
+
+    return (
+        <Fragment>
+            {ConnectedInventory && <ConnectedInventory />}
```
Hyperkid123, line 27: @karelhala I think it would be worth while to make the inventory component work on the same principle as a `lazy` and `Suspense` combo is working on.

You can return a component placeholder until the inventory is loaded and configured. We can avoid having this check and the above `setState` call. Waht do you think?

packages/utils/doc/useInventory.md
karelhala: Edded some more explanantion to the docs and proper import.

packages/utils/doc/useInventory.md
karelhala: The problem here was that the registry was causing too much re-renders. Fixed that by passing getter function for registry.

Hyperkid123: How come? We could limit that with the `memo` function.

karelhala: I tried that with `useCallback`, but didn't work as expected. The `useMemo` could fix it, if used in some other way, but that would only stop the problem. I think that passing a callback is slighlty better than passing the object.

Hyperkid123: Yeah I don't mind it I am just interested in why the issue has occurred.

PR #570, opened 2020-06-16
Deduplicate CSS file content.

Hyperkid123:
This is an example of how we can optimize CSS assets in this repo.

### Goals
Currently, each `*.css` file contains all the CSS for every component in any package. This is caused by the fact that we run multiple builds and the plugins are scanning for all the scss files in the src folder and then bundle them together (probably caused by the index.js, we could construct the index via script but I am not 100% sure that it is the cause of this problem).

There are two options:
- run rollup once per each file
  - this will have an unfortunate side effect of breaking the cjs shared files which is something we don't want imo
- build the CSS separately
  - that is what I have chosen for this example

The main focus should be on `packages/components/create-styles.js `
(currently, it's just for components but it is very generic and can be modified to be shared across all packages)
It scans the source folder for SCSS files and then builds independent CSS output. (index is still handled by rollup)

Each CSS file is deduplicated and contains styling only for a specific component:
```css
/**Truncate.css*/
:root{--ins-color--orange: #ec7a08}button:focus{outline:none}.ins-c-truncate button,.ins-c-truncate+button{padding:0}.ins-c-truncate.is-inline{margin-right:10px;margin-right:.625rem}

```


Hyperkid123:
failing tests are fixed in this PR: https://github.com/RedHatInsights/frontend-components/pull/571

Hyperkid123:
@karelhala do you want me to apply it on other packages? Are there even other packages that output multiple CSS files?

karelhala:
@Hyperkid123 I don't think so, if we'd see some problems in a future we can always copy this over to the package.

Hyperkid123:
Travis finished but the GH status is not updated :disappointed:

PR #573, opened 2020-06-16
Add dedicatedAction prop to PrimaryToolbar

LI0131:
This PR is in response to RHCLOUD-6054. This ticket explains a bug involving the RemediationsButton and RemediationsWizard as used in the Inventory Table within Advisor. The RemediationsWizard is not mounting when the button is listed as a `first-action` within the PrimaryToolbar actionConfig, since at a defined size the button appears in a dropdown menu. Similarly, this problem will occur if the button is listed later in the list of actions, since it will also be listed within the dropdown menu created in `Actions.js`.

This PR attempts to overcome this problem by creating a "dedicated" action, which will never enter the dropdown menu created in `Actions.js`. Below is an example of what this would look like on the Advisor Recommendations Table. The styling can be defined within the child component since the PropType for `dedicatedAction` is node.

Screen Shot 2020-06-16 at 3 11 36 PM


karelhala:
@LI0131 Thank you for the PR, looks like the problem is however on some different place. So the reason why first action goes to dropdown is because the real estate on smaller screens is limited, so the UX decided to move that into a kebab dropdown. That being said, some toolbras might actually want to have the first action always present. For instance if there is no filter we could show the first action still as a button.

Can you please attach, how this particular screen with button would look like on mobile version? If it wouldn't break the flow and such we could keep it. Any thought @katierik as you designed the original mocks, or anybody from @RedHatInsights/uxd

LI0131:
@karelhala ah! I figured that this was probably not the true source of the issue that I was having. I'll have to do some more research on that -- but glad this functionality might actually be useful 😁 . Here's what the screen would look like at a mobile screen size.

Screen Shot 2020-06-18 at 10 09 16 AM

AllenBW:
Revisiting this, @karelhala @katierik as this breaks nothin nor changes any design, any opposition to a merge of this?

PR #574, opened 2020-06-17
Upgrade inventory-compliance to Patternfly v4

bastilian:


karelhala:
@bastilian any updates on this? Looks like it's the last package to be updated. Do you mind merging and this and if we find any further bugs fix them later?

PR #575, opened 2020-06-17
Update inventory to patternfly-v4

karelhala:


packages/inventory/src/Inventory.js
```diff
@@ -12,6 +12,8 @@ import { Provider } from 'react-redux';
 import TagWithDialog from './TagWithDialog';
 export const InventoryContext = createContext('inventory');
 
+console.log(React, 'Inventory version!');
```
rvsia, line 15: Console.log

karelhala: Nice catch!

packages/inventory/src/InventoryList.scss
```diff
@@ -8,11 +8,19 @@
...
-        .pf-c-data-toolbar__item:last-child { @include rem('min-width', 200px); }
+
+.ins-c-inventory__table--toolbar {
+    &.ins-m-footer {
+        margin: 0;
+        padding: 0;
+    }
+
+    &:not(.ins-c-inventory__table--toolbar-has-items) {
+        .pf-c-data-toolbar__content {
```
Hyperkid123, line 19: The data toolbar component no longer exists. These classNames should change to the toolbar.

karelhala: Huh, I thought I already fixed them...

packages/remediations/config/webpack.constants.js
```diff
@@ -1,7 +1,6 @@
 module.exports = {
     externals: [
         /^@patternfly\/.*/,
-        /^@redhat-cloud-services\/.*/,
```
Hyperkid123, line 4: Is this intentional?

PR #581, opened 2020-06-23
Removed remediating modal and added created date

bond95:
image


doruskova:
@bond95 Thanks for the update. The published date should be closer to the title. The spacing between them should be smaller. And the lines of the title should be wrapped like this. Does your example act the same? 
Snímek obrazovky 2020-06-23 v 9 41 02

packages/rule-components/src/RuleTable/index.scss
```diff
@@ -1,7 +1,7 @@
-@import '~@redhat-cloud-services/frontend-components/components/battery.scss';
-@import '~@redhat-cloud-services/frontend-components/components/skeleton.css';
-@import '~@redhat-cloud-services/frontend-components/components/TableToolbar.scss';
-@import '~@redhat-cloud-services/frontend-components/components/primary-toolbar.scss';
+@import '~@redhat-cloud-services/frontend-components/components/Battery.css';
```
karelhala, line 1: Is there any reason why you moved to css instead of scss? It should be better to use sass files and css as fallback.

bond95:
@doruskova about padding, it will be fixed on OCM UI side. About acting as columns I think it could be possible if add title and actions directly to ReportDetails component, but I'm not sure that is good idea overweight component, @karelhala WDYT?

packages/rule-components/src/RuleTable/index.scss
bond95: I did this because when I did `npm start` it showed errors that, this files are not found.

karelhala: There were couple of big updates, recently. You should run `npm run clean; npm install` (in the roor folder) to clean and install the new versions

bond95: Ok, thank you

bond95:
Updated. @doruskova @karelhala please, take a look image

doruskova:
@bond95 looks good to me :)

PR #587, opened 2020-06-25
Fix group select by passing array of children and use label

karelhala:


packages/components/src/Components/ConditionalFilter/Group.js
```diff
@@ -236,13 +236,14 @@ class Group extends Component {
...
-                        return filteredItems.filter(Boolean).length > 0
+                        const filteredItems = this.mapItems({ groupValue, onSelect, groupLabel, groupId, type, items, ...group }, groupKey)
+                        .filter(Boolean);
+                        return filteredItems.length > 0
                             ? <SelectGroup
                                 {...group}
                                 key={groupId || groupValue || groupKey}
                                 value={groupId || groupValue || groupKey}
-                                label={groupLabel}
+                                label={`${groupLabel || ''}`}
```
Hyperkid123, line 246: `label={groupLabel || ''}` should be good enough

PR #591, opened 2020-06-26
Add option to use all details in inventory

karelhala:
Introduce detail render wrapper to render components from 3rd party react (bootstraped in chrome). If component mapper is passed use that instead of the things in redux. Use custom name for react so we can share react from app to inventory (not to details!)


packages/inventory/config/webpack.constants.js
```diff
@@ -31,7 +32,7 @@ module.exports = {
             'react-dom': 'react-dom',
             redux: 'redux',
             axios: 'axios',
-            react: 'react'
+            react: 'customReact'
```
Hyperkid123, line 35: Just to double-check. Does this mean that wen initializing inventory, the application will have to change the name of the react dependency when they call the `loadInventory` function?

karelhala: Nope, chrome does that for the app. I am working on a PR in chrome space.

karelhala: The chrome part can be found here https://github.com/RedHatInsights/insights-chrome/pull/856

packages/inventory/src/ApplicationDetails.js
```diff
@@ -11,26 +11,30 @@ class ApplicationDetails extends Component {
...
+        const { activeApp, items, appList } = this.props;
+        const defaultApp = (activeApp && activeApp.appName) ||
+            appList?.find(({ pageId, name }) => items?.[0]?.name === (pageId || name))?.name ||
+            items?.[0]?.name;
+        const applications = appList || items;
         return (
             <React.Fragment>
                 {
-                    items && items.length > 1 &&
+                    applications && applications.length > 1 &&
```
Hyperkid123, line 30: `applications?.length` will have the same effect `undefined > 0 === false`

packages/inventory/src/AppInfo.js
```diff
@@ -19,9 +27,18 @@ class AppInfo extends Component {
 }
 
 AppInfo.propTypes = {
+    store: PropTypes.any,
+    entity: PropTypes.shape({
+        id: PropTypes.string
+    }),
+    componentMapper: PropTypes.elementType,
```
Hyperkid123, line 34: I just thought of this. The element type is available only in recent versions of prop types. I would remove prop types from externals. It will be deduped and all apps should have prop types anyway

karelhala:
Closing in favor of https://github.com/RedHatInsights/frontend-components/pull/644

PR #592, opened 2020-06-29
PDF renderer table column header to support customization 

AllenBW:
while we are able to render a pdf with custom table cell styles (per rows), the headers remain unchangeable, SO if you wanted to have a column that was 200px wide, header wouldn't support that, table looks super wonky n outta place

this pr recreates the magic of the row for the header, if Imma send you something that isn't a string, show it in it's entirety 💝

related to https://github.com/RedHatInsights/insights-advisor-frontend/pull/680


AllenBW:
oooooooo naco-bot whys it taking 3 hours to release?

PR #594, opened 2020-07-01
Fix notifications babel config for direct imports

elad661:
Many parts of the documentation suggests importing the `addNotification` action directly.

However, that didn't work with the babel-plugin-transform-imports configuration offered in the README file.
This commit fixes the documented configuration to allow direct import of `addNotification` and `removeNotification`.

cc: @karelhala

### DIFF
--- a/packages/notifications/README.md
+++ b/packages/notifications/README.md
@@ -52,6 +52,10 @@ Since our components require a bit more setting up, we are recommending using `b
 Change your babel config to be javascript config `babel.config.js` so you can use JS functions to properly transform your imports
 
 ```JS
+notificationsMapper = {
+    addNotification: 'actions',
+    removeNotification: 'actions',
+};
 module.exports = {
     presets: [
         // your presets go here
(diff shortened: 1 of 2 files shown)


elad661:
Ugh, looks like while this suggested change fixes the import of `{ addNotification }`, it breaks the import of the `{ notifications }` reducer due to the `skipDefaultConversion` flag - it tries to import the name `notifications` from `notifications.js` instead of using the default.

I'm still trying to find ways to make this work, but I'm not sure that's possible without changing the exports structure. Any ideas?

karelhala:
@elad661 thanks for taking a lookg at this! I think that the best way here is to add additional export to notifications reducer and remove `skipDefaultConversion` from the docs.
```diff
diff --git a/packages/notifications/src/redux/reducers/notifications.js b/packages/notifications/src/redux/reducers/notifications.js
index 76e64714..aacd33a4 100644
--- a/packages/notifications/src/redux/reducers/notifications.js
+++ b/packages/notifications/src/redux/reducers/notifications.js
@@ -26,4 +26,6 @@ export const notificationsReducers = {
     [CLEAR_NOTIFICATIONS]: clearNotifications
 };
 
-export default applyReducerHash(notificationsReducers, defaultState);
+export const notifications = applyReducerHash(notificationsReducers, defaultState);
+
+export default notifications;
```

elad661:
Thanks @karelhala, I updated the commit accordingly

karelhala:
Build passed, travis just won't update the status... https://travis-ci.org/github/RedHatInsights/frontend-components/builds/704241129

PR #596, opened 2020-07-01
Change modal title to say display name instead of name

karelhala:
Fixes: https://projects.engineering.redhat.com/browse/RHCLOUD-5875

### New UI
Screenshot from 2020-07-01 17-22-52


karelhala:
@ryelo there's also `Edit Ansible host`, if we are going to use sentence case, let's use it in both!

ryelo:
I think `Edit Ansible host` is fine because `Ansible` is a proper noun whereas `display` is not

karelhala:
@ryelo good point!

PR #600, opened 2020-07-08
Make sources translatable

rvsia:
Make all texts in Sources wizard translatable

KNOWN ISSUES:
- [x] substepOf is not translated
- [x] translated step titles throws warnings

I will fix both issues in DDF. (Addressed here https://github.com/data-driven-forms/react-forms/pull/636 )

Also, there is one aria-label untranslated. Right now, make it translated would take to much time...

TODO:
- [x] tests

Please, can you take a look if everything is OK? @Hyperkid123  @boaz0 

(To check it in UI, uncomment `// modules: [resolve('./node_modules')], // uncomment when using linked package` in webpack config and update intl to the latest)

### DIFF
--- a/packages/sources/src/addSourceWizard/SSLFormLabel.js
+++ b/packages/sources/src/addSourceWizard/SSLFormLabel.js
@@ -1,23 +1,29 @@
 import React from 'react';
 import { QuestionCircleIcon } from '@patternfly/react-icons';
 import { Popover, Text, TextContent, TextVariants } from '@patternfly/react-core';
+import { FormattedMessage } from 'react-intl';
 
 const SSLFormLabel = () => (
     <React.Fragment>
-        SSL Certificate&nbsp;
+        <FormattedMessage id="wizard.SslCertificate" defaultMessage="SSL Certificate" />&nbsp;
         <Popover
             aria-label="Help text"
             maxWidth="50%"
             bodyContent={
                 <TextContent>
                     <Text component={ TextVariants.p }>
-                            You can obtain your OpenShift Container Platform provider’s CA
-                            certificate for all endpoints (default, metrics, alerts) from
-                        <b>/etc/origin/master/ca.crt</b>.
+                        <FormattedMessage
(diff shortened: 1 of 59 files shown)


packages/sources/src/addSourceWizard/SSLFormLabel.js
```diff
@@ -1,23 +1,29 @@
...
             aria-label="Help text"
             maxWidth="50%"
             bodyContent={
                 <TextContent>
                     <Text component={ TextVariants.p }>
-                            You can obtain your OpenShift Container Platform provider’s CA
-                            certificate for all endpoints (default, metrics, alerts) from
-                        <b>/etc/origin/master/ca.crt</b>.
+                        <FormattedMessage
+                            id="wizard.YouCanObtainYourOpenshiftContainerPlatformProviderSCaCertificateForAllEndpointsDefaultMetricsAlertsFrom"
```
Hyperkid123, line 16: I would definitely make these ids a bit more friendly. Don't forget that eventually, humans will have to go through the JSON and translate it. I would not want to read stuff like this.

Hyperkid123:
@rvsia we should also add the "scrape" script so we can get the translation structure. You can check this snippet: https://github.com/RedHatInsights/catalog-ui/pull/648/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R114

packages/sources/src/addSourceWizard/SSLFormLabel.js
rvsia: This can be changed later. Right now there is no common format and no one is translating it, so it doesn't matter. I don't want to come up with 207+ unique ids and scrap them later. :D

Hyperkid123: It will be up to us to make the ids. Nobody (i hope:crossed_fingers:) can dictate stuff like that. But we can change it later.

rvsia:
@Hyperkid123  this is ready to merge

PR #603, opened 2020-07-08
Update react-redux and react-router-dom to newest versions

karelhala:


Hyperkid123:
@karelhala can you check the failing test?

karelhala:
@Hyperkid123 good point, the component we are speaking about is kinda broken and fires multiple times. We already have a ticket in our JIRA to fix that.

PR #604, opened 2020-07-09
Enhance conditional filter with show more option

fhlavac:
* added "Show more" button and callback property
* updated styling - hidden group labels & hidden check icons

https://projects.engineering.redhat.com/browse/RHCLOUD-7737

@karelhala 
@PanSpagetka

### DIFF
--- a/packages/components/src/Components/ConditionalFilter/ConditionalFilter.js
+++ b/packages/components/src/Components/ConditionalFilter/ConditionalFilter.js
@@ -83,6 +83,7 @@ class ConditionalFilter extends Component {
                             {
                                 ActiveComponent && <SplitItem isFilled>
                                     <ActiveComponent
+                                        onShowMore={activeItem.onShowMore}
                                         {
                                         ...activeItem.type !== conditionalFilterType.custom &&
                                             {
--- a/packages/components/src/Components/ConditionalFilter/ConditionalFilter.js
+++ b/packages/components/src/Components/ConditionalFilter/ConditionalFilter.js
@@ -84,6 +84,7 @@ class ConditionalFilter extends Component {
                                 ActiveComponent && <SplitItem isFilled>
                                     <ActiveComponent
                                         onShowMore={activeItem.onShowMore}
+                                        showMoreTitle={activeItem.showMoreTitle}
                                         {
                                         ...activeItem.type !== conditionalFilterType.custom &&
                                             {
--- a/packages/components/src/Components/ConditionalFilter/Group.js
+++ b/packages/components/src/Components/ConditionalFilter/Group.js
@@ -54,7 +54,9 @@ class Group extends Component {
                 (groupValue && input.test(groupValue)) ||
                 (groupLabel && input.test(groupLabel)) ||
                 (item.value && input.test(item.value)) ||
-                (item.label && input.test(item.label)))
+                (item.label && input.test(item.label)) ||
+                item.isPersistentAction
+            )
         ).map(({ value, isChecked, onClick, label, props: itemProps, id, ...item }, key) => (
             <SelectOption
                 {...item}
@@ -67,7 +69,7 @@ class Group extends Component {
                         e.stopPropagation();
                     }
 
-                    if (e.target.matches('.pf-c-select__option-button') && onClick) {
+                    if (item.isPersistentAction) {
                         return onClick();
                     }
 
@@ -227,7 +229,7 @@ class Group extends Component {
         const showMore = {
             type: groupType.button,
             variant: 'link',
-            items: [{ label: 'Show more', type: groupType.button, onClick: onShowMore }]
+            items: [{ label: 'Show more', type: groupType.button, onClick: onShowMore, isPersistentAction: true }]
         };
         return (<Fragment>
             { !filterItems || (filterItems && filterItems.length <= 0) ? <Text { ...this.props } value={ `${selected}` } /> : <Select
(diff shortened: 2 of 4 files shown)


packages/components/src/Components/ConditionalFilter/ConditionalFilter.js
```diff
@@ -83,6 +83,8 @@ class ConditionalFilter extends Component {
                             {
                                 ActiveComponent && <SplitItem isFilled>
                                     <ActiveComponent
+                                        isFilterable={activeItem.isFilterable}
```
karelhala, line 86: Is there a reason for the `isFilterable` prop?

packages/components/src/Components/ConditionalFilter/Group.js
```diff
@@ -57,6 +59,7 @@ class Group extends Component {
         )
         ).map(({ value, isChecked, onClick, label, props: itemProps, id, ...item }, key) => (
             <SelectOption
+                className=""
```
karelhala, line 62: No need to assign empy `className`

packages/components/src/Components/ConditionalFilter/Group.js
```diff
@@ -215,9 +221,13 @@ class Group extends Component {
 
     render() {
         const { isExpanded, filterBy } = this.state;
-        const { groups, items, placeholder, className, selected, isFilterable, isDisabled, onFilter } = this.props;
-
+        const { groups, items, placeholder, className, selected, isFilterable, isDisabled, onFilter, onShowMore } = this.props;
         const filterItems = items || groups;
+        const showMore = {
```
karelhala, line 226: You can move that to consts file, since you are not accessing any props etc.

packages/components/src/Components/ConditionalFilter/Group.js
```diff
@@ -235,11 +245,19 @@ class Group extends Component {
                 { ...groups && groups.length > 0 && { isGrouped: true }}
             >
                 { groups && groups.length > 0 ? (
-                    groups.map(({ value: groupValue, onSelect, label: groupLabel, id: groupId, type, items, ...group }, groupKey) => {
+                    [ ...groups, onShowMore ? showMore : undefined ].map(({
```
karelhala, line 248: You should either spread an array
```JS
[
   ...groups,
  ...onShowMore ? [showMore]: []
]
```

Or you can use false value
```JS
[ ...groups, onShowMore && showMore ]
```

packages/components/src/Components/ConditionalFilter/Group.js
fhlavac: I suppose `onShowMore` property will be accessed there

packages/components/src/Components/ConditionalFilter/Group.js
```diff
@@ -38,23 +39,21 @@ class Group extends Component {
...
         let input;
-
         try {
             input = new RegExp(filterBy, 'i');
         } catch (err) {
             input = new RegExp(filterBy.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'), 'i');
         }
 
-        return items.filter(item => onFilter || (
+        return items.filter(item =>
```
karelhala, line 52: Is there a reason to remove the `onFilter` function?

packages/components/src/Components/ConditionalFilter/Group.js
```diff
@@ -67,6 +66,8 @@ class Group extends Component {
                         e.stopPropagation();
                     }
 
+                    if (e.target.type === 'button' && onClick) { return onClick(); }
```
karelhala, line 69: I guess that the problem is with buttons and calling `onSelect` etc, how about we move this a bit down and check if target matches https://developer.mozilla.org/en-US/docs/Web/API/Element/matches instead of calling `type`?

packages/components/src/Components/ConditionalFilter/Group.js
```diff
@@ -118,13 +119,16 @@ class Group extends Component {
...
+                                : [
                                 // we have to wrap it in array, otherwise PF will complain
-                                (type !== groupType.checkbox && type !== groupType.radio) ? label : ''
-                            ]
+                                    (type !== groupType.checkbox && type !== groupType.radio) ? label : ''
+                                ]
                 }
             </SelectOption>
-        ));
+        )) || null;
```
PanSpagetka, line 131: I think that this can be deleted

packages/components/src/Components/ConditionalFilter/Group.js
fhlavac: oh I was trying something and missed not writing it back then, thanks for catching that!

packages/components/src/Components/ConditionalFilter/ConditionalFilter.js
fhlavac: no, I missed that it is in filterItems, thanks!

packages/components/src/Components/ConditionalFilter/Group.js
```diff
@@ -215,10 +223,14 @@ class Group extends Component {
...
         const { isExpanded, filterBy } = this.state;
-        const { groups, items, placeholder, className, selected, isFilterable, isDisabled, onFilter } = this.props;
-
+        const { groups, items, placeholder, className, selected, isFilterable, isDisabled, onFilter, onShowMore } = this.props;
         const filterItems = items || groups;
 
+        const showMore = {
+            type: groupType.button,
+            variant: 'link',
+            items: [{ label: 'Show more', type: groupType.button, onClick: onShowMore, isPersistentAction: true }]
```
karelhala, line 232: Can we have this changable? Like default to `Show more`, but if app consumer decide that it should show `Show more 10 items` we should give them the option to do that.

PR #605, opened 2020-07-09
Update IAM policy to have the min requirement

boaz0:
Updating JSON to contain only the necessary permissions.

//cc @rvsia

### DIFF
--- a/packages/sources/src/addSourceWizard/hardcodedComponents/aws/arn.js
+++ b/packages/sources/src/addSourceWizard/hardcodedComponents/aws/arn.js
@@ -85,33 +85,19 @@ export const IAMPolicyDescription = () => {
             <ClipboardCopy isCode variant={ClipboardCopyVariant.expansion} className="pf-u-m-sm-on-sm" isReadOnly>
                 {JSON.stringify({
                     Version: '2012-10-17',
-                    Statement: [
-                        {
-                            Sid: 'VisualEditor0',
-                            Effect: 'Allow',
-                            Action: [
-                                's3:Get*',
-                                's3:List*'
-                            ],
-                            Resource: [
-                                `arn:aws:s3:::${s3Bucket}`,
-                                `arn:aws:s3:::${s3Bucket}/*`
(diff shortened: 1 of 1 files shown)


packages/sources/src/addSourceWizard/hardcodedComponents/aws/arn.js
```diff
@@ -84,34 +84,20 @@ export const IAMPolicyDescription = () => {
...
-                    Statement: [
-                        {
-                            Sid: 'VisualEditor0',
-                            Effect: 'Allow',
-                            Action: [
-                                's3:Get*',
-                                's3:List*'
-                            ],
-                            Resource: [
-                                `arn:aws:s3:::${s3Bucket}`,
```
rvsia, line 97: s3Bucket si no longer used, so you can remove the formOptions from this component

packages/sources/src/addSourceWizard/hardcodedComponents/aws/arn.js
```diff
@@ -84,34 +84,20 @@ export const IAMPolicyDescription = () => {
...
-                            Resource: '*'
-                        }
-                    ]
+                    "Version": "2012-10-17",
+                    "Statement": [{
+                        "Sid": "VisualEditor0",
+                        "Effect": "Allow",
+                        "Action": ["s3:Get*", "s3:List*"],
+                        "Resource": [
+                            "arn:aws:s3:::hccm-nise-data-01",
```
rvsia, line 93: or it is? should not it be here?

rvsia: same below

packages/sources/src/addSourceWizard/hardcodedComponents/aws/arn.js
boaz0: Right! I will fix it. Thanks

boaz0:
Thanks @rvsia

PR #606, opened 2020-07-09
Add Severity Line Chart

LI0131:
This PR adds a new charts to the `charts` package. This chart displays a level of severity from level 1 (low) up to level 4 (critical). The specifications for this chart are given [here](https://marvelapp.com/prototype/8d7cj33/screen/70992845).

Here's what it will look like:
Screen Shot 2020-07-09 at 2 57 58 PM

### DIFF
--- a/packages/charts/doc/severityLine.md
+++ b/packages/charts/doc/severityLine.md
@@ -1,0 +1,23 @@
+# Severity Line Chart
+
+Display a level of severity given a value between 1-Low and 4-Critical severity.
+
+```jsx
+import React from react;
+import { SeverityLine } from '@redhat-cloud-services/frontend-components-charts';
+
+<SeverityLine value={2} title='High Severity'/>
+```
+
+## Props
+
+```jsx
+SeverityLine.propTypes = {
+    value: PropTypes.number.isRequired,
+    title: PropTypes.string.isRequired,
+    config: PropTypes.shape({
+        height: PropTypes.number,
+        width: PropTypes.number
+    })
+};
+```
--- a/packages/charts/src/Charts/SeverityLine/SeverityLine.js
+++ b/packages/charts/src/Charts/SeverityLine/SeverityLine.js
@@ -1,0 +1,65 @@
+import './severity-line.scss';
+
+import React, { useMemo } from 'react';
+
+import PropTypes from 'prop-types';
+import { Tooltip } from '@patternfly/react-core';
+
+const chartMapper = (width) => ({
+    1: '1',
+    2: (width / 3),
+    3: (width * (2 / 3)),
+    4: (width - 1)
+});
+
+const SeverityLine = ({ title, value, tooltipMessage, config, chartProps }) => {
+
+    const { width, height } = config;
+    const severity = useMemo(() => chartMapper(width)?.[value], [ width, value ]);
+    const points = [
+        { className: 'horizontal', points: `${1}, ${height / 2} ${width - 1}, ${height / 2}` },
+        { className: 'vertical', points: `${width - 1}, ${0} ${width - 1}, ${height}` },
+        { className: 'vertical', points: `${width * (2 / 3)}, ${0} ${width * (2 / 3)}, ${height}` },
+        { className: 'vertical', points: `${width / 3}, ${0} ${width / 3}, ${height}` },
+        { className: 'vertical', points: `${1}, ${0} ${1}, ${height}` },
+        { className: `dataline-${value}`, points: `${1}, ${height / 2} ${severity}, ${height / 2}` }
+    ];
+
+    return <div className='ins-c-severity-line' { ...chartProps }>
+        <div className='ins-l-title'>
+            {title.length > 14 ? <Tooltip content={title}>
+                <span> {`${title.substring(0, 14)}...`} </span>
+            </Tooltip> : <span> {title} </span>}
+        </div>
+        {severity && <div className='ins-l-chart'>
+            <svg height={height} width={width} viewBox={`0 0 ${width} ${height}`}>
+                {points.map(({ className, points }, key) => <polyline key={key} className={className} points={points} />)}
+                {tooltipMessage ? <Tooltip content={tooltipMessage}>
+                    <circle className={`dataline-${value}`} cx={`${severity}`} cy={`${height / 2}`} r={`${2}`}/>
+                </Tooltip> : <circle className={`dataline-${value}`} cx={`${severity}`} cy={`${height / 2}`} r={`${2}`}/>}
+            </svg>
+        </div>}
+    </div>;
+};
+
+SeverityLine.propTypes = {
+    value: PropTypes.number.isRequired,
+    title: PropTypes.string.isRequired,
+    tooltipMessage: PropTypes.string,
+    config: PropTypes.shape({
+        height: PropTypes.number,
+        width: PropTypes.number
+    }),
+    chartProps: PropTypes.shape({
+        [PropTypes.string]: PropTypes.any
+    })
+};
+
+SeverityLine.defaultProps = {
+    config: {
+        height: 12,
+        width: 302
+    }
+};
+
+export default SeverityLine;
--- a/packages/charts/src/Charts/SeverityLine/severity-line.scss
+++ b/packages/charts/src/Charts/SeverityLine/severity-line.scss
@@ -1,0 +1,62 @@
+.ins-c-severity-line {
+    --ins-c-severity-line-text:         var(--pf-global--Color--200);
+    --inc-c-severity-line-base:         var(--pf-global--palette--black-400);
+    --ins-c-severity-line-low:          var(--pf-global--palette--blue-200);
+    --ins-c-severity-line-moderate:     var(--pf-global--palette--gold-200);
+    --ins-c-severity-line-high:         var(--pf-global--palette--orange-200);
+    --ins-c-severity-line-critical:     var(--pf-global--palette--red-100);
+
+    display: inline-flex;
+}
+
+.ins-c-severity-line {
+    .ins-l-title {
+        width: 115px;
+        span {
+            color: var(--inc-c-severity-line-text);
+            font-size: var(--pf-global--FontSize--sm);
+        }
+    }
+}
+
+.ins-c-severity-line {
+    .ins-l-chart {
+        position: relative;
+        svg {
+            overflow: visible;
+            .horizontal {
+                stroke: var(--inc-c-severity-line-base);
+                stroke-width: 2px;
+            }
+
+            .vertical {
+                stroke: var(--inc-c-severity-line-base);
+                stroke-width: 1px;
+            }
+
+            .dataline-1 {
+                stroke: var(--ins-c-severity-line-low);
+                fill: var(--ins-c-severity-line-low);
+                stroke-width: 4px;
+            }
+
+            .dataline-2 {
+                stroke: var(--ins-c-severity-line-moderate);
+                fill: var(--ins-c-severity-line-moderate);
+                stroke-width: 4px;
+            }
+
+            .dataline-3 {
+                stroke: var(--ins-c-severity-line-high);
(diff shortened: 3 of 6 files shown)


packages/charts/src/Charts/SeverityLine/SeverityLine.js
```diff
@@ -0,0 +1,77 @@
+import './severity-line.scss';
+
+import React, { Component } from 'react';
+
+import PropTypes from 'prop-types';
+
+class SeverityLine extends Component {
```
karelhala, line 7: I'd prefer function component for this, everything should be doable with `useState` hook.

packages/charts/src/Charts/SeverityLine/SeverityLine.js
```diff
@@ -0,0 +1,77 @@
...
+        super(props);
+        this._getSeverityValue = this._getSeverityValue.bind(this);
+        this.state = {
+            severity: 0
+        };
+    }
+
+    _getSeverityValue(value) {
+        const { config: { width } } = this.props;
+        const CHART_MAP = { 1: '1', 2: (width / 3), 3: (width * (2 / 3)), 4: (width - 1) };
```
karelhala, line 19: Can you move this above the component and space it a bit better?
```JS
const chartMapper = (width) => ({
  1: '1',
  2: (width / 3),
  3: (width * (2 / 3)),
  4: (width - 1)
})
```

packages/charts/src/Charts/SeverityLine/SeverityLine.js
```diff
@@ -0,0 +1,77 @@
...
+        this.setState({
+            severity: CHART_MAP[value]
```
karelhala, line 21: You can set the severity with proposed function as `chartMapper(width)?.[value]`

packages/charts/src/Charts/SeverityLine/SeverityLine.js
```diff
@@ -0,0 +1,77 @@
...
+                width
+            }
+        } = this.props;
+        return <div className='ins-c-severity-line'>
+            <div className='ins-l-title'>
+                <span> {title} </span>
+            </div>
+            <div className='ins-l-chart'>
+                <svg height={height} width={width} viewBox={`0 0 ${width} ${height}`}>
+                    <polyline id='horizontal' points={`${1}, ${height / 2} ${width - 1}, ${height / 2}`}/>
```
karelhala, line 45: I'd prefer `className` over `id` as this prevents using this chart multiple times on same page.

packages/charts/src/Charts/SeverityLine/SeverityLine.js
LI0131: Hey @karelhala! Thanks for the review 😁. Would 
```jsx 
useEffect(() => {
        value && setSeverity(CHART_MAP[value]);
    }, [value, setSeverity]);
```
be an acceptable way to format this change? I'm not 100% sure I understand how `chartMapper(width)?.[value]` works.

packages/charts/doc/severityLine.md
```diff
@@ -0,0 +1,23 @@
+# Severity Line Chart
+
+Display a level of severity given a value between 1-Low and 4-Critical severity.
+
+```jsx
+import React from react;
+import { SeverityLine } from '@redhat-cloud-services/frontend-components-charts';
+
+<SeverityLine value={2} title='High Severity'>
```
AllenBW, line 9: what happens when a really long title is entered?

packages/charts/src/Charts/SeverityLine/SeverityLine.js
AllenBW, line 45: could this be ah `.map` throw the points  strings in an array?

packages/charts/src/Charts/SeverityLine/severity-line.scss
```diff
@@ -0,0 +1,61 @@
+.ins-c-severity-line {
+    --inc-c-severity-line-base:         var(--pf-global--palette--black-400);
```
AllenBW, line 2: toats appreciate how yah make the color usage here very clear 🥰

packages/charts/doc/severityLine.md
ryelo, line 9: Nitpick, but this should be `<SeverityLine value={2} title='High Severity'/>`

packages/charts/src/Charts/SeverityLine/severity-line.scss
```diff
@@ -0,0 +1,61 @@
...
+    --ins-c-severity-line-critical:     var(--pf-global--palette--red-100);
+
+    display: inline-flex;
+}
+
+.ins-c-severity-line {
+    .ins-l-title {
+        width: 115px;
+        span {
+            color: var(--inc-c-severity-line-base);
```
ryelo, line 15: Looking at the marvel mockups, the title here seems a little too light. I'd just double check this passes with UX.

packages/charts/doc/severityLine.md
LI0131: Added in the Tooltip and truncation like we talked about 😁

packages/charts/src/Charts/SeverityLine/severity-line.scss
LI0131: I update this to the value they actually specified, so should be good now 👍

packages/charts/src/Charts/SeverityLine/SeverityLine.js
Hyperkid123: @LI0131 it is effectively the same thing. the `?.` notations exist to avoid trying to access object property on an undefined variable and causing runtime crash.
```jsx
x = undefined
console.log(x?.foo) // will not fail but logs undefined
console.log(x.foo) // will thrown an error (trying to call property foo on undefined)

```

packages/charts/src/Charts/SeverityLine/SeverityLine.js
```diff
@@ -0,0 +1,73 @@
...
+    ];
+    const truncate = (string) => `${string.substring(0, 14)}...`;
+
+    useEffect(() => {
+        value && setSeverity(CHART_MAP[value]);
+        //eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [ value ]);
+
+    useEffect(() => {
+        title.length > 14 && setIsTruncated(true);
```
karelhala, line 35: Since we are using this state value only at one place (line 40) you can move this calculation to that line and avoid multiple re-renders (every time `title` changes it fires new render and the value will be calculated ad-hoc.)

packages/charts/src/Charts/SeverityLine/SeverityLine.js
```diff
@@ -0,0 +1,73 @@
...
+import { Tooltip, TooltipPosition } from '@patternfly/react-core/dist/js/components/Tooltip/index';
+
+import PropTypes from 'prop-types';
+
+const SeverityLine = ({ title, value, tooltipMessage, config }) => {
+
+    const [ severity, setSeverity ] = useState();
+    const [ isTruncated, setIsTruncated ] = useState(false);
+    const { width, height } = config;
+    const CHART_MAP = {
```
karelhala, line 13: If you don't like the conditional accessor you can still make this into a function (and move it above the component definition), that way it won't be re-assigned on each render

```JS
const chartMapper = (width, value) => {
  return {
    1: '1',
    2: (width / 3),
    3: (width * (2 / 3)),
    4: (width - 1)
  }[value]
};
```

packages/charts/src/Charts/SeverityLine/SeverityLine.js
```diff
@@ -0,0 +1,73 @@
...
+        1: '1',
+        2: (width / 3),
+        3: (width * (2 / 3)),
+        4: (width - 1)
+    };
+    const points = [
```
karelhala, line 19: I really like this approach!

packages/charts/src/Charts/SeverityLine/SeverityLine.js
```diff
@@ -0,0 +1,73 @@
+import './severity-line.scss';
+
+import React, { useEffect, useState } from 'react';
+import { Tooltip, TooltipPosition } from '@patternfly/react-core/dist/js/components/Tooltip/index';
+
+import PropTypes from 'prop-types';
+
+const SeverityLine = ({ title, value, tooltipMessage, config }) => {
```
karelhala, line 8: You should take the rest of props in here and spread them in the div on line 38.

packages/charts/src/Charts/SeverityLine/SeverityLine.js
```diff
@@ -0,0 +1,73 @@
+import './severity-line.scss';
+
+import React, { useEffect, useState } from 'react';
+import { Tooltip, TooltipPosition } from '@patternfly/react-core/dist/js/components/Tooltip/index';
```
karelhala, line 4: No need to import like this, the app will decide what to use later on, you can use `import { Tooltip, TooltipPosition } from '@patternfly/react-core';`

packages/charts/src/Charts/SeverityLine/SeverityLine.js
LI0131: @Hyperkid123 Oh awesome, thanks so much for clearing that up for me. That seems like it's super helpful 😁

packages/charts/src/Charts/SeverityLine/SeverityLine.js
LI0131: I added an additional prop `chartProps` where additional props could be passed. Is that what you are intending?

packages/charts/src/Charts/SeverityLine/SeverityLine.js
```diff
@@ -0,0 +1,71 @@
...
+        //eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [ value ]);
+
+    return <div className='ins-c-severity-line' { ...chartProps }>
+        <div className='ins-l-title'>
+            {title.length > 14 ? <Tooltip content={title}>
+                <span> {`${title.substring(0, 14)}...`} </span>
+            </Tooltip> : <span> {title} </span>}
+        </div>
+        {severity && <div className='ins-l-chart'>
```
rvsia, line 39: I think you don't have to use state `severity`. Just compute the value here.

packages/charts/src/Charts/SeverityLine/SeverityLine.js
```diff
@@ -0,0 +1,71 @@
...
+        height: PropTypes.number,
+        width: PropTypes.number
+    }),
+    chartProps: PropTypes.shape({
+        [PropTypes.string]: PropTypes.any
+    })
+};
+
+SeverityLine.defaultProps = {
+    tooltipMessage: undefined,
```
rvsia, line 64: No need to define undefined props.

packages/charts/src/Charts/SeverityLine/SeverityLine.js
karelhala: `severity` is used on multiple places, I think it's fine to keep as state variable. Otherwise the variable would have been calculated on each re-render.

packages/charts/src/Charts/SeverityLine/SeverityLine.js
karelhala: Just this small fix and we are golden!

packages/charts/src/Charts/SeverityLine/SeverityLine.js
Hyperkid123: Is severity ever changed? If not  `useRef` is a better way to persist variables like that than `useState`.

rvsia: Yeah, it can be computed in a variable and used multiple times. And it's not so complex computation and I think the only reason to re-render this component will be to update the value, so in the end it will be basically the same, just with no need to use expensive hook. (And because of the useEffect, the component will be rendered twice on the initial render.) We can use memo to optimize it further.

PR #609, opened 2020-07-13
Add custom labels

LI0131:
This PR adds the following custom labels to the `redhat-cloud-services/frontend-components` package:
Screen Shot 2020-07-13 at 3 15 22 PM
Screen Shot 2020-07-13 at 3 14 22 PM
Screen Shot 2020-07-13 at 3 14 35 PM
Screen Shot 2020-07-13 at 3 14 02 PM

The icons can either be show or hidden based on use case. The specifications for these labels are listed in the mocks [here](https://marvelapp.com/prototype/8d7cj33/screen/70992845) and are pertinent to [this](https://projects.engineering.redhat.com/browse/RHCLOUD-7702) JIRA 🎫

### DIFF
--- a/packages/components/src/Components/InsightsLabel/CriticalIcon.js
+++ b/packages/components/src/Components/InsightsLabel/CriticalIcon.js
@@ -1,0 +1,9 @@
+import './critical-icon.scss';
+
+import React from 'react';
+
+const CriticalIcon = () => <svg className='ins-c-critical-icon' viewbox='0 0 10 10'>
+    <polygon points='10 10, 10 3, 5 0, 0 3, 0 10, 5 8'/>
+</svg>;
+
+export default CriticalIcon;
--- a/packages/components/src/Components/InsightsLabel/InsightsLabel.js
+++ b/packages/components/src/Components/InsightsLabel/InsightsLabel.js
@@ -1,0 +1,39 @@
+import './labels.scss';
+
+import { AngleDoubleDownIcon } from '@patternfly/react-icons';
+import { AngleDoubleUpIcon } from '@patternfly/react-icons';
+import CriticalIcon from './CriticalIcon';
+import { EqualsIcon } from '@patternfly/react-icons';
+import { Label } from '@patternfly/react-core';
+import PropTypes from 'prop-types';
+import React  from 'react';
+import classNames from 'classnames';
+
+const VALUE_TO_STATE = {
+    1: { icon: <AngleDoubleDownIcon/>, text: 'Low' },
+    2: { icon: <EqualsIcon/>, text: 'Moderate' },
+    3: { icon: <AngleDoubleUpIcon/>, text: 'High' },
+    4: { icon: <CriticalIcon/>, text: 'Critical' }
+};
+
+const InsightsLabel = ({ value, text, hideIcon, className, rest }) => {
+    return <Label
+        { ...rest }
+        className={ classNames(className, `ins-c-label-${value}`) }
+        icon={ !hideIcon && VALUE_TO_STATE[value].icon }
+    >
+        { text || VALUE_TO_STATE[value].text }
+    </Label>;
+};
+
+InsightsLabel.propTypes = {
+    value: PropTypes.number,
+    text: PropTypes.node,
+    hideIcon: PropTypes.bool
+};
+
+InsightsLabel.defaultProps = {
+    value: 1
+};
+
+export default InsightsLabel;
--- a/packages/components/src/Components/InsightsLabel/InsightsLabel.test.js
+++ b/packages/components/src/Components/InsightsLabel/InsightsLabel.test.js
@@ -1,0 +1,23 @@
+import { AngleDoubleDownIcon } from '@patternfly/react-icons';
+import InsightsLabel from './InsightsLabel';
+import React from 'react';
+import { shallow } from 'enzyme';
+import toJson from 'enzyme-to-json';
+
+describe('InsightsLabel component', () => {
+    it('should render with icon', () => {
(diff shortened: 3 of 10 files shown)


packages/components/src/Components/InsightsLabel/InsightsLabel.js
```diff
@@ -0,0 +1,55 @@
...
+const VALUE_TO_STATE = {
+    1: { icon: <AngleDoubleDownIcon/>, text: 'Low' },
+    2: { icon: <EqualsIcon/>, text: 'Moderate' },
+    3: { icon: <AngleDoubleUpIcon/>, text: 'High' },
+    4: { icon: <CriticalIcon/>, text: 'Critical' }
+};
+
+const InsightsLabel = ({ value, text, hideIcon, labelProps }) => {
+
+    const [ icon, setIcon ] = useState();
```
rvsia, line 21: Why do you use state to store the icon text and icon? 

It seems as an overhead to me. The component will be rendered at least twice on initial render (render, setState, setState... the state could be stored as one value when it’s so tightly coupled..)

You can compute the props at render time:

```jsx
InsightsLabel = (props) => <Label
         { ...labelProps }
         className={`ins-c-label-${value}`}
         icon={!hideIcon && VALUE_TO_STATE[value].icon}
     >
         { text || VALUE_TO_STATE[value].text }
     </Label>;
```

packages/components/src/Components/InsightsLabel/InsightsLabel.js
```diff
@@ -0,0 +1,55 @@
...
+import PropTypes from 'prop-types';
+
+const VALUE_TO_STATE = {
+    1: { icon: <AngleDoubleDownIcon/>, text: 'Low' },
+    2: { icon: <EqualsIcon/>, text: 'Moderate' },
+    3: { icon: <AngleDoubleUpIcon/>, text: 'High' },
+    4: { icon: <CriticalIcon/>, text: 'Critical' }
+};
+
+const InsightsLabel = ({ value, text, hideIcon, labelProps }) => {
```
rvsia, line 19: I think labelProps can be replaced by using `...rest` and `{...rest}` as it is the only component so I don’t see any reason to specify it.

packages/components/src/Components/InsightsLabel/InsightsLabel.js
```diff
@@ -0,0 +1,55 @@
...
+    text: PropTypes.string,
+    hideIcon: PropTypes.bool,
+    labelProps: PropTypes.shape({
+        [PropTypes.string]: PropTypes.any
+    })
+};
+
+InsightsLabel.defaultProps = {
+    value: 1,
+    hideIcon: false
```
rvsia, line 52: No need to set up a default false value.

packages/components/src/Components/InsightsLabel/InsightsLabel.test.js
```diff
@@ -0,0 +1,19 @@
+import InsightsLabel from './InsightsLabel';
+import React from 'react';
+import { shallow } from 'enzyme';
+import toJson from 'enzyme-to-json';
+
+describe('InsightsLabel component', () => {
+    it('should render with icon', () => {
+        const wrapper = shallow(<InsightsLabel value={1} showIcon/>);
```
rvsia, line 8: I thinks these tests are out of the sync: you have `showIcon` here. The test below does the same.

I would switch to explicitly checking if the icon is rendered or not: `expect(wrapper.find(Icon)).toHaveLenght(0)` or you can check if the PF label gets the right props.

packages/components/src/Components/InsightsLabel/InsightsLabel.js
```diff
@@ -0,0 +1,55 @@
...
+        const state = VALUE_TO_STATE?.[value];
+        if (state) {
+            setIcon(state.icon);
+            setLabelText(text || state.text);
+        }
+    }, [ setIcon, value ]);
+
+    return <Label
+        { ...labelProps }
+        className={`ins-c-label-${value}`}
```
rvsia, line 34: What if users will pass custom `className` to this component?

packages/components/src/Components/InsightsLabel/InsightsLabel.js
```diff
@@ -0,0 +1,55 @@
...
+        icon={!hideIcon && icon}
+    >
+        { labelText }
+    </Label>;
+};
+
+InsightsLabel.propTypes = {
+    value: PropTypes.number,
+    text: PropTypes.string,
```
rvsia, line 43: I would use node to allow put there anything. (For example, react intl is using elements to translate texts.)

packages/components/src/Components/InsightsLabel/__snapshots__/InsightsLabel.test.js.snap
```diff
@@ -0,0 +1,21 @@
...
+
+exports[`InsightsLabel component should render defaults 1`] = `
+<Component
+  className="ins-c-label-1"
+  icon={false}
+/>
+`;
+
+exports[`InsightsLabel component should render with icon 1`] = `
+<Component
```
rvsia, line 11: You should change `shallow` to `mount` or use `shallow.dive()`. These snapshots are not very useful.

packages/components/src/Components/InsightsLabel/CriticalIcon.js
```diff
@@ -0,0 +1,11 @@
+import './critical-icon.scss';
+
+import React from 'react';
+
+const CriticalIcon = () => <React.Fragment>
```
karelhala, line 5: Do we need to wrap it in `Fragment`? Since it's just a `svg` I think we are good without it.

terezanovotna:
adding @bdiring for review

PR #612, opened 2020-07-14
Start adding types to frontend-components

josejulio:
Initial PR to add some types (.d.ts) for the frontend-components

So far I've added:
- Main
- Dark
- PageHeader
- Section
- Skeleton
- Spinner

I'm only considering the public API of these (e.g. the ones that you can import directly from '@redhat-cloud-services/frontend-components'


karelhala:
@josejulio did you find an automated way how to do this, or you'd have to write that by hand? If by hand, we might want to sacrifice some time in plat-exp team and figure more automated way of doing this.

josejulio:
> @josejulio did you find an automated way how to do this, or you'd have to write that by hand? If by hand, we might want to sacrifice some time in plat-exp team and figure more automated way of doing this.

Its manually, that's how is usually done when you want to add types to a lib that is purely javascript. 
These `.d.ts` are bundled in the lib or as a [separated lib](https://github.com/DefinitelyTyped/DefinitelyTyped) for the types.

The only automated way I can think of is to write the code in typescript, not sure if there are others, but haven't researched a lot on this area.

Manual way should be stable as we are only writing the API of the public components, of course it would require updating if said API changes.

PR #614, opened 2020-07-15
Hotfix compliance intl so it won't break the entire screen

karelhala:
This is effectively the same as https://github.com/RedHatInsights/frontend-components/pull/598


karelhala:
Tests are failing on something completely different, can be merged since it is fix for older version.

bastilian:
Thank you @karelhala!

PR #617, opened 2020-07-16
Fetch subscription watch IAM policy config from API

rvsia:
subscription watch IAM policy is being loaded from the API.

It there is any error, a text stating that user should go back and try it again is used. (Or should we use hardcoded config instead of it?)

Success

success

Error

error

@infinitewarp @katherine-black


katherine-black:
Yay! Thank you so much for this change @rvsia!

PR #619, opened 2020-07-16
Make label styles important

LI0131:
This PR turns...

Screen Shot 2020-07-16 at 2 36 56 PM

into this...

Screen Shot 2020-07-16 at 2 34 29 PM

by making the styles important in the scss here.

### DIFF
--- a/packages/components/src/Components/InsightsLabel/labels.scss
+++ b/packages/components/src/Components/InsightsLabel/labels.scss
@@ -14,41 +14,41 @@
 }
 
 .ins-c-label-4 {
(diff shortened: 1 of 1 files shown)


karelhala:
@LI0131 @AllenBW can we avoid using `!important` in future and rather make the selector more specific so it will be applied properly? This way, nobody can ever change the styling (not that anybody would ever want to) without changing this code.

AllenBW:
@karelhala sure thing!  s a bummer the gold and orange values are just a little different, would be much simpler to use the default label colors here

PR #621, opened 2020-07-17
Check status of applications after source creation

rvsia:
https://projects.engineering.redhat.com/browse/TPINVTRY-1017
https://projects.engineering.redhat.com/browse/TPINVTRY-1016
https://projects.engineering.redhat.com/browse/TPINVTRY-1015
https://projects.engineering.redhat.com/browse/TPINVTRY-1014

**Timeouted source**

image

**Unsuccessful configuration**

image

Remove option

errorremove

Edit option

erroredit

**Some not-input error**

image

**Loading**

image

TODO:
- [x] what to show when app is not being checked?
  - [x] check source/application at least once
- [x] tests
- [x] integration with Add application wizard


packages/sources/src/addSourceWizard/FinalWizard.js
```diff
@@ -17,39 +21,104 @@ const FinalWizard = ({
...
+                secondaryActions={<Button variant="link" onClick={ reset }><FormattedMessage id="wizard.addAnotherSource" defaultMessage="Add another source"/></Button>}
+            />;
+        } else if (createdSource.applications[0]?.availability_status === 'unavailable') {
+            step = <ErroredStep
+                onRetry={ onRetry }
+                onClose={ afterSubmit }
+                secondaryActions={<Button variant="link" onClick={ removeSource }>
+                    {intl.formatMessage({ id: 'wizard.removeSource', defaultMessage: 'Remove source' })}
+                </Button>}
+                primaryAction={() => push(`/sources/edit/${createdSource.id}`)}
```
Hyperkid123, line 70: It's actually preferable to use link elements `a` for better accessibility. Is it possible to wrap this action into router link?

packages/sources/src/styles/cardSelect.scss
```diff
@@ -50,3 +50,7 @@ $card-height: 90px;
       text-overflow: ellipsis;
     }
 }
+
+.ins-c-sources__wizard--step-text {
+  max-width: 536px;
```
Hyperkid123, line 55: Is there any specific reason for static value? I would prefer some relative value based on parent element.

packages/sources/src/tests/addSourceWizard/__snapshots__/finalWizard.test.js.snap
```diff
@@ -1,4722 +0,0 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
```
Hyperkid123, line 1: This snapshot is no longer relevant. Either snapshot the router child or remove this test.

packages/sources/src/styles/cardSelect.scss
rvsia: ^ @beccaglass

beccaglass: This adheres to PF guidelines from my understanding. It should be built into the empty state component.

Hyperkid123: Alright. I would expect there would be some sort of variable? (--pf-c-empty-state__content--MaxWidth would be my guess)

beccaglass: This is what I'm seeing in PF docs:
https://www.patternfly.org/v4/documentation/core/components/emptystate

pf-c-empty-state__body

Hyperkid123: Nope, it was a good guess. I am just confused about the value
```css
.pf-c-empty-state__content {
    max-width: var(--pf-c-empty-state__content--MaxWidth); <-- is none
}
```
@beccaglass do you happen to know if this is the correct value?

packages/sources/src/addSourceWizard/FinalWizard.js
rvsia: added `component="a"`

packages/sources/src/styles/cardSelect.scss
Hyperkid123: I don't see any width attribute on `pf-c-empty-state__body`. Anyway, I don't want to be pedantic. You probably know better than me.

rvsia: I am not sure if I believe PF variables anymore, after the changes of success color... :smiley:

packages/sources/src/addSourceWizard/FinalWizard.js
Hyperkid123: I don't think the `a` component will work it will use standard HTML `a` tag and probably will do full page refresh. You would have to prevent the event bubble (or use a router link).

rvsia: It works.

packages/sources/src/styles/cardSelect.scss
beccaglass: All I'm seeing in the PF template is that text wraps at 536px... no mention of a variable :/

packages/sources/src/addSourceWizard/FinalWizard.js
Hyperkid123: Did you give it the `href` (to) property? That is the point of using the link element.

rvsia: Will fix it when the web works. :facepalm:

PR #629, opened 2020-07-20
Allow custom classNames in SeverityLine chart

LI0131:
@AllenBW This PR will allow us to pass custom classNames to the SeverityLine component


AllenBW:
ugh. travis is done n not reporting :rage4:

PR #632, opened 2020-07-20
Updating Cost Management Doc link URL to 4.5

chambridge:
The source creation wizard links are pointing to OCP 4.4 docs and those docs at that URL no longer exist due to the GA of 4.5.

We are investigating better ways of handling this with Content Services, so that we don't have to update this every time OCP is updated.


rvsia:
Maybe OCP could redirect these links automatically or it should make '/latest/' route alias. :thinking:

PR #634, opened 2020-07-21
Fix API error step in sources wizard

rvsia:
I removed these props from previous commits.. :-/

### DIFF
--- a/packages/sources/src/addSourceWizard/FinalWizard.js
+++ b/packages/sources/src/addSourceWizard/FinalWizard.js
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
-import { Wizard, Button } from '@patternfly/react-core';
+import { Wizard, Button, Text } from '@patternfly/react-core';
 import { FormattedMessage, useIntl } from 'react-intl';
 import { Link } from 'react-router-dom';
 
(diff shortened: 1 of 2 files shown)


packages/sources/src/addSourceWizard/FinalWizard.js
```diff
@@ -93,6 +93,12 @@ const FinalWizard = ({
         step = <ErroredStep
             onClose={ afterError }
             primaryAction={tryAgain}
+            secondaryActions={
+                <Text component='a' target="_blank" href="https://access.redhat.com/support/cases/#/case/new/open-case?caseCreate=true" rel="noopener noreferrer">
+                    {intl.formatMessage({ id: 'wizard.openTicket', defaultMessage: 'Open a support case' })}
+                </Text>
+            }
+            returnButtonTitle={ <FormattedMessage id="wizard.retryText" defaultMessage="Retry"/> }
```
Hyperkid123, line 101: can you use `formatMessage` here as well? (just to have both messages using the same API)

rvsia: I plan to make a PR that will change this everywhere in wizard package after I done with these changes.

Hyperkid123: alright then

PR #642, opened 2020-07-22
Add validating and validated states for source.name field

rvsia:
This add two states for source.name

**Validating**

image

**Validated**

image

**Error message**

image

Example

resolveProps

Any thoughts, @beccaglass ?

### DIFF
--- a/packages/sources/package.json
+++ b/packages/sources/package.json
@@ -28,8 +28,8 @@
         "awesome-debounce-promise": "^2.1.0"
     },
     "devDependencies": {
-        "@data-driven-forms/pf4-component-mapper": "^2.6.7",
-        "@data-driven-forms/react-form-renderer": "^2.6.7",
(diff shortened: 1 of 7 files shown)


beccaglass:
@rvsia looks great! Can we just change the error text to "That name is taken. Try another."

rvsia:
@beccaglass updated

image

packages/sources/package.json
```diff
@@ -28,8 +28,8 @@
         "awesome-debounce-promise": "^2.1.0"
     },
     "devDependencies": {
-        "@data-driven-forms/pf4-component-mapper": "^2.6.7",
-        "@data-driven-forms/react-form-renderer": "^2.6.7",
+        "@data-driven-forms/pf4-component-mapper": "^2.8.3",
```
Hyperkid123, line 31: make sure to bump the peer dependency as well (good practices and all that)

rvsia: Oh, you are right :sob:

PR #643, opened 2020-07-22
Allow custom text for PDF battery

LI0131:
This PR lets us use a custom text field to specify a unique title for any of the battery icon variants presently available in the `pdf-generator` Battery component. In advisor, this would allow us to change `High` to `Important` to better reflect the `Total Risk` hierarchy.

Screen Shot 2020-07-22 at 8 18 48 AM

### DIFF
--- a/packages/pdf-generator/src/components/Battery.js
+++ b/packages/pdf-generator/src/components/Battery.js
@@ -3,7 +3,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import {
     Canvas,
-    Text,
+    Text as PDFText,
     View
 } from '@react-pdf/renderer';
 // eslint-disable-next-line camelcase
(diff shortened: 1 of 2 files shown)


packages/pdf-generator/src/components/Battery.js
```diff
@@ -72,17 +72,18 @@ const Battery = ({ variant, ...props }) => {
...
-        </Text>
+            {text || currBattery.title}
+        </PDFText>
     </View>;
 };
 
 Battery.propTypes = {
-    variant: PropTypes.oneOf(Object.keys(batteryMapper))
+    variant: PropTypes.oneOf(Object.keys(batteryMapper)),
+    text: PropTypes.oneOfType([PropTypes.string, PropTypes.node])
```
rvsia, line 86: PropTypes.node is enough (string is also a node)

PR #644, opened 2020-07-22
Allow all details to be passed to inventory detail

karelhala:


packages/inventory/src/components/detail/AppInfo.js
```diff
@@ -9,21 +9,28 @@ import { Skeleton, SkeletonSize } from '@redhat-cloud-services/frontend-componen
...
     const loaded = useSelector(({ entityDetails: { loaded } }) => loaded);
+    const entity = useSelector(({ entityDetails: { entity } }) => entity);
     const activeApp = useSelector(({ entityDetails: { activeApps, activeApp, loaded } }) => {
         if (loaded) {
-            return activeApps?.find?.(item => item?.name === activeApp?.appName) || activeApps?.[0];
+            return (appList || activeApps)?.find?.(item => item?.name === activeApp?.appName) || activeApps?.[0];
         }
     });
-    const Cmp = componentsMapper?.[activeApp?.name] || activeApp?.component;
+    const Cmp = componentMapper || activeApp?.component;
```
Hyperkid123, line 21: How does this work if the `componentMapper` is an `object`? I think it will trigger the `cannot render objects as children` error. Or is the object prop type referring to react component?

packages/inventory/src/components/detail/ApplicationDetails.js
```diff
@@ -38,6 +42,11 @@ const ApplicationDetails = ({ onTabSelect, ...props }) => {
 };
 
 ApplicationDetails.propTypes = {
+    appList: PropTypes.arrayOf(PropTypes.shape({
+        title: PropTypes.node,
+        name: PropTypes.string,
```
Hyperkid123, line 47: Name seems to be required. Otherwise, the find inside onSelect callback will nor provide correct results.

packages/inventory/src/components/detail/AppInfo.js
karelhala: Oh yeah, I should remove the object from the props. Good point!

PR #646, opened 2020-07-22
Allow custom classnames in FilterChips component

LI0131:
This PR allows for the passage of a custom className prop to the FilterChips component in the `frontend-components` package.


packages/components/src/Components/FilterChips/FilterChips.js
```diff
@@ -38,12 +39,13 @@ const FilterChips = ({ filters, onDelete }) => {
                     </Chip>
                 </ChipGroup>
             )) }
-            { filters.length > 0 && <Button variant="link" onClick={ (event) => onDelete(event, filters, true) }>Clear filters</Button> }
```
karelhala, line 41: No need for this change.

PR #647, opened 2020-07-23
Spacing in platform toolbar

PanSpagetka:
In first two commits I have updated and imported Patternfly to the demo app. I think that in most cases developer would like to PF to be included, but even if not it is easier to turn it off than it was for me now turning it on.

Third commit is the actual spacing change for https://projects.engineering.redhat.com/browse/RHCLOUD-7554
Now there should be 16px bottom/top padding and 24 left/right padding. I think that other spacing is correct.

ping  @mmenestr @mattnolting

Screenshot from 2020-07-23 15-33-38

Screenshot from 2020-07-23 14-17-56
Screenshot from 2020-07-23 14-17-09
Screenshot from 2020-07-23 14-15-53
Screenshot from 2020-07-23 14-15-12
Screenshot from 2020-07-23 15-31-34


packages/demo/src/index.scss
```diff
@@ -1 +1,4 @@
+@import '~@patternfly/patternfly/patternfly.css';
```
karelhala, line 1: No need to commit these changes.

mmenestr:
Can you send me a screenshot of the inspect, with the spacers showing? The 24px is built into the toolbar, where as the 16px are above and below it (separate from the toolbar).

Also did you change the spacing inside the toolbar itself as well?

mmenestr:
Based on our Slack convo, it seems like the toolbar is now inline with the PF one! So it's good to merge. Thanks Robin!!

karelhala:
@PanSpagetka fix the failing travis by updating snapshots and we are good to go!

mattnolting:
Hey guys, this PR addresses your padding issues: https://github.com/patternfly/patternfly/pull/3317. We've recognized the padding discrepancies across components and are addressing them. In this case, we've updated the toolbar default padding to match `data-list`, `table` and `page`. We're also adding `inset`s to adjust fitting.

PanSpagetka:
Ok, I will wait for new PF and change our styling accordingly. I will leave this PR open until then.

karelhala (approved):

epwinchell:
@karelhala  @PanSpagetka Just an FYI - I tested against this branch and it seems unrelated to the bug I reported here: https://issues.redhat.com/browse/RHCLOUD-9144

PR #653, opened 2020-07-23
Fix component types

josejulio:
- Use HTMLElement instead of the non-existant React.HTMLElement
- Fix PageHeaderTitleProps (Omit expects 2 arguments, not 3)

Tested with https://github.com/ansible/ansible-hub-ui project using [yalc](https://github.com/whitecolor/yalc)


josejulio:
@karelhala could you take a look?
This fixes the issues with the definition files, also I'm now using the typescript compiler. This will ensure that the definitions aren't broken by any error in them.

Also this gives me some ideas on how to add `ts` files to this project without the problem I had before.

PR #661, opened 2020-07-29
Bug fix: Change status page link in Unavailable to go to status.redhat.com

LeshemNoa:
Before it led to a relative `./status` path.


elad661:
cc: @karelhala 

The link in the component was broken - cloud.redhat.com/status doesn't exist, and neither does cloud.redhat.com/openshift/status

our understanding is that this link should point to status.redhat.com, which shows the current status of OCM APIs and Insights APIs

PR #664, opened 2020-07-31
Updated Total risk icon in rule-components

bond95:
image
image
image


doruskova:
@bond95 looks good :)

PR #669, opened 2020-08-04
Add detail wrapper component and add basic inv info to it

karelhala:
Screenshot from 2020-08-04 17-44-14


packages/inventory/src/components/detail/TopBar.js
```diff
@@ -37,7 +42,21 @@ const TopBar = ({
...
                     loaded ? (
-                        <Title headingLevel="h1" size='2xl'>{ entity && entity.display_name }</Title>
+                        <Flex>
+                            <FlexItem>
+                                <Title headingLevel="h1" size='2xl'>{ entity && entity.display_name }</Title>
+                            </FlexItem>
+                            {
+                                showTags &&
+                                <FlexItem>
+                                    {loaded ?
```
Hyperkid123, line 52: This condition does not make sense. `loaded` must always be `true` at this point.

Hyperkid123:
@karelhala npm seems to have some issues atm (we have the same problem on multiple respositories)

Hyperkid123:
@karelhala is this ready for review or do you plan to do some changes?

karelhala:
@Hyperkid123 ready for review, I was just fixing small bugs here and there.

PR #671, opened 2020-08-05
Make PF react completely external

karelhala:


Hyperkid123:
@karelhala there is just missing jest config.

PR #673, opened 2020-08-06
Add sitemap plugin

PanSpagetka:
Add webpack plugin that generate sitemap as a part of build. Based on routes defined in `package,json` of build application.

Related: 
https://github.com/RedHatInsights/insights-rbac-ui/pull/395

### DIFF
--- a/packages/config/src/sitemap-plugin.js
+++ b/packages/config/src/sitemap-plugin.js
@@ -11,7 +11,7 @@ class SitemapPlugin {
         compiler.hooks.emit.tapAsync('SitemapPlugin', (compilation, callback) => {
             const packageJson = require(resolve(this.rootFolder, 'package.json'));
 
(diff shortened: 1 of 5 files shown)


packages/config/src/sitemap-plugin.js
```diff
@@ -0,0 +1,25 @@
...
+class SitemapPlugin {
+    constructor(rootFolder) {
+        this.rootFolder = rootFolder;
+    }
+
+    apply(compiler) {
+        compiler.hooks.emit.tapAsync('SitemapPlugin', (compilation, callback) => {
+            const packageJson = require(resolve(this.rootFolder, 'package.json'));
+
+            const paths = Object.values(packageJson.routes).map(value => value.path || value);
```
karelhala, line 14: One final adjustment so the build won't fail when `routes` are not defined
```suggestion
            const paths = Object.values(packageJson?.routes || {}).map(value => value.path || value);
```

PR #679, opened 2020-08-07
Fix remediations wrong hooks by using correct alias externals

karelhala:


karelhala:
Just a note that this will release remediations and utils package, in order to fix this properly in all apps chrome needs to update remediations package and each app has to update both utils and remediations packages.

PR #684, opened 2020-08-10
Fix remediation file generator by adding proper css chunks

karelhala:
Currently the build creates `1.css` which is incorrect, so I've changed the css generation to create `index.css`.

### DIFF
--- a/packages/remediations/.gitignore
+++ b/packages/remediations/.gitignore
@@ -7,3 +7,4 @@
 /esm
 /cjs
 /*.css
+*.txt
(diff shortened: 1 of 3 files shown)


packages/remediations/.gitignore
```diff
@@ -7,3 +7,4 @@
 /esm
 /cjs
 /*.css
+*.txt
```
Hyperkid123, line 10: was this change intended?

karelhala: Yup, the build now creates `.txt` files containing some license info from minimize plugin.

PR #688, opened 2020-08-12
Add global filter component

karelhala:
Screenshot from 2020-08-27 16-05-50

### DIFF
--- a/packages/components/src/Components/FilterHooks/constants.js
+++ b/packages/components/src/Components/FilterHooks/constants.js
@@ -2,35 +2,45 @@ import React from 'react';
 import { Badge, Tooltip } from '@patternfly/react-core';
 
 export function mapGroups(currSelection, valuesKey = 'values') {
-    return Object.entries(currSelection || {}).map(([ groupKey, groupValue ]) => {
+    return Object.entries(currSelection || {}).reduce((acc, [ groupKey, groupValue ]) => {
         const values = constructValues(groupValue, groupKey);
-        if (values?.length > 0) {
-            return {
-                type: 'tags',
-                key: groupKey,
-                category: values[0]?.group?.value || values[0]?.group?.label,
-                [valuesKey]: values
-            };
+        if (values.length > 0) {
+            return [
+                ...acc,
+                {
+                    type: 'tags',
+                    key: groupKey,
+                    category: values[0]?.group?.value || values[0]?.group?.label,
+                    [valuesKey]: values
+                }
+            ];
         }
-    }).filter(Boolean);
+
+        return acc;
+    }, []);
 }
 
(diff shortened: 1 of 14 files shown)


packages/components/src/Components/FilterHooks/constants.js
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import { Badge, Tooltip } from '@patternfly/react-core';
+
+export function mapGroups(currSelection, valuesKey = 'values') {
+    return Object.entries(currSelection || {}).map(([ groupKey, groupValue ]) => {
+        const values = constructValues(groupValue, groupKey);
+        if (values?.length > 0) {
```
Hyperkid123, line 7: values should be always at least empty array. No need for extra `?`.

packages/components/src/Components/FilterHooks/constants.js
```diff
@@ -0,0 +1,69 @@
...
+            };
+            return {
+                key,
+                tagKey,
+                value: tagValue,
+                name: `${tagKey}${tagValue ? `=${tagValue}` : ''}`,
+                group: { value: groupKey, ...group }
+            };
+        }
+    }).filter(Boolean);
```
Hyperkid123, line 33: Both map and filter actions could be done in a single `reduce` iteration. It should make it slightly quicker.

packages/components/src/Components/FilterHooks/tagFilterHook.js
```diff
@@ -0,0 +1,76 @@
+import React, { useState, useEffect } from 'react';
+import { constructGroups, mapGroups } from './constants';
+import { Spinner } from '@patternfly/react-core/dist/esm/components/Spinner';
```
Hyperkid123, line 3: We should not directly reference `esm` builds. https://github.com/RedHatInsights/frontend-components/pull/700 deals with proper build imports.

packages/components/src/Components/FilterHooks/tagFilterHook.js
```diff
@@ -0,0 +1,76 @@
...
+                    label: !state.loaded ?
+                        <span> <Spinner size="md" /> </span> :
+                        <div className="ins-c-tagfilter__no-tags"> No tags available </div>,
+                    isDisabled: true,
+                    className: 'ins-c-tagfilter__tail'
+                }
+            ]
+        }
+    };
+    return [ filter, mapGroups(selectedTags, 'chips'), selectedTags, setValue, filterTagsBy, seFilterTagsBy ];
```
Hyperkid123, line 75: I would probably prefer the return value to be an object. I have a hard time remembering the order of 3 items, let alone 6.

Hyperkid123:
@karelhala just some snapshot goodness.

PR #690, opened 2020-08-12
relative DateFormat: prepend "almost " when rounding up

cben:
Motivation: I want to use DateFormat more in [OpenShift Cluster Manager](https://cloud.redhat.com/openshift).  It's already used there in a couple places (cluster details -> Insights tab). 
But some things change behavior at some round thresholds (3 hours, 60 days) and I'd like wording to somehow change when we cross them.

If it was "2 months" before and is still "2 months" now, it's confusing — why did evaluation period expire??
Differentiating e.g. "almost 2 months" vs "2 months" would help.

### DIFF
--- a/packages/components/src/Components/DateFormat/helper.js
+++ b/packages/components/src/Components/DateFormat/helper.js
@@ -7,14 +7,22 @@ const hour = minute * 60;
 const day = hour * 24;
 const month = day * 30; // let's count that every month has 30 days
 const year = day * 365;
-const formatTime = (number, unit) => `${number} ${number > 1 ? `${unit}s` : unit} ago`;
+const formatTime = (number, unit) => {
+    const rounded = Math.round(number);
+    const unitPlural = rounded > 1 ? `${unit}s` : unit;
+    if (number < rounded) {
+        return `almost ${rounded} ${unitPlural} ago`;
+    } else {
+        return `${rounded} ${unitPlural} ago`;
+    }
+};
 
 const relativeTimeTable = [
-    { rightBound: Infinity, description: date => formatTime(Math.round(date / (year)), 'year') },
-    { rightBound: year, description: date => formatTime(Math.round(date / (month)), 'month') },
-    { rightBound: month, description: date => formatTime(Math.round(date / (day)), 'day') },
-    { rightBound: day, description: date => formatTime(Math.round(date / (hour)), 'hour') },
-    { rightBound: hour, description: date => formatTime(Math.round(date / (minute)), 'minute') },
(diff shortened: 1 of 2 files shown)


cben:
Good point, the doc gives examples like "4 hours ago", "1 day ago" but doesn't say anything about rounding.
@ncameronbritt what do you think about prepending "almost "?

- Another tweak that might(?) be positive for _my specific_ context is always rounding down, but calling 110 min ago "1 hour ago" is likely worse / more surprising than "2 hour ago" :neutral_face:.  
  (If the phrasing was "2 hours have passed", it would be commonly understood it's ≥2, but I don't think "ago" has any implied bias.)

- A related problem is that rounding small integers just loses too much _relative_ precision.  Humans don't think of 90 minutes as neither "1 hour" nor "2 hours".  This could be improved by delaying the switch to bigger units until the number is bigger, for example using up to "59 days" before "2 months"...  However, this would contradict the units by time frames currently recommended in https://www.patternfly.org/v4/design-guidelines/content/numerics#date-and-time.

Anyway the DateFormat component showing exact absolute date on hover is a good mitigation of precision problem, and to some degree of threshold problem too :heart:.  I'm gonna use it either way.

cben:
For reference, relative formatting in some popular frontend libraries:

- https://momentjs.com/docs/#/displaying/fromnow/ — always rounds to closest, and even starts using a unit even before it's whole e.g. 46 secs -> "a minute", 26 days -> "a month".
- https://date-fns.org/v2.15.0/docs/formatDistance — similar, likes to prepend "about ", a few uses of "almost " and "over ".
  https://date-fns.org/v2.15.0/docs/formatDistanceStrict — rounds (customizable).  starts each unit at exactly 1. no fluff, just number & unit.

ncameronbritt:
If we're going to prepend something, I'd prefer "about" as it can imply slightly more or slightly less. "Almost" to my mind implies less, i.e. it's not quite the number. Would we not prepend "about" if it was *exactly* 2 weeks/4 hours/1 month ago? 

In any case, we should be showing the timestamp in a tooltip.

cben:
> "Almost" to my mind implies less, i.e. it's not quite the number.

Yes, my thought was precisely to distinguish slightly-less-then-X from slightly-more-than-X.  "about" would add zero info.

----

"exactly X" vs "about X" would not be a useful distinction.  We're comparing at millisecond precision, so real match chance is negligible.   Also, nothing happens instantenously in the backend systems, and they have slightly different clocks, AND client's browser computing the difference has a different clock!  So this would require defining a range of "approximately exact" ⩰, which I don't think is sane.

Closing as I think the tooltip is a good mitigation of any rounding / ambiguity.

PR #695, opened 2020-08-17
Enhance conditional filter's "See more"

fhlavac:
needs https://github.com/RedHatInsights/frontend-components/pull/702

- added "See less" option and items scrolling
- added "No results" item
- added onShowChange default behavior

### DIFF
--- a/packages/components/src/Components/ConditionalFilter/Group.js
+++ b/packages/components/src/Components/ConditionalFilter/Group.js
@@ -15,7 +15,8 @@ class Group extends Component {
     state = {
         isExpanded: false,
         selected: {},
-        filterBy: ''
+        filterBy: '',
+        isCollapsed: true
     }
 
     onToggle = isExpanded => {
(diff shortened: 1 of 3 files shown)


packages/components/src/Components/ConditionalFilter/Group.js
```diff
@@ -229,21 +234,53 @@ class Group extends Component {
...
-                label: showMoreTitle,
+                label: this.state.isCollapsed ? showMoreTitle : showLessTitle,
                 type: groupType.button,
-                onClick: onShowMore,
-                isPersistentAction: true
+                onClick: onShowChange ? onShowChange : this.toggleCollapsed
             }]
         };
 
+        const { groups: filteredGroups, itemTotal } = groups.reduce((acc, group) => {
```
PanSpagetka, line 265: You are checking `isFilterExpandable` at several places in this reduce, I think you can extract it out.

PR #696, opened 2020-08-19
Reduced title text bottom margin in PDF generator

leSamo:
Fixes [VULN-1193](https://projects.engineering.redhat.com/projects/VULN/issues/VULN-1193).

### Before:
bad

### After:
good

### Mockup:
mockup


packages/pdf-generator/src/utils/styles.js
```diff
@@ -52,6 +52,9 @@ export default (style = {}) => StyleSheet.create({
         fontStyle: 'italic',
         color: '#6e6b6c'
     },
+    reportNameWrapper: {
+        margin: '30, 0, 15'
```
karelhala, line 56: I don't think you can do that in react-css, you gotta write it as
```js
marginTop: 30,
marginRight: 0,
marginBottom: 15
```

leSamo: You actually can, [these docs](https://reactnative.dev/docs/layout-props#margin) link [here](https://developer.mozilla.org/en-US/docs/Web/CSS/margin) for more information, which shows option to provide 3 values for top right bottom; so I'd assume it uses that specification. I have also tested It with exec. reports and it looks correct, for example:

```
reportNameWrapper: {
    margin: '300, 100, 200'
},
```

should produce top 300, left and right 100, botttom 200, and it does:
margins

If you want it to be clearer I can change it to your variant.

karelhala: I was checking docs in react-pdf where they hadn't this option. If it works as expected, then good to go!

karelhala (approved):

PR #700, opened 2020-08-25
Use correct env imports in build.

Hyperkid123:
### Changes
- configured babel to transform dependencies import based on `babel_env` property
  - cjs build bundle will use commojs imports
  - esm build bundle will use esm imports

This should help with the jest issues people are having in some platform apps. I have tested it with the vulnerability UI and if you use cjs imports, the jest issues are fixed and there is no need for transpiling some packages.

In the future, we should avoid directly referencing `cjs`/`esm` build of dependencies choose between them at build time.

There are some snapshot tests I will address and the skeleton size constant is a bit wonky but it should be fixed shortly.

@karelhala do we want to create some beta versions first?


Hyperkid123:
@rvsia go away you are on vacation!

Hyperkid123:
This time the build timed out :no_mouth:

Hyperkid123:
Finally! IT IS :green_circle:

config/rollup-contants.js
```diff
@@ -48,29 +48,34 @@ export const globals = {
     'react-redux': 'reactRedux'
 };
 
-export const rollupConfig = (external, plugins, globals, name, entries = [ globMapper('src/**/index.js') ], outputDir = './') => ([
-    ...[ 'esm', 'cjs' ].map(env => ({
+export const rollupConfig = (external, plugins, globals, name, entries = [ globMapper('src/**/index.js') ], outputDir = './') => {
+    console.warn(JSON.stringify(entries), process.env.FORMAT);
```
karelhala, line 52: Do we really want to print out entries and format?

Hyperkid123: no we don't I used it for debug :+1:

Hyperkid123:
@karelhala I will wait for https://github.com/RedHatInsights/frontend-components/pull/702 to be merged and then rebase it. It should make it easier to review without the gazillion snapshot updates.

karelhala:
Tag has been released, so we have a history point to go back to. Merging this!

karelhala (approved):

Hyperkid123:
@AllenBW let me check. What version are you using? We have made some fixes since this was merged,

AllenBW:
Hiiiii!! @Hyperkid123 just did n update, version 2.6.2 is the breaking one, 2.5.6 works just fine

Hyperkid123:
Alright, let's get into it.

PR #702, opened 2020-08-25
Update PF to v4.40.4

fhlavac:
- updated PF version to 4.40.4
- updated snapshots
- fixed addSourceWizard


karelhala (approved):

packages/components/src/Components/CullingInfo/__snapshots__/CullingInformation.test.js.snap
```diff
@@ -3,35 +3,8 @@
...
-        "right",
-        "bottom",
-      ]
-    }
-    id=""
-    isAppLauncher={false}
-    isContentLeftAligned={false}
-    isVisible={false}
-    maxWidth="18.75rem"
+    content="System scheduled for inventory removal in -18296 days"
```
karelhala, line 6: This should be `content="System scheduled for inventory removal in -18297 days"` because of travis in different timezone.

fhlavac: @karelhala thanks! I missed it in that snapshot storm :)

karelhala (approved):

PR #707, opened 2020-08-28
Fix all styling issues in inventory tagging

karelhala:


rvsia (approved):

packages/components/src/Components/TagModal/TagModal.js
```diff
@@ -203,7 +203,7 @@ TagModal.defaultProps = {
     toggleModal: () => undefined,
     columns: [
         { title: 'Name' },
-        { title: 'Tag Source' }
+        { title: 'Tag source' }
```
Hyperkid123, line 206: Snapshot failed after the case change.

PR #709, opened 2020-08-31
Add footer prop inside PDF generator

leSamo:
Unblocks [VULN-1061](https://projects.engineering.redhat.com/browse/VULN-1061).
[Mockup](https://marvelapp.com/prototype/737a64d/screen/69637228)

footer


packages/pdf-generator/src/components/PDFDocument.js
```diff
@@ -52,19 +53,22 @@ const PDFDocument = ({
                 <View style={appliedStyles.smallSpacing}>
                     {page}
                 </View>
-                <View style={[ appliedStyles.flexRow, {
+                <View style={[{
```
karelhala, line 56: I'd change the styling to use `flex-direction: column` and wrapped the footer in `flexRow` that way someone using this would have easier way of passing multiple elements in the footer and not just one.

leSamo: I was intending for footer prop to be wrapped in view to choose custom bottom margin and such with less layout constraints, like this:

```
footer={
  <View style={{ marginBottom: 20 }}>
    <Text>{intl.formatMessage(messages.customReportFooter)}</Text>
  </View>
}
```
`

karelhala: Sure thing, I was just worried that folks using it won't know that wrapping it in `View` is the right way and wanted it to be easier for them, but that being said consumers have to use `Text` component as well. So good to go!

karelhala (approved):

PR #710, opened 2020-08-31
Change TitleColumn fontsize to match pf default

LI0131:
This PR fixes https://projects.engineering.redhat.com/browse/RHCLOUD-8661. We would like the `FontSize` used in the title column to conform to the patternfly default to maintain uniformity.

### DIFF
--- a/packages/inventory/src/components/table/InventoryList.scss
+++ b/packages/inventory/src/components/table/InventoryList.scss
@@ -84,7 +84,7 @@
         }
 
         * {
-            font-size: var(	--pf-global--FontSize--lg);
+            font-size: var(--pf-global--FontSize--md);
         }
 
         .ins-composed-col {


packages/inventory/src/components/table/InventoryList.scss
```diff
@@ -84,7 +84,7 @@
         }
 
         * {
-            font-size: var(	--pf-global--FontSize--lg);
+            font-size: var(--pf-global--FontSize--md);
```
karelhala, line 87: I'd rather achieve this in app bundle than setting this globally for all apps.
```CSS
table.ins-c-entity-table.pf-c-table .ins-composed-col * {
  font-size: var(--pf-global--FontSize--md);
}
```

This in your SCSS file should do the trick.

karelhala (changes requested):

packages/inventory/src/components/table/InventoryList.scss
LI0131: Ah yes that makes a lot more sense to do it that way. I'll go ahead and close this out then 👍

PR #718, opened 2020-09-01
Fix inventory imports as they need direct imports instead of generic ones

karelhala:


packages/inventory/src/components/detail/AppInfo.js
```diff
@@ -3,7 +3,7 @@ import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { useLocation } from 'react-router-dom';
 import { useStore, useSelector } from 'react-redux';
-import { Skeleton, SkeletonSize } from '@redhat-cloud-services/frontend-components';
+import { Skeleton, SkeletonSize } from '@redhat-cloud-services/frontend-components/components/esm/Skeleton';
```
Hyperkid123, line 6: Can we use CJS imports instead? Ems will be breaking any CJS environments like tests. CJS builds are almost universal.

Hyperkid123 (approved):

rvsia (approved):

PR #719, opened 2020-09-01
Fix start scripts by adding generic configuration

karelhala:
When running `npm start` the build would produce `umd` package only, if app is using any other environment it won't include these changes in final build. This PR fixes it that if no env passed it'll produce one array with all envs.


fhlavac (approved):
Thanks for fixing @karelhala !

fhlavac (approved):

PR #731, opened 2020-09-07
Fire filter changes only after data are loaded

karelhala:


packages/inventory/src/components/table/EntityTableToolbar.js
```diff
@@ -177,31 +177,31 @@ const EntityTableToolbar = ({
     };
 
     useEffect(() => {
-        if (page && perPage && filters && showTags && !hasItems) {
+        if (page && perPage && filters && showTags && !hasItems && loaded) {
```
rvsia, line 180: What about to compute this calue before the hooks and then use the constant to deduplicate and clean this code?

```jsx
const shouldReload = page && perPage && filters && (!hasItems || items) && loaded;

useEffect(() => {
     shouldReload && onSetFilter(...)
}, ...)

useEffect(() => {
     shouldReload && onSetTextFilter(...)
}, ...)

...
```

karelhala: Good point! I'd still use an `if` statement instead of shortcutting so anybody reading this later on would have easier to understand what that means.

packages/inventory/src/components/table/EntityTableToolbar.js
```diff
@@ -176,32 +176,34 @@ const EntityTableToolbar = ({
...
     useEffect(() => {
-        if (page && perPage && filters && (!hasItems || items)) {
+        if (shouldReload) {
             onSetFilter(registeredWithFilter, 'registeredWithFilter', debouncedRefresh);
         }
     }, [ registeredWithFilter ]);
 
     useEffect(() => {
-        if (page && perPage && filters && showTags && !hasItems) {
+        if (shouldReload) {
```
rvsia, line 206: Here you need to use showTags too according to previos revision.

karelhala: Aaah, yes! Good point.

rvsia (approved):

PR #734, opened 2020-09-14
Allow custom filters to be passed from parent

karelhala:
This PR allows consumers to pass custom filters (mainly global filter) to inventory component trough `customFilters`

### DIFF
--- a/packages/inventory/src/components/table/EntityTableToolbar.js
+++ b/packages/inventory/src/components/table/EntityTableToolbar.js
@@ -56,6 +56,7 @@ const EntityTableToolbar = ({
     isLoaded,
     items,
     sortBy,
+    customFilters,
     ...props
 }) => {
     const dispatch = useDispatch();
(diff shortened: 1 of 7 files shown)


packages/inventory/src/components/table/EntityTableToolbar.js
```diff
@@ -92,7 +93,10 @@ const EntityTableToolbar = ({
      */
     const debounceGetAllTags = useCallback(debounce((config, options) => {
         if (showTags && !hasItems) {
-            dispatch(fetchAllTags(config, options));
+            dispatch(fetchAllTags({
+                ...customFilters,
+                config
```
Hyperkid123, line 98: @karelhala here's a question. Do we want the config to have a higher priority? Also, I think the `...config` should be spread as well.

karelhala: Good point with the spread! And yes, we want the config to have higher priority.

Hyperkid123 (approved):

PR #736, opened 2020-09-16
Add information about linking react and react-dom

karelhala:


README.md
```diff
@@ -123,9 +123,14 @@ To test changes from packages in this repository in other applications follow th
   rm -rf node_modules/react; rm -rf node_modules/react-dom
   ```
   This is because we want to use hooks and different reacts are not playing nicely with hooks [facebook/react/issues/15315](https://github.com/facebook/react/issues/15315)
+3. Link `react` and `react-dom` from your application.
```
rvsia, line 126: Put a space before this line, otherwise there will be no new line,

README.md
```diff
@@ -123,9 +123,14 @@ To test changes from packages in this repository in other applications follow th
...
+```
+ln -s $PWD/insights-inventory-frontend/node_modules/react frontend-components/node_modules/react
```
rvsia, line 128: Maybe I would use some placeholder` $PWD/<application-folder>/...`

bastilian: +1 and possible use something like `/path/to/frontend-components/...` for the target to not confuse people thinking it is implied that it is within the apps directory. I don't think anyone would think that, but to be sure. :)

bastilian:
Thank you for adding more information @karelhala!

Following the steps solves the issues with React'n'Hooks! \o/

One thing that may be confusing might be where to run each command,
so it might make sense to be very explicit in each step with `run from the frontend-components directory/application`.

karelhala:
@bastilian good point! I've added a note from where to run the `ln -s` command, should be more clear now.

bastilian (approved):
Awesome! Thank you @karelhala!

rvsia (approved):

PR #742, opened 2020-09-17
Idea for running locally

josejulio:
I'm providing another idea for running locally, could probably be improved by adding yalc as dev dependency and including it on the `npm run start` to avoid having to run `yalc push` manually.


README.md
```diff
@@ -117,27 +117,16 @@ These tasks are preconfigured
...
-
-3. Link `react` and `react-dom` from your application. Running from folder that contains your application and frontend components. Running `ls` in this folder would yield `<application-folder>  insights-proxy  frontend-components`
-```
-ln -s $PWD/<application-folder>/node_modules/react frontend-components/node_modules/react
-ln -s $PWD/<application-folder>/node_modules/react-dom frontend-components/node_modules/react-dom
-```
-
-4. Change into the directory of the package you are working on, for example `cd packages/components` and run `npm link`*
-5. Change into the directory of the application you'd like to include the package and run `npm link @redhat-cloud-services/frontend-components`*
+1. Install [yalc](https://github.com/whitecolor/yalc) globally. e.g. `npm install -g yalc`.
```
karelhala, line 120: Can you make this as seperate section? I like this idea, let's just allow consumers to choose which one they prefer.

README.md
```diff
@@ -117,27 +117,16 @@ These tasks are preconfigured
...
+2. Run `npm install` in the root of the `frontend-components` working copy.
+3. Change into the directory of the package you are working on, for example `cd packages/components` and run `yalc publish`*
```
bastilian, line 122: The `yalc publish` documentation mentions it will "copy" assets that would normally be pushed, does this mean that `publish` needs to be run anytime assets are rebuilt when running `rpm start`?

README.md
josejulio: Sure, I'll put into a separate section.

README.md
josejulio: That's correct, it doesn't have a built-in `watch mode`. You can use something like `watch -n 0.5 yalc publish --push --changed` to push assets only when there are changes.

karelhala: Can you include that in the docs as well? I like this solution!

josejulio: Yes, I'll work on it now.

josejulio: Done

karelhala (approved):

PR #744, opened 2020-09-22
[RHICOMPL-732] Values for radio filters aren't arrays

bastilian:
When enabling `hidePassed` the filter value needs to be a single one and not an array.


akofink (approved):
Thanks, @bastilian!

PR #749, opened 2020-09-24
[RHICOMPL-859][RHICOMPL-955] Improve compliance policy card labels and tooltip

vkrizan:
Improved Policy Card to match the UI mocks:
* show number failed rules only
* mention Score
* tooltip about scoring moved to the score and shown on the bottom of  the text
* corrected "Noncompliant" label to "Not compliant". 

Screenshot from 2020-09-24 18-17-35

### DIFF
--- a/packages/inventory-compliance/src/SystemPolicyCard.js
+++ b/packages/inventory-compliance/src/SystemPolicyCard.js
@@ -29,7 +29,7 @@ class SystemPolicyCard extends React.Component {
                 <CheckCircleIcon /> Compliant
             </div> :
             <div className='ins-c-policy-card ins-m-noncompliant'>
-                <ExclamationCircleIcon/> Noncompliant
+                <ExclamationCircleIcon/> Not compliant
             </div>;
     }
 
--- a/packages/inventory-compliance/src/SystemPolicyCard.js
+++ b/packages/inventory-compliance/src/SystemPolicyCard.js
@@ -54,7 +54,7 @@ class SystemPolicyCard extends React.Component {
     }
 
     render() {
-        const { policy, benchmark, rulesPassed, rulesFailed, compliant, lastScanned, score } = this.state.policy;
+        const { policy, benchmark, rulesFailed, compliant, lastScanned, score } = this.state.policy;
         const { refIdTruncated, cardTitle } = this.state;
         const passedPercentage = this.fixedPercentage(score);
         const FormattedRelativeCmp = FormattedRelativeTime || FormattedRelative || Fragment;
@@ -81,15 +81,20 @@ class SystemPolicyCard extends React.Component {
                         </Tooltip> }
                     </TextContent>
                     <div className='margin-bottom-md' >
-                        <Tooltip content={
-                            'The system compliance score is calculated by OpenSCAP and ' +
-                            'is a normalized weighted sum of rules selected for this policy.'
-                        }>
-                            { this.complianceIcon(compliant) }
-                            <Text component={ TextVariants.small }>
-                                { rulesPassed } of { rulesPassed + rulesFailed } rules passed ({ passedPercentage })
-                            </Text>
-                        </Tooltip>
+                        { this.complianceIcon(compliant) }
+                        <Text component={ TextVariants.small }>
+                            { rulesFailed } rule{ rulesFailed === 1 ? '' : 's' } failed
+                            {' '}
+                            <Tooltip
+                                position='bottom'
+                                maxWidth='22em'
+                                content={
+                                    'The system compliance score is calculated by OpenSCAP and ' +
+                                    'is a normalized weighted sum of rules selected for this policy.'
+                                }>
+                                <span>(Score: { passedPercentage })</span>
+                            </Tooltip>
+                        </Text>
                     </div>
                     <div className='margin-bottom-md' >
                         <Text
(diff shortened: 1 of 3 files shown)


packages/inventory-compliance/src/SystemPolicyCard.js
```diff
@@ -81,15 +81,19 @@ class SystemPolicyCard extends React.Component {
...
-                            'is a normalized weighted sum of rules selected for this policy.'
-                        }>
-                            { this.complianceIcon(compliant) }
-                            <Text component={ TextVariants.small }>
-                                { rulesPassed } of { rulesPassed + rulesFailed } rules passed ({ passedPercentage })
-                            </Text>
-                        </Tooltip>
+                        { this.complianceIcon(compliant) }
+                        <Text component={ TextVariants.small }>
+                            { rulesFailed } rule{ rulesFailed == 1 ? '' : 's' } failed
```
bastilian, line 86: You could write this as `{ rulesFailed != 1 && 's' }`

packages/components/src/Components/CullingInfo/__snapshots__/CullingInformation.test.js.snap
```diff
@@ -3,7 +3,7 @@
 exports[`CullingInformation it should render correctly without any date 1`] = `
 <Fragment>
   <Tooltip
-    content="System scheduled for inventory removal in -18297 days"
+    content="System scheduled for inventory removal in -18296 days"
```
bastilian, line 6: This will probably cause TravisCI to fail. Better not commit this.

packages/inventory-compliance/src/SystemPolicyCard.js
```diff
@@ -54,7 +54,7 @@ class SystemPolicyCard extends React.Component {
     }
 
     render() {
-        const { policy, benchmark, rulesPassed, rulesFailed, compliant, lastScanned, score } = this.state.policy;
+        const { policy, benchmark, _, rulesFailed, compliant, lastScanned, score } = this.state.policy;
```
bastilian, line 57: Why the `_`? If it's not used don't assign it. ;)

vkrizan: Heh, I thought the policy is an array. I still have to get use to this React-fu. :-)

packages/inventory-compliance/src/SystemPolicyCard.js
vkrizan: Indeed. Next time. :wink:

packages/inventory-compliance/src/SystemPolicyCard.js
```diff
@@ -81,15 +81,19 @@ class SystemPolicyCard extends React.Component {
...
-                        }>
-                            { this.complianceIcon(compliant) }
-                            <Text component={ TextVariants.small }>
-                                { rulesPassed } of { rulesPassed + rulesFailed } rules passed ({ passedPercentage })
-                            </Text>
-                        </Tooltip>
+                        { this.complianceIcon(compliant) }
+                        <Text component={ TextVariants.small }>
+                            { rulesFailed } rule{ rulesFailed === 1 ? '' : 's' } failed
+                            { " " }
```
bastilian, line 87: This is not needed. Just use a space within the `span` It should be preserved and not ignored when rendering.

vkrizan: If you mean `<span> </span>`, yes. I wanted to avoid adding the space into the span that has the tooltip. `{ " " }` was one of the (shorter) options mentioned on forums.

bastilian: Shorter than a ` `?

bastilian: It doesn't hurt if it is also triggered on that space. Makes the target bigger. ;)

vkrizan: Here are the options:
* `{ " " }`
* `<span> </span>`
* `&nbsp;` (not sure if we want to have a non-breaking space)

A space in the span down below in the tooltip is not what I wanted to do. It changes the tooltip positioning as well as the trigger (if we want to be precise).

packages/inventory-compliance/src/SystemPolicyCard.js
```diff
@@ -81,15 +81,19 @@ class SystemPolicyCard extends React.Component {
...
+                            <Tooltip
+                                position='bottom'
```
bastilian, line 89: The `maxWidth` should be set to make the content fit into 3 lines and not have `policy` in a single line as mentioned in the Jira

bastilian (changes requested):

packages/inventory-compliance/src/SystemPolicyCard.js
vkrizan: Oh, I missed that.

vkrizan: Done.

packages/inventory-compliance/src/SystemPolicyCard.js
bastilian: I'd prefer just adding a space to the span at the bottom.
It doesn't need to be precise. And it only marginally, about 1,5px, skews the position.

Behaviour around preserving space in JSX has been changing over time so I'd avoid it.

bastilian (approved):
LGTM! Thank you @vkrizan!

packages/inventory-compliance/src/SystemPolicyCard.js
vkrizan: I would not worry about the bracket notation with whitespace. If JSX would break it, then majority of notations, not related to whitespace, will break with it.

PR #752, opened 2020-09-25
Add NODE_ENV=production to rollup build.

Hyperkid123:
This removes the annoying redux warning that we use a production bundle outside of the production environment. Rollup does not take the variable from CLI so we have to use replace plugin to manually set it in the build process.

### DIFF
--- a/config/rollup-plugins.js
+++ b/config/rollup-plugins.js
@@ -1,0 +1,9 @@
+import replace from 'rollup-plugin-replace';
+
+const rollupPlugins = [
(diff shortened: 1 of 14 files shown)


karelhala (approved):

Hyperkid123:
@karelhala any idea why the tests fail? This just adds a variable to build and does not modify any package or component.

Hyperkid123:
nvm i know I have removed the prepare install command for testing...

PR #753, opened 2020-09-25
Add tooltip to all tag filter options

karelhala:
### Tooltip in tags filter

Since tags can be long we want to wrap them in tooltip so users can hover over them and see full tag.

### DIFF
--- a/packages/components/src/Components/FilterHooks/constants.js
+++ b/packages/components/src/Components/FilterHooks/constants.js
@@ -49,35 +49,38 @@ export function constructGroups(allTags, item = 'item') {
         label: name,
         value: name,
         type,
-        items: tags.map(({ count, tag: { key: tagKey, value } }) => ({
-            className: 'ins-c-tagfilter__option',
-            label: <React.Fragment>
-                <Tooltip
-                    content={`${tagKey}${value ? `=${value}` : ''}`}
-                >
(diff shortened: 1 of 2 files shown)


packages/components/src/Components/FilterHooks/constants.js
```diff
@@ -52,7 +52,11 @@ export function constructGroups(allTags, item = 'item') {
         items: tags.map(({ count, tag: { key: tagKey, value } }) => ({
             className: 'ins-c-tagfilter__option',
             label: <React.Fragment>
-                <div className="ins-c-tagfilter__option-value">{tagKey}{value ? `=${value}` : ''}</div>
+                <Tooltip
+                    content={`${tagKey}${value ? `=${value}` : ''}`}
```
rvsia, line 56: just a nitpick: this string could be calculated in a variable and be used at both places

rvsia (approved):

packages/components/src/Components/FilterHooks/constants.js
karelhala: Yeah, good point. I was thinking about that as well, but didn't wanted to change this to function body. But it might be better to use constant, just for sanity and future improvements.

Hyperkid123 (approved):

PR #756, opened 2020-09-30
Remove reference to apply reducer hash in notifications.

Hyperkid123:
part of: https://issues.redhat.com/browse/RHCLOUD-6078

The inner reference to utils redux helper functions cause the `you are currently using minified code outside of node_env === 'production'`. I was able to successfully replicate this issue and fix it inside the landing page frontend app.

Using the redux utils functions directly in apps **does not** cause the error to show up!

This will be a problem with Lerna hoisted dependencies BUT I do not yet know how to fix that. Because the notifications package was only using the apply reducer hash function, we can simply replace it by box standard switch with no regressions. I will try and find why and how to fix that but I don't have any idea other than migrating to yarn workspaces :smile: .


Hyperkid123:
I have been digging a bit and turns out, there is a `lerna link` feature which will create a symlink to packages, however that breaks the whole repository and it won't compile. So that would be probably the solution in the future, but we would probably have to make some changes to our build process first.

karelhala:
I would much rather try to solve this universally for all packages, notifications is pulling single dependency from utils package, however some other packages are pulling a lot of dependencies. I think that this will be partially solved by npm v 7 and how links and workspaces will work.

Hyperkid123:
@karelhala I would like to treat this more like a hotfix. I want to revisit the Lerna link and see where the issue might be. I just want to get rid of that error.

PR #757, opened 2020-09-30
[RHICOMPL-993] Use remediationIssueId from backend

vkrizan:
to avoid leaking the id generation outside of backend.


vkrizan:
Closing, as this would could/would be solved differently.

PR #758, opened 2020-10-01
fix(RHICOMPL-880): Update Ansible icon to use patternfly version

bastilian:


packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -14,6 +14,15 @@ import {
 import { HIGH_SEVERITY, MEDIUM_SEVERITY, LOW_SEVERITY } from './Constants';
 import ComplianceRemediationButton from './ComplianceRemediationButton';
 
+export const columns = [
+    { title: 'Rule', transforms: [ sortable ] },
+    { title: 'Policy', transforms: [ sortable ] },
+    { title: 'Severity', transforms: [ sortable ] },
+    { title: 'Passed', transforms: [ sortable ] },
+    { title: <React.Fragment><Ansible size='lg'/> Ansible</React.Fragment>,
```
vkrizan, line 22: The icon is too big compared to the text. Try medium, or maybe the default would inherit font-size.

Screenshot_2020-10-02 Compliance

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import propTypes from 'prop-types';
 import { Pagination, PaginationVariant, ToolbarItem } from '@patternfly/react-core';
-import { CheckCircleIcon, ExclamationCircleIcon } from '@patternfly/react-icons';
-import { Table, TableHeader, TableBody } from '@patternfly/react-table';
+import { CheckCircleIcon, ExclamationCircleIcon, AnsibeTowerIcon as Ansible } from '@patternfly/react-icons';
```
vkrizan, line 4: What's the rationale on renaming the icon name?

vkrizan (changes requested):

packages/inventory-compliance/src/SystemRulesTable.js
bastilian: ... evening musings. :D I removed the alias.

packages/inventory-compliance/src/SystemRulesTable.js
bastilian: Changed to 'sm' that should fit better.

vkrizan: I guess the medium would fit better, but I'll leave it to your discretion.

vkrizan (approved):

PR #761, opened 2020-10-06
fix(RHICOMPL-1005): Address issues with more spaces in filter labels

bastilian:
Fixes the issues mentioned in https://github.com/RedHatInsights/compliance-frontend/blob/master/src/PresentationalComponents/ReportsTable/Filters.js#L45

### DIFF
--- a/packages/inventory-compliance/src/Utilities/FilterConfigBuilder.js
+++ b/packages/inventory-compliance/src/Utilities/FilterConfigBuilder.js
@@ -3,7 +3,7 @@ import FilterBuilder from './FilterBuilder';
 import ChipBuilder from './ChipBuilder';
 
 export const stringToId = (string) => (
-    string.replace(' ', '').toLowerCase()
+    string.split(' ').join('').toLowerCase()
 );
 
 class FilterConfigBuilder {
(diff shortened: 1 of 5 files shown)


packages/inventory-compliance/src/Utilities/FilterConfigBuilder.js
```diff
@@ -3,7 +3,7 @@ import FilterBuilder from './FilterBuilder';
 import ChipBuilder from './ChipBuilder';
 
 export const stringToId = (string) => (
-    string.replace(' ', '').toLowerCase()
+    string.split(' ').join('').toLowerCase()
```
vkrizan, line 6: Nitpick: `.replace(/ /g, '')`

vkrizan (approved):
Works!

packages/inventory-compliance/src/Utilities/FilterConfigBuilder.js
bastilian: Why?

vkrizan: Does the same with one call.

PR #762, opened 2020-10-06
Fire refresh data on load if inventory has items

karelhala:
### Fix infinit load on systems pages

There is a problem when app passes `items` to inventory the table wouldn't load for some cases, this can happen when app waits for `items` to be loaded and passes them to inventory at the same time as inventory first renders. We don't want to fire the first fetch at all if app didn't defined `items` it means that the app is using full inventory and first fetch is fired from `InventoryTableToolbar`.


jiridostal (approved):
Makes sense to me!

PR #763, opened 2020-10-06
Fix conditional toolbar groups component by adding value to empty fields

karelhala:
### Error while selecting tags in global filter

With new addition to tags filter and option to scroll the filter broke, because PF expects `value` on each children, and when spinner is shown no such thing is passed in any component. This PR fixes such issue by adding empty values to parent component and button to manage all tags.

I've also added option to run start for only JS, UMD or ESM modules.

### DIFF
--- a/packages/components/src/Components/ConditionalFilter/Group.js
+++ b/packages/components/src/Components/ConditionalFilter/Group.js
@@ -253,7 +253,7 @@ class Group extends Component {
                 { ...(isFilterable || onFilter) && { onFilter: this.customFilter } }
                 { ...groups && groups.length > 0 && { isGrouped: true }}
             >
-                <div className="ins-c-select__scrollable-section">
+                <div className="ins-c-select__scrollable-section" value="">
                     { groups && groups.length > 0 ? (
                         groups.map(({
                             value: groupValue,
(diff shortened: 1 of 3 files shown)


packages/components/src/Components/ConditionalFilter/Group.js
```diff
@@ -285,11 +285,12 @@ class Group extends Component {
                             className="pf-c-select__menu-item"
                             variant={showMore.variant}
                             onClick={showMore.items[0].onClick}
+                            value="Show more"
                         >
                             {showMore.items[0].label}
                         </Button>
                     </SelectGroup>
-                    : <Fragment />}
+                    : <Fragment value="" />}
```
rvsia, line 293: When we are here we can do this

```jsx
onShowMore && <SelectGroup>
                        <Button
                            {...showMore.items[0]}
                            className="pf-c-select__menu-item"
                            variant={showMore.variant}
                            onClick={showMore.items[0].onClick}
                            value="Show more"
                        >
                            {showMore.items[0].label}
                        </Button>
                    </SelectGroup>
```

to make it less ugly. :smile:

karelhala: Not really, because PF complains that `React.children([false]).map(child => child.props.value.toString() === 'something')` (`false` is calculated from this expression) is not working :disappointed:

rvsia: :boom:

rvsia (approved):

PR #765, opened 2020-10-06
fix(RHICOMPL-880): Set fitContent for columns to not trigger tooltip

bastilian:
When a column header contains html it will render this as a string for the tooltip.
Setting `props: { tooltip '...' }` should allow to set an alternative, according to the Patternfly (code), but this appears broken.

Using `fitContent` will avoid the issue by avoiding to show a tooltip. 

This also adds a function and exports it to select certain columns instead of redefine them everywhere[1].

[1] https://github.com/RedHatInsights/compliance-frontend/pull/788


akofink (approved):
Looks good, thanks!

PR #768, opened 2020-10-08
Allow passing custom filter for systems table

karelhala:
### Tags fixtures

When switching between pages in tagsModal the width changes, this PR fixes such issue by passing `medium` variant to this component.

### Inotroduce filter query param

We have new feature in inventory to allow custom filters being passed, however our JS client generator is not working correctly and is not properly mapping `type: object` to proper shape, we have to walk trough the object and map paths as string of `[]`.

### DIFF
--- a/packages/utils/src/helpers.js
+++ b/packages/utils/src/helpers.js
@@ -82,3 +82,19 @@ export function getBaseName(pathname, level = 2) {
         return `${acc}${pathName[key] || ''}${key < (level - 1) ? '/' : ''}`;
     }, release);
 }
+
+export const generateFilter = (data, path = 'filter') =>
+    Object.entries(data || {}).reduce((acc, [ key, value ]) => {
+        const newPath = `${path || ''}[${key}]`;
+        if (value instanceof Function || value instanceof Date) {
+            return acc;
+        }
+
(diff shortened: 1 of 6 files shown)


packages/utils/src/helpers.js
```diff
@@ -82,3 +82,19 @@ export function getBaseName(pathname, level = 2) {
...
+        }
+
+        return {
+            ...acc,
+            ...(Array.isArray(value) || typeof value !== 'object')
+                ? { [newPath]: value } :
+                generateFilter(value, newPath)
+        };
+
+    }, {});
```
rvsia, line 100: You added this function 2x. Is it intentional?

rvsia (commented):
I would add tests for `generateFilter` to test it and document what it does as it is rather a complex function.

packages/utils/src/helpers.js
karelhala: Yup, intentional as a midterm sollution (I don't want to break the UI because of mismatch version of `utils` package), we can always remove it later down the line.

rvsia (approved):
Thanks for considering my feedback! Great job! :trophy:

PR #769, opened 2020-10-08
GeneralInfo component should get writePermissions

rvsia:
See https://github.com/RedHatInsights/insights-inventory-frontend/pull/817

This PR makes **GeneralInfo** to accept `writePermissions` prop.

1. `writePermissions === false`

Edit buttons are **hidden**

2. `writePermissions === true` / `isProd === true`

Edit buttons are **shown**

3. `writePermissions === undefined`

Edit button will use **rbac hook** to decide if it's shown or not. This result is being **cached** to eliminate pop-in when switching tabs.

### DIFF
--- a/packages/inventory-general-info/src/EditButton.js
+++ b/packages/inventory-general-info/src/EditButton.js
@@ -5,33 +5,60 @@ import { usePermissions } from '@redhat-cloud-services/frontend-components-utili
 
 import { PencilAltIcon } from '@patternfly/react-icons';
 
-const EditButton = ({ link, onClick }) => {
+const InnerButton = ({ link, onClick }) => (
+    <a
+        className="ins-c-inventory__detail--action"
+        href={ `${window.location.href}/${link}` }
+        onClick={ onClick }
+    >
+        <PencilAltIcon />
+    </a>
+);
+
+InnerButton.propTypes = {
+    link: PropTypes.string.isRequired,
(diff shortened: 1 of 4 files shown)


packages/inventory-general-info/src/EditButton.js
```diff
@@ -5,33 +5,60 @@ import { usePermissions } from '@redhat-cloud-services/frontend-components-utili
...
         'inventory:*:write'
     ]);
 
-    const canPerformActions = insights.chrome.isProd || hasAccess;
+    if (hasAccess) {
+        permissionsCache = hasAccess;
+    }
 
-    if (!canPerformActions) {
+    if (!hasAccess) {
```
karelhala, line 36: By default the `hasAccess` will have `undefined` might want to check `isLoading` from `useRbacHook` as well and perhaps show some loader/spinner.

karelhala (approved):
Just one minor question, but overall looks good!

packages/inventory-general-info/src/EditButton.js
rvsia: I hope the loading would be too fast to show anything. :smiley: 

Otherwise, all applications should probably set up the `writePermissions` prop, this should serve only as a fallback (probably it would be best to make the `writePermissions` required but right now I am not sure how many apps are using this component and how hard would be to get all updated).

PR #770, opened 2020-10-08
Clean tests for inventory general info

rvsia:
There were many warnings/errors in jest log

- added whatwg-fetch to simulate fetch
- fixed tests data to correspond with real data (numbers > string)
- added aria-labels where missing
- value in filterConfig has to be a string according to primarytoolbar

**After**

Screenshot from 2020-10-08 14-52-30

### DIFF
--- a/config/globalSetup.js
+++ b/config/globalSetup.js
@@ -1,0 +1,3 @@
+module.exports = async () => {
+    process.env.TZ = 'UTC';
+};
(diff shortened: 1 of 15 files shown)


karelhala (approved):

rvsia:
@karelhala Because of issues with timezones I added a global setup to make sure that it will show the same dates locally/on CI. So I re-requested your review, because it affects all other tests too.

karelhala (approved):

karelhala:
@rvsia yeah, good point! This was bugging me for some time, but hadn't had much time to investigate it. Thank you!

rvsia:
@karelhala oohhhh it works! :fireworks: :fireworks:

PR #772, opened 2020-10-09
Show currently selected tags above tags table

karelhala:
### Show currently selected tags

The selection of tags can be a bit confusing and when users switch page they can loose focus on what has been selected. We don't want to group these tags to chipGroups based on the source as this is only informative.

animation

### DIFF
--- a/packages/components/src/Components/TagModal/TagModal.js
+++ b/packages/components/src/Components/TagModal/TagModal.js
@@ -9,7 +9,9 @@ import {
     EmptyStateVariant,
     Title,
     EmptyStateBody,
-    Button
+    Button,
+    ChipGroup,
+    Chip
 } from '@patternfly/react-core';
 import classNames from 'classnames';
 import {
@@ -63,7 +65,6 @@ export default class TagModal extends React.Component {
             primaryToolbarProps,
             ...props
         } = this.props;
-
         return (
             <Modal
                 {...props}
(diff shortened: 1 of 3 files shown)


karelhala:
ping @terezanovotna what do you think about it?

packages/inventory/src/shared/TagsModal.js
```diff
@@ -69,7 +69,7 @@ class TagsModal extends React.Component {
                     }))
                 }}
                 loaded={ loaded }
-                width="auto"
+                width={'50%'}
```
rvsia, line 72: no need to use block

karelhala: Yeah! I've copied that from PF and they have it block, dumb...

packages/components/src/Components/TagModal/TagModal.js
```diff
@@ -86,6 +87,15 @@ export default class TagModal extends React.Component {
                     ]
                 }}
             >
+                <ChipGroup>
+                    {selected?.map(({ id, key }) => <Chip
```
rvsia, line 91: just a nitpick: maybe we could refactor this to 
```jsx
{selected.length > 0 &&  <ChipGroup>
     {selected.map( ....
```

to not show ChipGroup when no chip is selected.

rvsia (approved):
Code looks solid! :t-rex: 

Just one small optional suggestion.

packages/components/src/Components/TagModal/TagModal.js
karelhala: `selected?.map` should do the trick

Hyperkid123 (approved):
I'll wait for @terezanovotna before merging.

PanSpagetka (approved):

terezanovotna:
Nice!

There was no direct guidance on this, but could we move the chips above the Title?
It would look like this: 

Selected tags - chip, chip, chip
Title - All tags

checkout this example from RBAC: https://github.com/patternfly/patternfly-design/issues/918

I will put together guidelines to keep it consistent.

karelhala:
@terezanovotna The difference is that in RBAC it is used in wizard and here it is used in modal. However with SIDs in modal (planned to be implemented this week) we'll have to change the modal title to something else `Apply multiple tags/SIDs`(?) and have tags and SIDs in tabs. Then it would be reasonable to place these chips above tabs.

terezanovotna:
Screen Shot 2020-10-12 at 11 58 33

karelhala:
Closing due to inactivity

PR #773, opened 2020-10-12
Add [] for array values

karelhala:


PanSpagetka (approved):

Hyperkid123:
@karelhala This can be potentially a breaking change. Is there a way that we can prevent possible issues with URL formats? Maybe add this as an optional configuration option and log deprecation warning?

karelhala:
@Hyperkid123 this has been added 3 days ago, I don't think anybody started to use this :smile:

PR #776, opened 2020-10-13
Fixed missing aria label on text search input.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-6077

### Changes
- added default prop 'aria-label' to text

### DIFF
--- a/packages/components/src/Components/ConditionalFilter/ConditionalFilter.js
+++ b/packages/components/src/Components/ConditionalFilter/ConditionalFilter.js
@@ -24,7 +24,7 @@ class ConditionalFilter extends Component {
     }
 
     render() {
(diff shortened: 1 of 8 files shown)


karelhala (approved):

Hyperkid123:
@karelhala oh I have not even noticed that. I just run npm install as usual. Let me try again.

Hyperkid123:
@karelhala I have reinstalled dependencies and the lock file change is still there. Possibly it may be just my version of node and npm. Or we forgot it to add it in some previous PR. Can you check locally if you have these after running the install on master?

Hyperkid123:
@karelhala I have restored the lock file to master. We can deal with it later.

PR #777, opened 2020-10-13
Allow multiple entities to be passed into tags modal

karelhala:
### Enable multiple tabs in tags modal

Since we have tags and SIDs fro users as single global filter, we should also allow them to see all SIDs and tags on one place. By default only 10 SIDs is visible so if user has more then these 10 items they could be hidden, introduce tabbed tags modal in order to show all tags and SIDs on one place. This is achieved by passing array of `rows`, `columns`, `paginations`, `loaded`, `filters`, `selected`, `onSelect` and `onUpdateData`. This means that consumers willing to show more than one table will have to pass array of these props matching tabs and it will show tables with corresponding data. If they pass something bad in there, for instance wrong rows it will break, but that's to be expected.

### Screenshots
Screenshot from 2020-10-14 14-32-57
Screenshot from 2020-10-14 14-33-03

### DIFF
--- a/packages/components/src/Components/TagModal/TableWithFilter.js
+++ b/packages/components/src/Components/TagModal/TableWithFilter.js
@@ -37,7 +37,7 @@ const TableWithFilter = ({
     entityName
 }) => {
     const onRowSelect = ({ isSelected, rowId }) => {
-        const currRow = rows[rowId];
+        const currRow = rows?.[rowId];
         if (currRow) {
             onSelect(isSelected ? [ ...selected, currRow ] : selected.filter(({ id }) => id !== currRow.id));
         }
@@ -50,7 +50,6 @@ const TableWithFilter = ({
                     bulkSelect: {
                         count: selected?.length,
                         onSelect: (isSelected) => {
-                            console.log('mhhh');
                             if (isSelected) {
                                 onSelect(unique?.([ ...rows, ...selected ]));
                             } else {
@@ -63,7 +62,7 @@ const TableWithFilter = ({
                             onClick: () => onSelect([])
                         },
                         {
-                            ...loaded && rows && rows.length > 0 ? {
+                            ...loaded && rows?.length > 0 ? {
                                 title: `Select page (${ rows.length })`,
                                 onClick: () => onSelect(unique([ ...rows, ...selected ]))
                             } : {}
--- a/packages/components/src/Components/TagModal/TableWithFilter.js
+++ b/packages/components/src/Components/TagModal/TableWithFilter.js
@@ -48,15 +48,16 @@ const TableWithFilter = ({
             {onUpdateData && <PrimaryToolbar
                 {...onSelect && pagination && {
                     bulkSelect: {
-                        count: selected.length,
+                        count: selected?.length,
                         onSelect: (isSelected) => {
+                            console.log('mhhh');
                             if (isSelected) {
-                                onSelect(unique([ ...rows, ...selected ]));
+                                onSelect(unique?.([ ...rows, ...selected ]));
                             } else {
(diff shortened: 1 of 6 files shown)


packages/components/src/Components/TagModal/TableWithFilter.js
```diff
@@ -0,0 +1,162 @@
...
+    primaryToolbarProps,
+    children,
+    title,
+    systemName,
+    columns,
+    tableProps,
+    entityName
+}) => {
+    const onRowSelect = ({ isSelected, rowId }) => {
+        const currRow = rows[rowId];
```
Hyperkid123, line 40: Rows is not required nor it has a default value. Potential runtime crash.

packages/components/src/Components/TagModal/TableWithFilter.js
```diff
@@ -0,0 +1,162 @@
...
+    };
+
+    return (
+        <Fragment>
+            {onUpdateData && <PrimaryToolbar
+                {...onSelect && pagination && {
+                    bulkSelect: {
+                        count: selected?.length,
+                        onSelect: (isSelected) => {
+                            console.log('mhhh');
```
Hyperkid123, line 53: console

packages/components/src/Components/TagModal/TagModal.js
```diff
@@ -78,103 +85,59 @@ export default class TagModal extends React.Component {
...
-                                } else {
-                                    onSelect(selected.filter(({ id }) => !rows.find(({ id: rowId }) => rowId === id)));
-                                }
-                            },
-                            checked: loaded && calculateChecked(rows, selected),
-                            items: [{
-                                title: 'Select none (0)',
-                                onClick: () => onSelect([])
-                            },
+                {Array.isArray(tabNames) ?
```
Hyperkid123, line 96: This can be stored in a variable. There are multiple calls with the same argument.

packages/components/src/Components/TagModal/TagModal.js
```diff
@@ -78,103 +85,59 @@ export default class TagModal extends React.Component {
...
-                        itemCount={pagination.count}
-                        perPage={pagination.perPage}
-                        page={pagination.page}
-                        variant="bottom"
-                        onSetPage={(_event, page) => onUpdateData({ ...pagination, page })}
-                        onPerPageSelect={(_event, perPage) => onUpdateData({ ...pagination, page: 1, perPage })}
-                    />
-                </TableToolbar> }
+                        </Tabs>
+                    </Fragment> :
```
Hyperkid123, line 122: The fragment is unnecessary.

packages/components/src/Components/TagModal/TableWithFilter.js
rvsia, line 53: console.log

packages/components/src/Components/TagModal/TableWithFilter.js
```diff
@@ -0,0 +1,162 @@
...
+                                onSelect(selected.filter(({ id }) => !rows.find(({ id: rowId }) => rowId === id)));
+                            }
+                        },
+                        checked: loaded && calculateChecked?.(rows, selected),
+                        items: [{
+                            title: 'Select none (0)',
+                            onClick: () => onSelect([])
+                        },
+                        {
+                            ...loaded && rows && rows.length > 0 ? {
```
rvsia, line 66: can be shortened to rows?.length > 0

packages/components/src/Components/TagModal/TableWithFilter.js
karelhala: Should be fixed

rvsia (approved):

PR #779, opened 2020-10-14
Disable bulk select in inventory if isDisabled passed in bulk props

karelhala:
### Do not shadow `isDisabled`

Since inventory consumers might decide to disable bulk select on their own do not shadow that property and rather use the one passed in from consumer app and only if that is either undefined or false use `hasAccess`.


johnsonm325 (approved):
LGTM! Thank you.

rvsia (approved):

PR #782, opened 2020-10-15
Fix showing of bulk select all the time in inventory

karelhala:


packages/inventory/src/components/table/EntityTableToolbar.js
```diff
@@ -296,9 +296,11 @@ const EntityTableToolbar = ({
     return <Fragment>
         <PrimaryToolbar
             {...props}
-            bulkSelect={{
-                ...bulkSelect,
-                isDisabled: bulkSelect?.isDisabled || !hasAccess
+            {...bulkSelect && {
```
rvsia, line 299: why is there a spread operator? :thinking:

karelhala: So we don't populate the `bulkSelect` at all if no value is passed. Without it this component could have send `bulkSelect={}` or (which is the point of this PR) `bulkSelect={{ isDisabled: false }}`.

karelhala: You can actually see in the failing tests that previsouly it was passing `bulkSelect` with disabled = false and that was causing this exact error.

rvsia: Oh.. I am wrong the whole time, I thought it was `(...bulkSelect) && {}` :facepalm: :facepalm: 

I am used to write these conditions as `...(bulkSelect && {})`

Sorry!!! :service_dog:

rvsia (approved):
snapshots :skull_and_crossbones:

PR #784, opened 2020-10-19
Pdf generator without clicking

katerinapat:
The goal is to provide a way to render `DownloadButton` and generate the PDF by having another event handler other than clicking the button 

For example, this can be used for modal, the action button inside `<Modal/>` will update the state of the app and render the `DownloadButton` ouside of it .

An example for /demo. 

```
import React, { Component, Fragment } from 'react';
import ReactDOM from 'react-dom';
import { DownloadButton, Section, Column, Table, Panel, PanelItem, Battery, Chart } from '@redhat-cloud-services/frontend-components-pdf-generator';

class DemoApp extends Component {
    state = {
        cats: 35,
        dogs: 55,
        birds: 10,
        generatePDF: false
    }

  
    render() {
        const { cats, dogs, birds, generatePDF } = this.state;
        const handleDownloadButton = () => {
            this.setState({generatePDF: true})
        }
        return (
            <div className="pf-m-redhat-font">
               

                <button onClick={() => handleDownloadButton()}>export pdf</button>

                {generatePDF && 
                    <DownloadButton showButton={false} asyncFunction={() => (new Promise((res) => {
                        setTimeout(() => {
                            res({ rows: [ [ 'one', 'two' ] ] });
                        }, 1000);
                    })).then(data => ([
                        <Fragment key="first-section">
                            <Section title="I am donut chart about pets">
                                <Column>
                                    <Table
                                        withHeader
                                        rows={data.rows}
                                    />
                                </Column>
                            </Section>
                        </Fragment>
                    ]))} buttonProps={{
                        variant: 'link',
                        className: 'some-btn-class'
                    }} />
                }
            </div>
        );
    }
}
ReactDOM.render(<DemoApp />, document.querySelector('.demo-app'));

```

### DIFF
--- a/packages/pdf-generator/src/DownloadButton.js
+++ b/packages/pdf-generator/src/DownloadButton.js
@@ -5,62 +5,68 @@ import { Button } from '@patternfly/react-core';
 import PDFDocument from './components/PDFDocument';
 
 class DownloadButton extends React.Component {
-    state = {
-        asyncPages: []
-    };
-    render() {
-        const {
-            fileName,
-            label,
-            isPreview,
-            asyncFunction,
-            buttonProps,
-            ...props
-        } = this.props;
+  state = {
+      asyncPages: []
+  };
+  render() {
+    console.log(new Date())
+    const {
+        fileName,
+        label,
+        isPreview,
+        asyncFunction,
+        buttonProps,
+        ...props
+    } = this.props;
 
-        return (
-            <React.Fragment>
-                { isPreview ? <PDFViewer>
-                    <PDFDocument { ...props } />
-                </PDFViewer> : asyncFunction ? <React.Fragment>
-                    <Button onClick={ () => {
-                        this.setState({
-                            asyncPages: []
-                        }, () => {
-                            Promise.resolve(asyncFunction()).then(asyncPages => this.setState({
-                                asyncPages
-                            }));
-                        });
-                    }} { ...buttonProps }>{label}</Button>
-                    {this.state.asyncPages.length > 0 && (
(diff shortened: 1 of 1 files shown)


packages/pdf-generator/src/DownloadButton.js
```diff
@@ -5,77 +5,106 @@ import { Button } from '@patternfly/react-core';
...
-    render() {
-        const {
-            fileName,
-            label,
-            isPreview,
-            asyncFunction,
-            buttonProps,
-            ...props
-        } = this.props;
+    this.updateAsyncPages = this.updateAsyncPages.bind(this);
```
karelhala, line 14: No need to use `bind` this repository is using autobinding from babel.

packages/pdf-generator/src/DownloadButton.js
```diff
@@ -5,77 +5,106 @@ import { Button } from '@patternfly/react-core';
...
-                </PDFDownloadLink> }
-            </React.Fragment>
-        );
+    if (!showButton) {
+      this.updateAsyncPages();
     }
+  }
+
+  render() {
+    console.log(new Date())
```
karelhala, line 39: `console.log`

karelhala:
Looking hella good! Just 2 minor issues and we can release it.

packages/pdf-generator/src/DownloadButton.js
katerinapat: ah, forgot it

packages/pdf-generator/src/DownloadButton.js
katerinapat: removed

karelhala (approved):
Perfect!

PR #786, opened 2020-10-19
Use correct permissions for READ inventory

karelhala:


karelhala:
Just a sidenote, this has been released as hotfix so chrome can consume it right away. We want to release it in standard pipeline as well so we have stable release as well.

rvsia (approved):

PR #787, opened 2020-10-20
Add status callbacks for pdf-generator

katerinapat:


karelhala (changes requested):
No need for that `if (typeof onSuccess === 'function') {`, this is already gated by `propTypes` if someone passes something else it'll blow because of it. Also add default functions to `DownloadButton.defaultProps`
```JS
onSuccess: () => undefined,
onError: () => undefined,
onLoading: () => undefined
```

katerinapat:
> No need for that `if (typeof onSuccess === 'function') {`, this is already gated by `propTypes` if someone passes something else it'll blow because of it. Also add default functions to `DownloadButton.defaultProps`
> 
> ```js
> onSuccess: () => undefined,
> onError: () => undefined,
> onLoading: () => undefined
> ```

done :)

karelhala (approved):

PR #788, opened 2020-10-21
Custom text not auth

### LINKED ISSUE #689: Display text translations in Components
**Is your feature request related to a problem? Please describe.**
[Subscription Watch](https://github.com/RedHatInsights/curiosity-frontend) uses this library and all of the text ther is translated with i18n/i18next.  We want to be able to use shared components from this library that contain standard text (in particular Maintenance and NotAuthorized).  Unfortunately the text of those components is not translated, so using them would cause us to have a mix of translated and untranslated text.

**Describe the solution you'd like**
I would like for translations of all display text in components of this library to be built into the library somehow.

**Describe alternatives you've considered**
An alternative solution might be to have all text in components overwritable via props, so that we could replace every instance of text with similar text that's translated, but that would create unnecessary redundancy and diversion between apps using this library.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-8925

closes #689 

### Changes
- allow customizing all texts in Maintenance and NotAuthorized components
- add docs for Maintenance and NotAuthorized components

### DIFF
--- a/packages/components/doc/maintenance.md
+++ b/packages/components/doc/maintenance.md
@@ -1,0 +1,15 @@
+# Maintenance
+
+## Props
+
+|name|type|required|default|
+|---|---|---|---|---|
+|utcStartTime|string|`false`|10am|
+|utcEndTime|string|`false`|12am|
+|startTime|string|`false`|6am|
+|endTime|string|`false`|8am|
+|timeZone|string|`false`|EST|
+|description|node|`false`|`Description`|
+|redirectLink|string|`false`|https://status.redhat.com/incidents|
+|title|node|`false`|Maintenance in progress|
+|className|string|`false`|`undefined`|
--- a/packages/components/doc/maintenance.md
+++ b/packages/components/doc/maintenance.md
@@ -3,7 +3,7 @@
 ## Props
 
 |name|type|required|default|
-|---|---|---|---|---|
+|---|---|---|---|
 |utcStartTime|string|`false`|10am|
 |utcEndTime|string|`false`|12am|
 |startTime|string|`false`|6am|
--- a/packages/components/doc/notAuthorized.md
+++ b/packages/components/doc/notAuthorized.md
@@ -1,0 +1,15 @@
+# NotAuthorized
+
+## Props
+
+|name|type|required|default|description|
+|---|---|---|---|---|---|
+|serviceName|node|if `title` prop is undefined|`undefined`|Name of a service for which current user is not authorized.
+|title|node|`false`|`undefined`||
+|description|node|`ContactBody`|`false`||
+|icon|`React.ComponentType`|`false`|`LockIcon`|Main empty state icon|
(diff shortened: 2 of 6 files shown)


packages/components/src/Components/Maintenance/Maintenance.js
```diff
@@ -35,13 +35,15 @@ const Maintenance = ({ utcStartTime, utcEndTime, startTime, endTime, timeZone, d
...
 
 Maintenance.propTypes = {
-    utcStartTime: propTypes.string,
-    utcEndTime: propTypes.string,
-    startTime: propTypes.string,
-    endTime: propTypes.string,
-    timeZone: propTypes.string,
-    description: propTypes.node,
-    redirectLink: propTypes.string
+    utcStartTime: PropTypes.string,
```
rvsia, line 38: Maybe when we are here, we could make these `node`s

Hyperkid123: Yeah but I think we could do a larger scale revision of prop types. There is a lot of this stuff all over the repository.

packages/components/doc/notAuthorized.md
```diff
@@ -0,0 +1,15 @@
+# NotAuthorized
+
+## Props
+
+|name|type|required|default|description|
+|---|---|---|---|---|---|
+|serviceName|node|if `title` prop is undefined|`undefined`|Name of a service for which current user is not authorized.
```
rvsia, line 7: Missing |

packages/components/doc/notAuthorized.md
```diff
@@ -0,0 +1,15 @@
...
+|title|node|`false`|`undefined`||
+|description|node|`ContactBody`|`false`||
+|icon|`React.ComponentType`|`false`|`LockIcon`|Main empty state icon|
+|showReturnButton|`boolean`|`false`|`true`|Flag to display the return button. BUtton will return users to previous page or to landing page. User will be redirected to landing page if there is no history stack.|
```
rvsia, line 11: `BUtton`

packages/components/doc/maintenance.md
```diff
@@ -0,0 +1,15 @@
+# Maintenance
+
+## Props
+
+|name|type|required|default|
+|---|---|---|---|---|
```
rvsia, line 6: You have 4 columns, not 5

packages/components/doc/notAuthorized.md
```diff
@@ -0,0 +1,15 @@
+# NotAuthorized
+
+## Props
+
+|name|type|required|default|description|
+|---|---|---|---|---|---|
```
rvsia, line 6: Wrong number of columns

packages/components/doc/maintenance.md
Hyperkid123: There is no description required really.

rvsia (approved):

PR #790, opened 2020-10-21
pdf-generator: bind "this"

katerinapat:
it seems autobinding didn't work and as a result, this is undefined
image

### DIFF
--- a/packages/pdf-generator/src/DownloadButton.js
+++ b/packages/pdf-generator/src/DownloadButton.js
@@ -11,9 +11,10 @@ class DownloadButton extends React.Component {
         this.state = {
             asyncPages: []
         };
+
     }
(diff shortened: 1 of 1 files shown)


karelhala (commented):
Aaah, I see. I just glanced over the code and missed the function definition, the problem is not in autobinding, it's with how the function `updateAsyncPages` is written, can you change it to `updateAsyncPages = () => {`?

katerinapat:
> Aaah, I see. I just glanced over the code and missed the function definition, the problem is not in autobinding, it's with how the function `updateAsyncPages` is written, can you change it to `updateAsyncPages = () => {`?

sure, it's done

karelhala (approved):

PR #791, opened 2020-10-21
Change RBAC wording for inventory

karelhala:
### RBAC wording changed

Since we want to be more technical and straight forward when user has no access to inventory, we have to slightly change what user sees when looking at page without access to inventory.

### DIFF
--- a/packages/inventory/src/shared/AccessDenied.js
+++ b/packages/inventory/src/shared/AccessDenied.js
@@ -1,17 +1,26 @@
-import React, { Fragment } from 'react';
+import React from 'react';
+import PropTypes from 'prop-types';
 import { NotAuthorized } from '@redhat-cloud-services/frontend-components/components/cjs/NotAuthorized';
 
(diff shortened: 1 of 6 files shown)


packages/inventory/src/shared/AccessDenied.js
```diff
@@ -1,17 +1,26 @@
...
 );
 
+AccessDenied.propTypes = {
+    title: PropTypes.string,
+    description: PropTypes.node
+};
+
+AccessDenied.defaultProps = {
+    title: 'This application requires Inventory permissions',
+    description: <div>
```
rvsia, line 21: Is there a reason why it's needed be wrapped in a div?

karelhala: To properly style the message, it needs to be on 2 lines, similiar to dashboard. The selector here would be too weak to overpower PF's one.

rvsia (approved):

PR #792, opened 2020-10-22
Add new rules to eslint plugin

karelhala:
### Update default rules

When consuming eslint-plugin it will throw errors when destructing object with unused vars, this PR fixes it by using `ignoreRestSiblings` rule. Also force prettier to use single quotes over double ones by adding `prettier/prettier: [ { singleQuote: true } ]`.

### DIFF
--- a/packages/eslint-config/index.js
+++ b/packages/eslint-config/index.js
@@ -22,6 +22,16 @@ module.exports = {
         'plugin:react/recommended',
         'prettier/react'
     ],
+    rules: {
+        'no-unused-vars': [
+            'error',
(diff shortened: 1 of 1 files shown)


rvsia (approved):

Hyperkid123:
@karelhala :+1: I am not 100%sure but I believe the build process for this package is still broken.

karelhala:
@Hyperkid123 if this won't publish it I'll try to investigate why is it not releasing

karelhala:
release

karelhala:
release

PR #802, opened 2020-11-02
Disable apply selected if nothing is selected

karelhala:
### Fixes RHCLOUD-9829

When no tags are selected we don't have to allow users to submit selection of tags via modal. This PR fixes such issue by checking if every tab has nothing selected.

### DIFF
--- a/packages/components/src/Components/TagModal/TagModal.js
+++ b/packages/components/src/Components/TagModal/TagModal.js
@@ -88,8 +88,8 @@ class TagModal extends Component {
                             variant="primary"
                             isDisabled={
                                 isTabbed ?
-                                    Object.values(selected).every(values => values?.length === 0) :
-                                    selected?.length === 0
+                                    Object.values(selected || {}).every(values => !values || values?.length === 0) :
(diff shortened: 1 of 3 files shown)


rvsia (approved):

packages/components/src/Components/TagModal/TagModal.js
```diff
@@ -83,10 +83,19 @@ class TagModal extends Component {
...
+                            key="confirm"
+                            variant="primary"
+                            isDisabled={
+                                isTabbed ?
+                                    Object.values(selected || {}).every(values => !values || values?.length === 0) :
+                                    !selected || selected?.length === 0
+                            }
+                            onClick={(e) => {
+                                onApply();
+                                toggleModal(e, true);
```
Hyperkid123, line 96: This event passing could potentially trigger the react synthetic event console error when used in async way (don't know if that is possible).

You can avoid that calling `event.persist()` before passing it to the `toggleModal` callback.

Hyperkid123 (approved):

PR #806, opened 2020-11-10
Add select options to SkeletonTable

johnsonm325:
This PR adds onSelect and canSelectAll options to Skeleton table. This is added to make sizing of loading table consistent with table that has checkboxes

### DIFF
--- a/packages/components/src/Components/SkeletonTable/SkeletonTable.js
+++ b/packages/components/src/Components/SkeletonTable/SkeletonTable.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Table, TableHeader, TableBody } from '@patternfly/react-table';
+import { RowSelectVariant, Table, TableHeader, TableBody } from '@patternfly/react-table';
 import { Skeleton, SkeletonSize } from '../Skeleton';
 import PropTypes from 'prop-types';
 
@@ -19,19 +19,29 @@ class SkeletonTable extends React.Component {
     createRows = () => {
         const { colSize, rowSize, columns, paddingColumnSize } = this.props;
         const numberOfCols = columns ? columns.length : colSize;
-        return this.newArray(rowSize).map(() => ({ cells: this.newArray(paddingColumnSize).map(() => '').concat(
+        return this.newArray(rowSize).map(() => ({ disableSelection: true, cells: this.newArray(paddingColumnSize).map(() => '').concat(
             this.newArray(numberOfCols).map(() => ({ title: <Skeleton size={ SkeletonSize.md } /> }))
         ) }));
+    };
+
+    selectVariant = () => {
+        const { hasRadio } = this.props;
+
+        return hasRadio ? (RowSelectVariant?.radio || 'radio') : (RowSelectVariant?.checkbox || 'checkbox');
     };
 
     newArray = (size) => [ ...Array(size) ];
 
     render() {
+        const { canSelectAll, isSelectable, sortBy } = this.props;
         return (
             <Table cells={ this.getColumns() }
                 rows={ this.createRows() }
-                sortBy={ this.props.sortBy }
-                aria-label="Loading">
+                sortBy={ sortBy }
+                aria-label="Loading"
+                onSelect={ isSelectable }
+                selectVariant={ isSelectable ? this.selectVariant() : null }
+                canSelectAll={ canSelectAll }>
                 <TableHeader />
                 <TableBody />
             </Table>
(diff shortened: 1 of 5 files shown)


packages/components/package.json
```diff
@@ -31,7 +31,7 @@
     },
     "homepage": "https://github.com/RedHatInsights/frontend-components/tree/master/packages/components#readme",
     "peerDependencies": {
-        "@patternfly/react-table": ">=4.5.7",
+        "@patternfly/react-table": ">=4.19.5",
```
karelhala, line 34: If you want to update the version of `react-table` you'll have to do it in root package.

packages/components/src/Components/SkeletonTable/SkeletonTable.js
```diff
@@ -47,12 +56,16 @@ SkeletonTable.propTypes = {
     sortBy: PropTypes.shape({
         index: PropTypes.number,
         direction: PropTypes.oneOf([ 'asc', 'desc' ])
-    })
+    }),
+    onSelect: PropTypes.bool,
+    canSelectAll: PropTypes.bool
```
karelhala, line 61: Missing `isRadio` prop.

packages/components/src/Components/SkeletonTable/SkeletonTable.js
```diff
@@ -47,12 +56,16 @@ SkeletonTable.propTypes = {
     sortBy: PropTypes.shape({
         index: PropTypes.number,
         direction: PropTypes.oneOf([ 'asc', 'desc' ])
-    })
+    }),
+    onSelect: PropTypes.bool,
```
karelhala, line 60: This should be `func`
```suggestion
    onSelect: PropTypes.func,
```

packages/components/src/Components/SkeletonTable/SkeletonTable.js
```diff
@@ -47,12 +56,16 @@ SkeletonTable.propTypes = {
...
+    canSelectAll: PropTypes.bool
 };
 
 SkeletonTable.defaultProps = {
     rowSize: 0,
-    paddingColumnSize: 0
+    paddingColumnSize: 0,
+    canSelectAll: false,
+    onSelect: false
```
karelhala, line 68: Empty function instead of bool
```suggestion
    onSelect: () => undefined
```

packages/components/src/Components/SkeletonTable/SkeletonTable.js
```diff
@@ -19,19 +19,28 @@ class SkeletonTable extends React.Component {
...
-        return this.newArray(rowSize).map(() => ({ cells: this.newArray(paddingColumnSize).map(() => '').concat(
+        return this.newArray(rowSize).map(() => ({ disableSelection: true, cells: this.newArray(paddingColumnSize).map(() => '').concat(
             this.newArray(numberOfCols).map(() => ({ title: <Skeleton size={ SkeletonSize.md } /> }))
         ) }));
     };
 
+    selectVariant = () => {
+        const { isRadio } = this.props;
+
+        return isRadio ? RowSelectVariant.radio : RowSelectVariant.checkbox;
```
karelhala, line 30: Unfortunately this is not backwards compatible as older version of PF react table doesn't have `RowSelectVariant` present. We can always fallback to string value, once apps update their version of react-table they'll start using the imported `RowSelectVariant`. With this you don't even have to update the version of react-table in root package.
```suggestion
        return isRadio ? (RowSelectVariant?.radio || 'radio') : (RowSelectVariant?.checkbox || 'checkbox');
```

karelhala (commented):
Looking good, just a few changes and we are golden! While including this I'd slightly change the API and changed the `onSelect` to `isSelectable` and `isRadio` to `hasRadio` these props indicate more cleanly what it does. The `onSelect` would indicate for me without reading the code that I have to pass in function, this function couldn't be fired since we are forbiding the selection of each row.
```JSX
<SkeletonTable
  columns={[ 'first', 'second' ]}
  rowSize={15}
  isSelectable
  hasRadio
/>
```

If you want you can make this future stable as well, by adding 3 more props `tableProps` which you can spread in the table; `rowProps` spread in each row and `cellProps` these will be spread in each cell. This way we will give consumers full access to each partial and they can add/remove/change props on the fly.

packages/components/src/Components/SkeletonTable/SkeletonTable.test.js
```diff
@@ -18,4 +18,25 @@ describe('SkeletonTable component', () => {
         const wrapper = shallow(<SkeletonTable columns={[ 'first', 'second' ]} rowSize={15} />);
         expect(toJson(wrapper)).toMatchSnapshot();
     });
+
+    it.skip('should render correctly with checkboxes', () => {
```
karelhala, line 22: No need to skip these
```suggestion
    it('should render correctly with checkboxes', () => {
```

packages/components/src/Components/SkeletonTable/SkeletonTable.test.js
```diff
@@ -18,4 +18,25 @@ describe('SkeletonTable component', () => {
...
+        const wrapper = shallow(<SkeletonTable
+            columns={[ 'first', 'second' ]}
+            rowSize={15}
+            isSelectable={ true }
+        />);
+
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+
+    it.skip('should render correctly with radio buttons', () => {
```
karelhala, line 32: No need to skip these
```suggestion
    it('should render correctly with radio buttons', () => {
```

karelhala:
Looking good! Just un those skipped tests and we are golden!

karelhala (approved):

PR #813, opened 2020-11-19
Update config file to use new webpack 5 config

karelhala:


Hyperkid123:
@karelhala we should add these aswell:
```
```

packages/config/src/chrome-render-loader.js
```diff
@@ -0,0 +1,5 @@
+function chromeRenderLoader(source) {
+    return `if(!window.isChrome2){${JSON.stringify(source)}}`;
```
karelhala, line 2: I would change this to something more insights specific `insights.chrome.isChrome2`

Hyperkid123: That can be would be a problem. The flag should be set in the chrome HTML template to ensure that it's there even if the chrome was not loaded yet.

If we put that into the chrome init function it may be too late.

Hyperkid123: NVM, i see that the chrome init is not overriding the existing chrome object on the window. We can change it.

packages/config/src/chrome-render-loader.js
```diff
@@ -0,0 +1,5 @@
+function chromeRenderLoader(source) {
+    return `var isChrome2 = window.insights && window.insights.chrome && window.insights.chrome.isChrome2 || false; if(!window.isChrome2){${JSON.stringify(source)}}`;
```
karelhala, line 2: the var `isChrome2` is not used and `window.isChrome2` is still there.

packages/config/package.json
```diff
@@ -17,30 +17,35 @@
...
-        "webpack-hot-client": "^4.1.1",
-        "webpack-serve": "1.0.2",
-        "write-file-webpack-plugin": "^4.5.0"
+        "sass-loader": "^10.1.0",
+        "source-map-loader": "^1.1.2",
+        "stream-browserify": "^3.0.0",
+        "style-loader": "^2.0.0",
+        "ts-loader": "^8.0.11",
+        "util": "^0.12.3",
+        "webpack": "^5.6.0",
```
Hyperkid123, line 43: @karelhala we have to bump the webpack version to latest due to the issue with dev-server

packages/config/federated-modules.js
```diff
@@ -0,0 +1,77 @@
...
+
+module.exports = ({
+    root,
+    exposes,
+    shared = [],
+    debug,
+    moduleName
+}) => {
+    const { dependencies, insights } = require(resolve(root, './package.json')) || {};
+    const appName = moduleName || (insights && insights.appname);
```
redallen, line 12: Manged Kafka uses `crc.name`. Which is correct?

karelhala: It should be `insights.appname` this is used widely accross many configs. It's unfortunate that the name changed midway, but switching to a new one would be a few inconveniences.

redallen: Can't you just short-circuit an `|| crc.name`?

packages/config/federated-modules.js
```diff
@@ -0,0 +1,77 @@
...
+        console.log('Using package at path: ', resolve(root, './package.json'));
+        console.log('Using appName: ', appName);
+        console.log(`Using ${exposes ? 'custom' : 'default'} exposes`);
+        console.log('Number of custom shared modules is: ', shared.length);
+        console.log('Number of default shared modules is: ', sharedDeps.length);
+    }
+
+    return new ModuleFederationPlugin({
+        name: appName,
+        filename: `${appName}.[chunkhash].js`,
```
redallen, line 65: This makes it near impossible for an app to load another app unless the `chunkhash`s for each build are stored in yet another file that is `fetch`ed on each page load.

In managed Kafka's case mk-ui-host would like to load a module from mk-ui-frontend (and it's even kindly abstracted the URL into a `fetch`ed json file). Is there support for this or should our frontends just not use this helper?

Hyperkid123: The reason why we have hashes even in fed modules is that akamai has a nasty habit of serving stale JS assets.

The common config generates a `fedmods.json` file with all necessary metadata to then register the federated module in chrome.

This registration happens only once, on the first app load. We chose to use runtime module registration over build time in order to prevent additional chrome builds whenever any app changes its filenames or new app is added etc.

That said, applications do not have to use the common plugin for federated modules. We realize that many apps may choose different naming schemes. This plugin is optional. You can check the docs here: https://github.com/RedHatInsights/insights-chrome/blob/chrome-2.0/docs/migrationGuide.md#add-federated-modules-plugin

packages/config/federated-modules.js
karelhala: We could do that, but the problem is that the `insights.appname` is used widely across plathera configs not only at this place. QE is relying on it as well, couple of other configs (including shared config) uses this constant. We could change it, but that can lead to unwanted bugs without proper investigation where it's used and changing it on all other places. Further this PR does not aim to address this change at all.

Hyperkid123 (approved):

Hyperkid123:
release major

PR #819, opened 2020-11-30
fix(Compliance SystemRulesTable): RHICOMPL-1032 filtering by policy and unique policies

vkrizan:
Fixes filter by policy on SystemRulesTable:
* filtering is done by policy id (or by external profile id) instead of ref_id

Screenshot_2020-11-30 Compliance
(Note, the one with a custom name is a policy and the other is an external profile/policy.)

### DIFF
--- a/packages/inventory-compliance/src/Compliance.js
+++ b/packages/inventory-compliance/src/Compliance.js
@@ -19,6 +19,7 @@ query System($systemId: String!){
         id
         name
         profiles {
+            id
             name
             policyType
             refId
@@ -56,9 +57,9 @@ const SystemQuery = ({ data: { system }, loading, hidePassed }) => (
         <SystemRulesTable hidePassed={ hidePassed }
             system={ system }
             columns={ columns }
-            profileRules={ system?.profiles.map((profile) => ({
+            profileRules={ system?.profiles.map(profile => ({
                 system: system.id,
-                profile: { refId: profile.refId, name: profile.name },
+                profile,
                 rules: profile.rules
             })) }
             loading={ loading } />
--- a/packages/inventory-compliance/src/SystemRulesTable.js
+++ b/packages/inventory-compliance/src/SystemRulesTable.js
@@ -52,7 +52,7 @@ class SystemRulesTable extends React.Component {
 
     parentRowForRule = (rule) => {
         const {
-            title, identifier, policies, severity, compliant, rowKey,
+            title, identifier, profiles, severity, compliant, rowKey,
             remediationAvailable, refId, isOpen, isSelected
         } = rule;
         const cells = this.props.columns.map((column) => {
@@ -63,7 +63,7 @@ class SystemRulesTable extends React.Component {
                     cell = <RuleTitle title={ title } identifier={ identifier } />;
                     break;
                 case 'Policy':
-                    cell = policies.map((p) => p.name).slice(0, 1).join(', ');
+                    cell = profiles.map((p) => p.name).slice(0, 1).join(', ');
                     break;
                 case 'Severity':
                     cell = (severity.toLowerCase() === 'high' ? HIGH_SEVERITY :
--- a/packages/inventory-compliance/src/SystemRulesTable.js
+++ b/packages/inventory-compliance/src/SystemRulesTable.js
@@ -184,9 +184,12 @@ class SystemRulesTable extends React.Component {
     }
 
     updatePolicyFilterConfig = () => {
-        const policies = this.props.profileRules.map((p) => (
-            p.profile
-        )).filter((p) => !!p);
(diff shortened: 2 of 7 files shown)


packages/inventory-compliance/src/Compliance.js
```diff
@@ -56,10 +57,10 @@ const SystemQuery = ({ data: { system }, loading, hidePassed }) => (
         <SystemRulesTable hidePassed={ hidePassed }
             system={ system }
             columns={ columns }
-            profileRules={ system?.profiles.map((profile) => ({
+            profileRules={ system?.profiles.map(({ id, refId, name, policy, rules }) => ({
```
bastilian, line 60: Writing it like this would avoid having to assign all properties:
```js
profileRules={ system?.profiles.map((profile) => ({
    system,
    profile,
    rules: profile.rules
})) } 
```

We don't really need to limit what passes through.

packages/inventory-compliance/src/SystemRulesTable.js
```diff
@@ -184,9 +184,13 @@ class SystemRulesTable extends React.Component {
...
-        const policies = this.props.profileRules.map((p) => (
-            p.profile
-        )).filter((p) => !!p);
+        let policies = this.props.profileRules.filter(({ profile }) => !!profile).map(({ profile }) => (
+            {
+                id: profile.policy ? profile.policy.id : profile.id,
+                name: profile.name
+            }
+        ));
+        policies = [ ...new Map(policies.map(policy => [ policy.id, policy ])).values() ];
```
bastilian, line 193: Is there a specific reason, why we need to de-duplicate here (in the frontend)?

In case `SystemRulesTable` is passed duplicate profiles it should rather be de-duplicated where the table is used and not pass in duplicate profiles.

packages/inventory-compliance/src/Utilities/RulesHelper.js
```diff
@@ -1,15 +1,14 @@
-export const toRulesArray = (policiesWithRules) => {
-    const rules = policiesWithRules.flatMap((policy) => (
-        policy.rules.map((rule) => {
+export const toRulesArray = (profilesAndRules) => (
+    profilesAndRules.flatMap((record) => (
```
bastilian, line 2: Please use either `profile` or `policy` which ever one is more appropriate.

packages/inventory-compliance/src/Compliance.js
vkrizan: The limitation was there before. I don't see a reason why not to include the whole profile object.

packages/inventory-compliance/src/SystemRulesTable.js
vkrizan: There are `profileRules` passed consisting of a profile and its rules. What do you have in mind? Passing them as a property?

packages/inventory-compliance/src/Utilities/RulesHelper.js
vkrizan: Neither is in this case. `record` changed to `profileRule`, and `profilesAndRules` to `profileRules`.

packages/inventory-compliance/src/Compliance.js
vkrizan: Updated.
I guess the original idea was not passing the rules two different ways.

packages/inventory-compliance/src/Utilities/RulesHelper.js
```diff
@@ -1,15 +1,14 @@
-export const toRulesArray = (policiesWithRules) => {
-    const rules = policiesWithRules.flatMap((policy) => (
-        policy.rules.map((rule) => {
+export const toRulesArray = (profileRules) => (
```
bastilian, line 1: It is passed an array if profiles with an array of rules, so, `ruleProfiles` would be more appropriate here.

packages/inventory-compliance/src/SystemRulesTable.js
vkrizan: After a discussion, we can remove this line in favor of an assumption that the backend would keep one test result profile per policy.

packages/inventory-compliance/src/Utilities/RulesHelper.js
akofink: As discussed in slack, I'd prefer just naming this `profilesWithRules` or simply `profiles`. I really don't like `profileRule` being a profile and not a rule.

vkrizan: Okay, I've composed it into `profilesWithRules > profileWithRules > profile/rules`.

akofink (approved):
Looks good to me. I was not able to test it, though.

PR #820, opened 2020-12-01
Init DDF Remediations Wizard

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-10893

Added empty remediations DDF Wizard for remediations debug mode

Needs https://github.com/RedHatInsights/insights-chrome/pull/1088

### DIFF
--- a/packages/remediations/src/NewRemediationWizard.js
+++ b/packages/remediations/src/NewRemediationWizard.js
@@ -1,9 +1,10 @@
 import React, { Component, Fragment, createContext } from 'react';
-import componentMapper from '@data-driven-forms/pf4-component-mapper/dist/cjs/component-mapper';
 import FormRenderer from '@data-driven-forms/react-form-renderer/dist/cjs/form-renderer';
 import Pf4FormTemplate from '@data-driven-forms/pf4-component-mapper/dist/cjs/form-template';
 import schemaBuilder from './schema';
 import Deferred from '@redhat-cloud-services/frontend-components-utilities/files/Deffered';
+import Wizard from '@data-driven-forms/pf4-component-mapper/dist/cjs/wizard';
+import componentTypes from '@data-driven-forms/react-form-renderer/dist/cjs/component-types';
 
 const RemediationWizardContext = createContext({
     success: false,
@@ -63,7 +64,7 @@ class RemediationWizard extends Component {
     render () {
 
         return (
-            this.state.open !== false ?
+            this.state.open ?
                 <RemediationWizardContext.Provider
                     value={{ ...this.state.wizardContextValue,
                         setWizardError: this.setWizardError,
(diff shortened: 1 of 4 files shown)


packages/remediations/src/NewRemediationWizard.js
```diff
@@ -0,0 +1,87 @@
...
+                    value={{ ...this.state.wizardContextValue,
+                        setWizardError: this.setWizardError,
+                        setWizardSuccess: this.setWizardSuccess,
+                        setHideForm: this.setHideForm }}>
+                    <FormRenderer
+                        schema={this.state.schema}
+                        container={this.container}
+                        subscription={{ values: true }}
+                        FormTemplate={this.getFormTemplate}
+                        componentMapper={{ ...componentMapper }}
```
Hyperkid123, line 77: there is no need to copy the component mapper object. It's also better for performance, to only add a minimal set of components to the mapper. Instead of passing the whole object, it's better to directly import components (text fields, radio buttons, etc.) and create a mapper object from those.

packages/remediations/src/schema.js
```diff
@@ -0,0 +1,22 @@
+
+export default (container) => ({
+    fields: [
+        {
+            component: 'wizard',
```
Hyperkid123, line 5: use constants instead: https://data-driven-forms.org/schema/constants#heading-componenttypes

karelhala (commented):
Apart from Martin's requests it is looking good!

packages/remediations/src/NewRemediationWizard.js
```diff
@@ -0,0 +1,87 @@
...
+    setHideForm = (hideForm) => this.setState({ wizardContextValue: { ...this.state.wizardContextValue, hideForm } });
+
+    closeWizard = () => {
+        this.setOpen(false);
+    }
+
+    render () {
+
+        return (
+            this.state.open !== false ?
```
PanSpagetka, line 66: Please change this to `this.state.open` :D

PanSpagetka (changes requested):
One small thing, but otherwise it looks good to me!

packages/remediations/src/NewRemediationWizard.js
```diff
@@ -0,0 +1,86 @@
...
+                    value={{ ...this.state.wizardContextValue,
+                        setWizardError: this.setWizardError,
+                        setWizardSuccess: this.setWizardSuccess,
+                        setHideForm: this.setHideForm }}>
+                    <FormRenderer
+                        schema={this.state.schema}
+                        container={this.container}
+                        subscription={{ values: true }}
+                        FormTemplate={this.getFormTemplate}
+                        componentMapper={{}}
```
Hyperkid123, line 76: :mage_man: is missing my friend!

PanSpagetka (approved):

packages/remediations/src/NewRemediationWizard.js
fhlavac: +1 mage_man

karelhala (approved):

packages/remediations/src/NewRemediationWizard.js
fhlavac: +1 :mage_man: :heavy_check_mark:

Hyperkid123 (approved):

PR #823, opened 2020-12-02
fix(Compliance SystemRulesTable): RHICOMPL-1183 dupe policies on filter

vkrizan:
Pull testResultProfiles instead of composite where all policy profiles
are returned.
This will deduplicate listed policies on the SystemRulesTable filter,
with the assumption that a system has only one test result/profile per
policy.


bastilian (approved):
LGTM! Thank you @vkrizan!

akofink (approved):
Thanks!

PR #824, opened 2020-12-02
fix(Compliance SystemRulesTable): RHICOMPL-1184 sorting by policy

vkrizan:


akofink (approved):
Seems straightforward :)

PR #826, opened 2020-12-07
Add playbook select step

PanSpagetka:
Add playbook select step. I still need to figure out from where I should get the info about what can or cannot be remediated. And also what should be displayed in expandable panel. However, since this is gated behind debug, I would merge it anyway so Filip can continue.

https://issues.redhat.com/browse/RHCLOUD-10894


packages/remediations/src/newSteps/selectPlaybook.scss
```diff
@@ -0,0 +1,13 @@
+.ins-c-remediations-can-be-remediated { 
+    color: var(--pf-global--success-color--100);
+    margin-bottom: 0px !important;
```
Hyperkid123, line 3: Can you try using `pf-u-mb-0` and other spacing classes directly on the elements? It might work.

fhlavac (approved):
For now it looks good, thanks :)

packages/remediations/src/newSteps/selectPlaybook.scss
PanSpagetka: Yep, it works. Thanks!

Hyperkid123 (approved):

PR #831, opened 2020-12-08
Add content to expandable issues list

PanSpagetka:
Add content to expandable issues list.

Screenshot from 2020-12-08 10-42-14

### DIFF
--- a/packages/remediations/src/newSteps/selectPlaybook.scss
+++ b/packages/remediations/src/newSteps/selectPlaybook.scss
@@ -4,7 +4,8 @@
 .ins-c-remediations-cannot-be-remediated { 
     color: var(--pf-global--warning-color--100);
 }
-
-.ins-c-remediations-select-playbook-description { 
-    margin-left: var(--pf-global--spacer--md);;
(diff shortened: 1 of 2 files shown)


packages/remediations/src/newSteps/selectPlaybook.scss
```diff
@@ -6,5 +6,11 @@
 }
 
 .ins-c-remediations-select-playbook-description { 
-    margin-left: var(--pf-global--spacer--md);;
-}
\ No newline at end of file
+    margin-left: var(--pf-global--spacer--md);
```
Hyperkid123, line 9: you can use PF class names for these: 'pf-u-ml-md'. You can interpret the `ml` like **m** is margin or padding (m/p) and the **l** os optional position of the spacing. None means all four sides and then you have t, b, r, l for a specific side.

So there is no need for custom classes just for spacing.

PanSpagetka: Thanks!

Hyperkid123 (approved):

PR #837, opened 2020-12-09
Improve loading of deps to speed up and remove unwanted chunks

karelhala:
### Import optimization

When importing advisor or compliance inv details these chunks will bloat the final build by 7MB and by 15MB, by further inpection with @Hyperkid123 we've found that remediations button is importing all `icons` and all `components` from PF. This PR fixes such issue by:
* direct importing tokens in compliance (removes about 7MB of deps)
* add esm and cjs remediations build (removes unwanted imports)
* import remediations button from remediations package and change how the button is imported in babel (removes about 5MB of deps)

This results in trimming the imported chunks to 1MB for advisor and 2MB for compliance, which is acceptable.


packages/remediations/package.json
```diff
@@ -7,7 +7,10 @@
         "access": "public"
     },
     "scripts": {
-        "build": "NODE_ENV=production webpack --config config/webpack.config.js",
+        "build": "NODE_ENV=production webpack --config config/webpack.config.js && npm run build:js && npm run build:esm && npm run build:umd",
```
Hyperkid123, line 10: Is the webpack build still relevant?

karelhala: Yup, we still want to build the remediations wizard that is used in chrome. Now there's webpack build just for the wizard and rollup build for button.

Hyperkid123 (approved):

PR #843, opened 2020-12-14
feat(Remediation): RHICOMPL-1124 Disabled for unsupported

akofink:
Disable remediations for rules or systems that were uploaded from an
unsupported configuration.

### DIFF
--- a/packages/inventory-compliance/src/ComplianceRemediationButton.js
+++ b/packages/inventory-compliance/src/ComplianceRemediationButton.js
@@ -31,7 +31,7 @@ class ComplianceRemediationButton extends React.Component {
     )
 
     rulesWithRemediations = (rules, system) => {
-        return rules.filter(rule => rule.remediationAvailable).map(
+        return rules.filter(rule => rule.remediationAvailable && system.supported).map(
             rule => this.formatRule(rule, this.ruleProfile(rule, system).refId, system.id)
         );
     }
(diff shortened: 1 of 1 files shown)


packages/inventory-compliance/src/ComplianceRemediationButton.js
```diff
@@ -31,7 +31,7 @@ class ComplianceRemediationButton extends React.Component {
     )
 
     rulesWithRemediations = (rules, system) => {
-        return rules.filter(rule => rule.remediationAvailable).map(
+        return rules.filter(rule => rule.remediationAvailable && system.supported).map(
```
vkrizan, line 34: Is the `system.supported` the correct property for checking whether the test result is supported?

How do you ensure that the consumer of this button provides that information? Can you please verify it?

akofink: Couldn't we say the same thing about `remediationAvailable`? If `supported` doesn't exist on the system, it will show as unsupported. I think that's OK behavior for any consumers of this component - it doesn't break, but it errs on the cautious side (unsupported rather than supported by default).

vkrizan: So is the `supported` on the `system`? Isn't it part of a Profile or a TestResult? Or is it injected somewhere?

akofink: Yes, check the SystemStore

vkrizan: Oh, right. This has the same confusion as in https://github.com/RedHatInsights/compliance-frontend/pull/883#discussion_r544982666. The `system` is actually `facts.compliance`. Mind to rename the variable also here?

vkrizan (approved):

Hyperkid123 (approved):

PR #849, opened 2020-12-17
fix(RulesTable): RHICOMPL-914 - Placeholders should be sentence case

bastilian:


Hyperkid123 (approved):

Hyperkid123:
Merging this PR without Travis check. Code is simple enough and snapshots are proving the change is legit.

PR #851, opened 2020-12-17
Use new version of scalprum package

karelhala:


rvsia (commented):
Shouldn't package-lock be changed too?

karelhala:
@rvsia this is peer dep, it won't update the lock file.

Hyperkid123 (approved):

PR #853, opened 2020-12-21
fix(ComplianceRemediationButton): RHICOMPL-873 color

vkrizan:
Fixes color of ComplianceRemediationButton to match button primary color
(white).  The color is also used for on disabled state.

In general scenario the icon within a button would share the same color
as the text.  Unfortunately, this is not the case (?).

Use of direct CSS variables is deliberate for consistency.  Previously
used background color (provided from react-tokens) had a wrong color.

Drops unused patternfly/react-tokens from peer dependencies.

Remediation Button-active Remediation Button-disabled


akofink (approved):
Thanks!

Hyperkid123 (approved):

PR #854, opened 2020-12-21
Polish docs

Hyperkid123:
### Changes
- Added deprecated alert
- Added JSON viewer for extra complex prop types
- Added simple examples for battery components
  - We should start injecting CSS with components. It will improve the DX.
- Restrict the layout content max width
- Format the prop table content (new lines, spacing, code wrappers, links,...)

### Deprecation alert
screenshot-localhost_3000-2020 12 21-15_09_16

### Very complex props
screenshot-localhost_3000-2020 12 21-15_10_43

### Layout width
screenshot-localhost_3000-2020 12 22-09_11_18


karelhala (approved):

karelhala:
> We should start injecting CSS with components. It will improve the DX.

I guess that this change will bresak current import way, if app is importing anything it will break because of missing css file.

Hyperkid123:
@karelhala not necessarily. We can provide empty dummy CSS files and deprecation warnings via custom webpack loader to sjared config.

PR #858, opened 2021-01-05
Add named remediation button to remediation package

karelhala:
### No component found

When importing Remediation button from UMD package, there is an error with missing remediation button. This PR fixes such issue by exporting the button as named dependency as well.

### DIFF
--- a/packages/inventory-compliance/package.json
+++ b/packages/inventory-compliance/package.json
@@ -4,6 +4,7 @@
     "description": "Detail part of compliance for RedHat Cloud Services.",
     "browser": "index.js",
     "module": "esm/index.js",
+    "main": "cjs/main.js",
     "publishConfig": {
         "access": "public"
     },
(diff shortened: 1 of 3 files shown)


PanSpagetka (approved):

Hyperkid123 (changes requested):
@karelhala can you also point the "main" in package JSON to the CJS bundle? Right now there is only the browser entry and it points to the UMD version.

packages/inventory-compliance/package.json
```diff
@@ -4,6 +4,7 @@
     "description": "Detail part of compliance for RedHat Cloud Services.",
     "browser": "index.js",
     "module": "esm/index.js",
+    "main": "cjs/main.js",
```
Hyperkid123, line 7: isnt it `cjs/index.js`?

Hyperkid123 (approved):

Hyperkid123 (approved):

PR #872, opened 2021-01-07
Dont show issues coutn that cant be remediated

PanSpagetka:
We only have data to show issues that can be remediated
Screenshot from 2021-01-11 13-58-45


fhlavac (approved):
Looks good for now, it would probably need some data adjustments when finishing the review step.

PR #873, opened 2021-01-07
Shuffle deps around so we can run docs server from root

karelhala:


Hyperkid123 (commented):
@karelhala did check if the build command and the prod server is still working?

karelhala:
@Hyperkid123 unfortunately the build is broken and serving won't work because of broken MD generation.

Hyperkid123:
@karelhala which component was it? The build was passing before and it's even deployed on test hosting atm.

PR #876, opened 2021-01-11
Include remediations wizard in build

PanSpagetka:
When I was trying to run local remediation setup, it always blowed up. It seemed that `RemediationWizard` was not build properly and this seems to fix it. I don't really understand FEC build process, so feel free to request changes :laughing:

### DIFF
--- a/packages/remediations/config/rollup.config.js
+++ b/packages/remediations/config/rollup.config.js
@@ -65,6 +65,13 @@ export default rollupConfig(
     }, {
         RemediationButton: 'src/RemediationButton.js',
         index: 'src/index.js'
+    }, {
+        RemediationWizard: 'src/RemediationWizard.js',
+        index: 'src/index.js'
+    },
+    {
+        NewRemediationWizard: 'src/NewRemediationWizard.js',
+        index: 'src/index.js'
     }],
     './'
 );
--- a/packages/remediations/config/webpack.config.js
+++ b/packages/remediations/config/webpack.config.js
@@ -13,7 +13,7 @@ module.exports = (env) => ({
         minimizer: [ new TerserJSPlugin({}), new OptimizeCSSAssetsPlugin({}) ]
     },
     entry: {
-        index: [ './src/index.js', './src/RemediationButton.js' ],
+        index: [ './src/index.js', './src/RemediationButton.js', './src/NewRemediationWizard.js', './src/RemediationWizard.js' ],
         remediationsApi: './src/api/index.js'
     },
     output: {


Hyperkid123:
@PanSpagetka this won't be enough. I think we are missing the export of `RemediationButton` in the index file. Can you add that instead?

Hyperkid123 (changes requested):
RemediationButton must be exported from index.js

karelhala:
> RemediationButton must be exported from index.js

Not really, as this would bloat the build. You should from now on import remediations button from `esm/cjs`

karelhala (changes requested):
`RemediationWizard` and `NewRemediationWizard` should be in the first and second element of array, not as 3rd and 4th. Looking further at the build proccess, no need to have the 2nd array element at all if you remove the `umd` build via rollup in package.json

```JS
[{
        RemediationButton: 'src/RemediationButton.js',
        RemediationWizard: 'src/RemediationWizard.js',
        NewRemediationWizard: 'src/NewRemediationWizard.js',
        index: 'src/index.js'
    }, {
        RemediationButton: 'src/RemediationButton.js',
        RemediationWizard: 'src/RemediationWizard.js',
        NewRemediationWizard: 'src/NewRemediationWizard.js',
        index: 'src/index.js'
    }]
```

Hyperkid123:
The `RemediationWizard` and `NewRemediationWizard` are not really the issue. The chrome is importing only the index. 
```jsx
const remediationsData = await import(/* webpackChunkName: "remediations" */ '@redhat-cloud-services/frontend-components-remediations');
```
We can keep it there though. It's actually better to adjust the chrome to only import what is needed. But if we want to make it work and completely remove the index import, we would also need to export the `openWizard` function. If we don't do that there is no point in export these components as partials.

The remediations button is the main problem. It actually also breaks the inventory.

As @karelhala stated the fix is to remove the rollup version of the UMD build because it overrides the webpack build output.
@PanSpagetka  remove "build:umd" line and from the top-level "build" command from remediations `package.json` and it will work. I've just tested that it fixes the undefined remediations button component.

PanSpagetka:
@Hyperkid123 @karelhala Maybe this PR is not relevant. What actually fixed my issue was, that I ran `npm run build` instead of `npm start` even without my changes `npm run build` does give working result while `npm start` does not.

Hyperkid123:
@PanSpagetka yes that is beause the start is running the webpack build. But build runs webpack and the rollup which overrides the webpack output.

```
"build": "NODE_ENV=production webpack --config config/webpack.config.js && npm run build:js && npm run build:esm && npm run build:umd",
```

If you remove the ` && npm run build:umd` part it will stop overriding the webpack output.

PanSpagetka:
Closing in favor of #878

PR #880, opened 2021-01-11
Add icon for disabled insights systems

rvsia:
https://issues.redhat.com/browse/RHCLOUD-10958

image

### DIFF
--- a/packages/inventory/src/redux/entities.js
+++ b/packages/inventory/src/redux/entities.js
@@ -17,6 +17,7 @@ import { CullingInformation } from '@redhat-cloud-services/frontend-components/c
 import { TagWithDialog } from '../shared';
 import groupBy from 'lodash/groupBy';
 import TitleColumn from '../components/table/TitleColumn';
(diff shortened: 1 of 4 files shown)


packages/inventory/src/redux/entities.js
```diff
@@ -50,18 +51,21 @@ const defaultColumns = [
...
+                render={({ msg }) => <React.Fragment>
+                    <DateFormat date={ value } extraTitle={ (
+                        <React.Fragment>
+                            <div>{ msg }</div>
                         Last seen:{` `}
-                    </React.Fragment>
-                ) }/>
+                        </React.Fragment>
+                    ) }/>
+                    {rest.insights_id && <InsightsDisconnected />}
```
karelhala, line 67: Shouldn't this be the other way around? If `insights_id` is not present show the `InsightsDisconnected` component?

rvsia: It's a draft. :smile_cat:

karelhala (approved):

PR #881, opened 2021-01-11
Fix async inventory ref by introducing inner ref

karelhala:
### Align async inventory refs

When using new async inventory component on old chrome the refs are broken, because they are wrongly passed to error component. This PR fixes such thing by introducing `innerRef` prop to all async inventory components which is then consumed by Async inventory itself (when errored = no scalprum present) or used as `ref` for the new chrome.

### DIFF
--- a/packages/components/src/Components/Inventory/AppInfo.js
+++ b/packages/components/src/Components/Inventory/AppInfo.js
@@ -6,11 +6,6 @@ import { useHistory } from 'react-router-dom';
 import { useStore } from 'react-redux';
 import { Bullseye, Spinner } from '@patternfly/react-core';
 
-/**
- * Inventory sub component.
- *
- * This component shows tab(s) with detail information about selected system.
- */
 const BaseAppInfo = (props) => {
     const history = useHistory();
     const store = useStore();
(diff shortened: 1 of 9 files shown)


packages/components/src/Components/Inventory/AppInfo.js
```diff
@@ -17,20 +22,23 @@ const AppInfo = React.forwardRef((props, ref) => {
...
+    /** React Suspense fallback component. <a href="https://reactjs.org/docs/code-splitting.html#reactlazy" target="_blank">Learn more</a>. */
     fallback: PropTypes.node
 };
 
-AppInfo.defaultProps = {
+BaseAppInfo.defaultProps = {
     fallback: <Bullseye><Spinner size="xl" /></Bullseye>
 };
 
+const AppInfo = React.forwardRef((props, ref) => <BaseAppInfo innerRef={ref} {...props} />);
```
Hyperkid123, line 42: I know it's not important right now but just a note that this will mess up the new docs generator because the props will be missing.

Hyperkid123 (approved):

packages/components/src/Components/Inventory/AppInfo.js
karelhala: I should actually change the prop definition as well, because consumers will use these when using this.

Hyperkid123 (approved):

PR #882, opened 2021-01-11
Remediation wizard review step

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-10896

This PR is putting the new remediations wizard steps altogether.
- added review step and submitting of the form
- enhanced selectPlaybook step to fetch data necessary for the review step
- added missing issueResolution tile description text
- added error state when resolution fetching fails

@katierik Please if you could take one more look if everything is correct from the UX side, thank you :-)

### DIFF
--- a/packages/remediations/src/NewRemediationWizard.js
+++ b/packages/remediations/src/NewRemediationWizard.js
@@ -15,7 +15,6 @@ import {
     MANUAL_RESOLUTION
 } from './utils';
 import * as api from './api';
-import Deferred from '@redhat-cloud-services/frontend-components-utilities/files/Deffered';
 import Wizard from '@data-driven-forms/pf4-component-mapper/dist/esm/wizard';
 import componentTypes from '@data-driven-forms/react-form-renderer/dist/esm/component-types';
 import SelectPlaybook from './newSteps/selectPlaybook';
@@ -30,7 +29,7 @@ function createNotification (id, name, isNewSwitch) {
     return {
         variant: 'success',
         title: `Playbook ${verb}`,
-        description: <span>You have successfully { isNewSwitch ? 'created' : 'updated' } <a href={ remediationUrl(id) } >{ name }</a>.</span>,
+        description: <span>You have successfully {verb} <a href={ remediationUrl(id) } >{ name }</a>.</span>,
         dismissable: true
     };
 }
@@ -70,7 +69,6 @@ class RemediationWizard extends Component {
         });
 
     openWizard = (data, basePath) => {
-        const deferred = new Deferred();
         const issuesById = keyBy(data.issues, issue => issue.id);
 
         this.loadResolutions(data.issues).then(
@@ -80,7 +78,6 @@ class RemediationWizard extends Component {
                 this.setState({
                     open: true,
                     data,
-                    deferred,
                     basePath,
                     onRemediationCreated: data.onRemediationCreated,
                     schema: schemaBuilder(this.container.current, issuesMultiple),
@@ -112,8 +109,6 @@ class RemediationWizard extends Component {
                 });
             }
         );
-
-        return deferred.promise;
     }
 
     closeWizard = () => {
@@ -147,7 +142,7 @@ class RemediationWizard extends Component {
 
     getResolution = issueId => this.state.resolutions.find(r => r.id === issueId)?.resolutions || [];
 
-    resolver = () => (id, name, isNewSwitch) => (
+    resolver = (id, name, isNewSwitch) => (
         this.state.onRemediationCreated({
             remediation: { id, name },
             getNotification: () => createNotification(id, name, isNewSwitch)
@@ -165,11 +160,11 @@ class RemediationWizard extends Component {
             const { id, name } = values[EXISTING_PLAYBOOK];
             // eslint-disable-next-line camelcase
             api.patchRemediation(id, { add, auto_reboot: values[AUTO_REBOOT] }, this.state.basePath)
-            .then(() => this.resolver(this.state.deferred)(id, name, false));
+            .then(() => this.resolver(id, name, false));
         } else {
         // eslint-disable-next-line camelcase
             api.createRemediation({ name: values[SELECT_PLAYBOOK], add, auto_reboot: values[AUTO_REBOOT] }, this.state.basePath)
-            .then(({ id }) => this.resolver(this.state.deferred)(id, values[SELECT_PLAYBOOK], true));
+            .then(({ id }) => this.resolver(id, values[SELECT_PLAYBOOK], true));
         }
     }
 
--- a/packages/remediations/src/NewRemediationWizard.js
+++ b/packages/remediations/src/NewRemediationWizard.js
@@ -1,28 +1,37 @@
-import React, { Component, Fragment, createContext } from 'react';
+import React, { Component, Fragment } from 'react';
 import keyBy from 'lodash/keyBy';
 import transform from 'lodash/transform';
 import FormRenderer from '@data-driven-forms/react-form-renderer/dist/esm/form-renderer';
 import Pf4FormTemplate from '@data-driven-forms/pf4-component-mapper/dist/esm/form-template';
 import schemaBuilder from './schema';
+import { remediationUrl } from './utils';
+import * as api from './api';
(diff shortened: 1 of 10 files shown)


packages/remediations/src/NewRemediationWizard.js
```diff
@@ -1,28 +1,37 @@
...
+
+    return {
+        variant: 'success',
+        title: `Playbook ${verb}`,
+        description: <span>You have successfully { isNewSwitch ? 'created' : 'updated' } <a href={ remediationUrl(id) } >{ name }</a>.</span>,
+        dismissable: true
+    };
+}
 
 class RemediationWizard extends Component {
```
Hyperkid123, line 29: One thing that we should definitely do before actually deploying this is also refactor the class component to the functional component. Does not have to be in this PR but in a follow-up.

packages/remediations/src/NewRemediationWizard.js
```diff
@@ -137,33 +140,54 @@ class RemediationWizard extends Component {
...
+                    initialValues={{
+                        multiple: !!this.state.resolutions?.find(r => r.resolutions.length > 1),
+                        'manual-resolution': true,
+                        'selected-resolutions': {}
+                    }}
+                    componentMapper={{
+                        [componentTypes.WIZARD]: Wizard,
+                        ...this.state.mapperExtension
+                    }}
+                    onSubmit={(_, formOptions) => {
```
Hyperkid123, line 168: This function is quite big. Can it be moved to a stand-alone function?

packages/remediations/src/NewRemediationWizard.js
```diff
@@ -137,33 +140,54 @@ class RemediationWizard extends Component {
...
+                        const add = { issues, systems: this.state.data.systems };
+                        if (values['existing-playbook-selected']) {
+                            const { id, name } = values['existing-playbook'];
+                            // eslint-disable-next-line camelcase
+                            api.patchRemediation(id, { add, auto_reboot: values['auto-reboot'] }, this.state.basePath)
+                            .then(() => this.resolver(this.state.deferred)(id, name, false));
+                        } else {
+                            // eslint-disable-next-line camelcase
+                            api.createRemediation({ name: values['select-playbook'], add, auto_reboot: values['auto-reboot'] }, this.state.basePath)
+                            .then(({ id }) => this.resolver(this.state.deferred)(id, values['select-playbook'], true));
```
Hyperkid123, line 184: I would also create constants for values like `select-playbook`. It will make the code bit safer and we can reuse them in the schema as well.

packages/remediations/src/newSteps/issueResolution.js
```diff
@@ -89,7 +92,6 @@ const IssueResolution = (props) => {
 
 IssueResolution.propTypes = {
     systems: propTypes.array,
-    issuesById: propTypes.object,
     issue: propTypes.object,
```
Hyperkid123, line 95: We should define some proper shape of object props.

packages/remediations/src/newSteps/review.js
```diff
@@ -0,0 +1,154 @@
...
+import useFormApi from '@data-driven-forms/react-form-renderer/dist/esm/use-form-api';
+import {
+    Text,
+    TextContent,
+    Stack,
+    StackItem
+} from '@patternfly/react-core';
+import { Table, TableVariant, TableHeader, TableBody, sortable } from '@patternfly/react-table';
+import { CloseIcon, ExclamationTriangleIcon, RedoIcon } from '@patternfly/react-icons';
+import { Button } from '@patternfly/react-core/dist/esm/components/Button/Button';
```
Hyperkid123, line 12: No need to import PF components directly from esm. Just use the shorthand as with the other react core imports.

packages/remediations/src/newSteps/review.js
```diff
@@ -0,0 +1,154 @@
...
+
+    const getResolution = issueId => {
+        const allResolutions = resolutions.find(r => r.id === issueId)?.resolutions || [];
+
+        if (allResolutions.length > 1)  {
+            if (formOptions.getState().values['manual-resolution'] && issueId in formOptions.getState().values['selected-resolutions']) {
+                return allResolutions.filter(r => r.id === formOptions.getState().values['selected-resolutions'][issueId]);
+            }
+
+            if (formOptions.getState().values['existing-playbook-selected']) {
```
Hyperkid123, line 34: Again we could use constants for the value names

packages/remediations/src/newSteps/issueResolution.js
```diff
@@ -66,7 +66,10 @@ const IssueResolution = (props) => {
                                     }
                                     title={resolution.description}
                                 >
-                                    <TextContent className="pf-u-pt-md">
+                                    <TextContent className="pf-u-pt-sm">
+                                        <Text className="ins-c-playbook-description">
+                                            {`Resolution from "${issue.id.split(/:|\|/)[1]}"`}
```
rvsia, line 71: You don't have to use template literals.

```
Resolution from "{issue.id....}"
```

packages/remediations/src/newSteps/issueResolution.scss
```diff
@@ -21,6 +21,10 @@
             }
         }
     }
+    p.ins-c-playbook-description {
+        margin-bottom: var(--pf-global--spacer--sm);
```
rvsia, line 25: You can use PF4 [utilities](https://www.patternfly.org/v4/utilities/text) `className="pf-u-mb-sm pf-u-font-size-sm"` instead of custom css.

packages/remediations/src/NewRemediationWizard.js
fhlavac: Sure, I'll finish this one and take a look :-)

packages/remediations/src/NewRemediationWizard.js
```diff
@@ -1,28 +1,37 @@
...
+import FetchError from './newSteps/fetchError';
+import Review from './newSteps/review';
+
+function createNotification (id, name, isNewSwitch) {
+    const verb = isNewSwitch ? 'created' : 'updated';
+
+    return {
+        variant: 'success',
+        title: `Playbook ${verb}`,
+        description: <span>You have successfully { isNewSwitch ? 'created' : 'updated' } <a href={ remediationUrl(id) } >{ name }</a>.</span>,
```
PanSpagetka, line 24: you are duplicating condition that you resolved line above.

packages/remediations/src/NewRemediationWizard.js
```diff
@@ -137,33 +140,54 @@ class RemediationWizard extends Component {
...
+                            id,
+                            resolution: values['manual-resolution'] ? this.getResolution(id)?.[0]?.id : undefined,
+                            systems: this.state.data.systems
+                        }));
+                        const add = { issues, systems: this.state.data.systems };
+                        if (values['existing-playbook-selected']) {
+                            const { id, name } = values['existing-playbook'];
+                            // eslint-disable-next-line camelcase
+                            api.patchRemediation(id, { add, auto_reboot: values['auto-reboot'] }, this.state.basePath)
+                            .then(() => this.resolver(this.state.deferred)(id, name, false));
```
PanSpagetka, line 180: Resolver doesn't take any parameters, so you shouldn't pass it. Also is `this.state.deferred` actually necessary? If not, we might get rid of it.

packages/remediations/src/newSteps/review.js
```diff
@@ -0,0 +1,154 @@
...
+    const selectedPlaybook = formOptions.getState().values['existing-playbook'];
+
+    useEffect(() => {
+        formOptions.change('auto-reboot', formOptions.getState().values['existing-playbook-selected'] && selectedPlaybook.auto_reboot);
+    }, []);
+
+    const getResolution = issueId => {
+        const allResolutions = resolutions.find(r => r.id === issueId)?.resolutions || [];
+
+        if (allResolutions.length > 1)  {
```
PanSpagetka, line 29: Is this condition necessary?

PanSpagetka (changes requested):

katierik:
UX looks good

rvsia (approved):
Code looks good to me! :+1:

PanSpagetka (approved):

PR #886, opened 2021-01-13
fix(Remediations): RHICOMPL-1124 available only for supported

akofink:
Remediations are only available for rules which have failed against a
supported OS/SSG configuration.

This ComplianceRemediationButton needs to be refactored IMO because it makes a lot of assumptions about the data based on the two states it can be in (either a systems list or a rules list). It works for the system details page (i.e. rules table) when `system.ruleObjectsFailed == [] && selectedRules == []`, and it works for the policy details page (i.e. systems table) when `system.ruleObjectsFailed.length > 0`. It'd be way better if we just pass in `selectedRules` in the latter case and stop using `ruleObjectsFailed` completely. Just some (frustrated) thoughts after working through this fix :)

As you can see from the fix itself, there's a lot of different checks and, well nonsense, we have to include to allow the button to work in both contexts, now checking if (1) a rule is selected, (2) is not compliant, (3) has a remediation available, and (4) is part of a supported report.

**Testing hints**:

Please check both pages (system details and policy details), for supported and unsupported systems/rules, for systems/rules with and without remediations available, and for systems/rules that have both passed and failed results.

For all these combinations, please check that the button enables/disables properly and that the correct issues are sent to the remediations service and are able to be saved to a playbook.

### DIFF
--- a/packages/inventory-compliance/src/ComplianceRemediationButton.js
+++ b/packages/inventory-compliance/src/ComplianceRemediationButton.js
@@ -30,7 +30,11 @@ class ComplianceRemediationButton extends React.Component {
     )
 
     rulesWithRemediations = (rules, system) => {
-        return rules.filter(rule => rule.remediationAvailable).map(
+        return rules.filter((rule) => (
+            rule.remediationAvailable &&
+            this.ruleProfile(rule, system).supported &&
+            rule.compliant === false
+        )).map(
             rule => this.formatRule(rule, this.ruleProfile(rule, system).refId, system.id)
         );
     }
@@ -53,9 +57,17 @@ class ComplianceRemediationButton extends React.Component {
         });
     }
 
-    noRemediationAvailable = () => {
+    remediationAvailable = () => {
         const { allSystems, selectedRules } = this.props;
-        return selectedRules.length === 0 && !allSystems.some((system) => system.ruleObjectsFailed.some((rule) => rule.remediationAvailable && system.supported));
+        let rules = selectedRules.length ? selectedRules : allSystems.flatMap((system) => system.ruleObjectsFailed);
+        return rules.some((rule) => (
+            rule.remediationAvailable &&
+            (
+                rule.profiles?.some((profile) => profile.supported) ||
+                allSystems.some((system) => this.ruleProfile(rule, system).supported)
+            ) &&
+            rule.compliant === false
+        ));
     }
 
     render() {
@@ -64,7 +76,7 @@ class ComplianceRemediationButton extends React.Component {
         return (
             <React.Fragment>
                 <RemediationButton
-                    isDisabled={ this.noRemediationAvailable() }
+                    isDisabled={ !this.remediationAvailable() }
                     onRemediationCreated={ result => addNotification(result.getNotification()) }
                     dataProvider={ this.dataProvider }
                 >


packages/inventory-compliance/src/ComplianceRemediationButton.js
```diff
@@ -53,9 +57,17 @@ class ComplianceRemediationButton extends React.Component {
         });
     }
 
-    noRemediationAvailable = () => {
+    remediationAvailable = () => {
```
akofink, line 60: I found this logic way more confusing for the negative case :) It's confusing even for the positive case.

packages/inventory-compliance/src/ComplianceRemediationButton.js
```diff
@@ -53,9 +57,17 @@ class ComplianceRemediationButton extends React.Component {
...
         const { allSystems, selectedRules } = this.props;
```
akofink, line 61: Providing some background here, in case you're unsure how the component works:

On system details, rules table:
`allSystems` always just contains the one system in question
`selectedRules` reflects what the user has selected in the rules table

On the policy details, systems table:
`allSystems` reflects what the user has selected in the systems table (misleading name; not _all_ systems, only selected)
`selectedRules` is always empty

packages/inventory-compliance/src/ComplianceRemediationButton.js
```diff
@@ -30,7 +30,11 @@ class ComplianceRemediationButton extends React.Component {
     )
 
     rulesWithRemediations = (rules, system) => {
-        return rules.filter(rule => rule.remediationAvailable).map(
+        return rules.filter((rule) => (
+            rule.remediationAvailable &&
+            this.ruleProfile(rule, system).supported &&
+            rule.compliant === false
```
akofink, line 36: Apparently it was possible before to try and remediate rules which passed :man_shrugging:

packages/inventory-compliance/src/ComplianceRemediationButton.js
```diff
@@ -53,9 +57,17 @@ class ComplianceRemediationButton extends React.Component {
         });
     }
 
-    noRemediationAvailable = () => {
+    remediationAvailable = () => {
         const { allSystems, selectedRules } = this.props;
-        return selectedRules.length === 0 && !allSystems.some((system) => system.ruleObjectsFailed.some((rule) => rule.remediationAvailable && system.supported));
```
akofink, line 58: Note here that `system.supported` is always `undefined`. `supported` is an attribute on `profile`.

packages/inventory-compliance/src/ComplianceRemediationButton.js
```diff
@@ -53,9 +57,17 @@ class ComplianceRemediationButton extends React.Component {
...
+        let rules = selectedRules.length ? selectedRules : allSystems.flatMap((system) => system.ruleObjectsFailed);
+        return rules.some((rule) => (
+            rule.remediationAvailable &&
+            (
+                rule.profiles?.some((profile) => profile.supported) ||
+                allSystems.some((system) => this.ruleProfile(rule, system).supported)
+            ) &&
+            rule.compliant === false
```
vkrizan, line 69: Nitpick: I'd recommend to move simpler condition at the top. I know that we do not program on Intel 8080 and this is a premature optimization. If we'd have lots of systems and rules, it could make an impact.

packages/inventory-compliance/src/ComplianceRemediationButton.js
```diff
@@ -30,7 +30,11 @@ class ComplianceRemediationButton extends React.Component {
     )
 
     rulesWithRemediations = (rules, system) => {
-        return rules.filter(rule => rule.remediationAvailable).map(
+        return rules.filter((rule) => (
+            rule.remediationAvailable &&
+            this.ruleProfile(rule, system).supported &&
```
vkrizan, line 35: Observation: I've checked out `ruleProfile()` and it pulls in a profile by referencing `ref_id` of a rule and its profile. That seems insufficient and could uncover some bugs. UUID from `id` field would be preferred.

vkrizan (approved):
The fix is concise and looks good. Thanks for tacking this!

I'll try to do some testing...

vkrizan:
Works great!

Interesting observation is that the performance of rule selection on system tab is a bit slow and concerning. Every time a rule is selected a big part gets re-rendered and the heavy function `toRulesArray` gets called several times and slows the whole thing. You can see that there is a bit of delay after (de)selecting a rule.

PR #887, opened 2021-01-14
New build process proposal

Hyperkid123:
This PR introduces new safer build output which will automatically use cjs/esm assets based on the built environment. This will make it safer to use absolute import paths and reduce bundle sizes.

```jsx
/** legacy */
import NotificationPortal from '@prefix/esm/NotificationPortal'
/** new. Import gets automaticaly resolved to esm/cjs modules */
import NotificationPortal from '@prefix/NotificationPortal'
``` 
It also removes any rollup build tools and relies purely upon babel for the build.

### Changes
- automatic module resolution
- uses just babel for build
- requires CSS files with the component (but still creates both CSS and SCSS files)
  - not CSS in JS, consuming apps will have to configure CSS and style loaders
- no longer builds UMD packages, nobody is ever going to use this library in the browser without bundlers
- absolute esm imports still work but cjs folder is gone
  - we could copy the cjs as well to prevent potential breaking changes but then we have duplications
  
@karelhala If you are OK with this I would mark this as a major release.


karelhala (approved):
Looking hella good! Let's merge this and see if anybody struggles with using it. Can you also add some README info how to properly import packages?

Hyperkid123:
> Looking hella good! Let's merge this and see if anybody struggles with using it. Can you also add some README info on how to properly import packages?

@karelhala ofc. I was hoping we could use our new docs to add a detailed migration guide. But we can have it here until it's avaiable.

One more question. Do we have a major release tag?

karelhala:
@Hyperkid123 no tag, but comment trigger.

karelhala (approved):

karelhala:
release major

PR #889, opened 2021-01-15
Show notification when create Remediaiton

PanSpagetka:
I am not very happy with this fix, but it works. The root cause of the problem is that deferred promise gets resolved, however I couldn't find where. 

https://issues.redhat.com/browse/RHCLOUD-10682

### DIFF
--- a/packages/remediations/src/RemediationButton.js
+++ b/packages/remediations/src/RemediationButton.js
@@ -38,7 +38,7 @@ class RemediationButton extends React.Component {
 
     onClick = () => {
         Promise.resolve(this.props.dataProvider())
-        .then(data => this.state.remediations.openWizard(data))
+        .then(data => this.state.remediations.openWizard({ ...data, onRemediationCreated: this.props.onRemediationCreated }))
         .then(result => result && this.props.onRemediationCreated(result));
     }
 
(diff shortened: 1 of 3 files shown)


Hyperkid123 (approved):

packages/remediations/src/RemediationButton.js
```diff
@@ -38,7 +38,7 @@ class RemediationButton extends React.Component {
 
     onClick = () => {
         Promise.resolve(this.props.dataProvider())
-        .then(data => this.state.remediations.openWizard(data))
+        .then(data => this.state.remediations.openWizard({ ...data, onRemediationCreated: this.props.onRemediationCreated }))
         .then(result => result && this.props.onRemediationCreated(result));
```
karelhala, line 42: We don't really need this line anymore, right?

karelhala (commented):
Looks good, the wizard is being reworked, we should also change the whole logic there. No need to use promises while the data are bing processed.

packages/remediations/src/RemediationButton.js
PanSpagetka: Yep. I think this logic can be simplified in new wizard

karelhala (approved):

PR #898, opened 2021-01-20
Fix inventory detail inifite load if no system found

karelhala:
### Broken inv detail when incorrect UUID

When navigating to inv detail with incorrect UUID the UI ends in infinite loading state, because entity is not found

### Broken ui upon navigating back to list

When user navigates back to inventory list the UI breaks because it thinks that the tags were loaded, but no array of tags is actually passed in


katerinapat (approved):
Looks good to me

Hyperkid123 (approved):

rvsia (approved):

PR #899, opened 2021-01-20
Convert remediation wizard to a functional component

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-11793

- converted NewRemediationWizard to functional component + added a helper class component to preserve the API
- moved duplicate logic to `./utils`
- fixed bug showing resolution steps also for issues already present in the selected playbook
- did some more code cleaning

### DIFF
--- a/packages/remediations/src/NewRemediationWizard.js
+++ b/packages/remediations/src/NewRemediationWizard.js
@@ -1,5 +1,5 @@
-import React, { Fragment, useEffect, useReducer } from 'react';
-import PropTypes from 'prop-types';
+import React, { useEffect, useState } from 'react';
+import propTypes from 'prop-types';
 import keyBy from 'lodash/keyBy';
 import transform from 'lodash/transform';
 import FormRenderer from '@data-driven-forms/react-form-renderer/dist/esm/form-renderer';
@@ -36,40 +36,26 @@ function createNotification (id, name, isNewSwitch) {
     };
 }
 
-const initialState = {
-    open: false,
-    schema: {},
-    errors: []
-};
-
-function reducer(state, action) {
-    switch (action.type) {
-        case 'update':
-            return {
-                ...state,
-                ...action.payload
-            };
-        default:
-            throw new Error();
-    }
-}
-
 const RemediationWizard = ({
-    isOpen,
+    setOpen,
     data,
     basePath
 }) => {
 
-    const [ state, dispatch ] = useReducer(reducer, initialState);
+    const [ state, setState ] = useState({ errors: [] });
 
     const getIssuesMultiple = (issuesById, resolutions) =>
         data.issues.map(issue => {
+            const issueResolutions = resolutions.find(r => r.id === issue.id).resolutions;
+            const { description, needs_reboot: needsReboot  } = issueResolutions?.[0] || {};
             return {
                 action: issuesById[issue.id].description,
                 systemsCount: issue.systems ? issue.systems.length : data.systems.length,
                 id: issue.id,
                 shortId: issue?.id?.split('|')?.slice(-1)?.[0] || issue.id,
-                alternate: resolutions.find(r => r.id === issue.id).resolutions?.length - 1
+                resolution: description,
+                needsReboot,
+                alternate: issueResolutions?.length - 1
             };
         }).filter(record => record.alternate > 0);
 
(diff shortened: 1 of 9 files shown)


packages/remediations/src/NewRemediationWizardHelper.js
```diff
@@ -0,0 +1,25 @@
+import React, { Component, Fragment } from 'react';
+import RemediationWizard from './NewRemediationWizard';
+
+class NewRemediationWizard extends Component {
```
rvsia, line 4: What is is the reason behind this component? And why is it a class component? :thinking:

packages/remediations/src/NewRemediationWizard.js
```diff
@@ -138,63 +91,109 @@ class RemediationWizard extends Component {
...
+                    closeWizard();
+                }}
+                onCancel={closeWizard}
+            />
+            : <Fragment/>
+    );
+};
+
+RemediationWizard.propTypes = {
+    isOpen: PropTypes.bool,
```
rvsia, line 194: This is a little bit strange. You pass `isOpen` in props, but after that, you store it in state. Why not use `isOpen` from the props all the time?

packages/remediations/src/NewRemediationWizard.js
```diff
@@ -138,63 +91,109 @@ class RemediationWizard extends Component {
...
+                    ...mapperExtension
+                }}
+                onSubmit={(_, formOptions) => {
+                    onSubmit(formOptions.getState().values);
+                    closeWizard();
+                }}
+                onCancel={closeWizard}
+            />
+            : <Fragment/>
+    );
```
rvsia, line 190: Showing an empty fragment is unnecessary. You can use `null`.

You can also optimize it a little if you move the condition right after the hook.

```jsx
useEfffect(...);

if(!isOpen) { 
  return null;
}
```

So the code won't go over the mapper and functions definitions.

packages/remediations/src/NewRemediationWizard.js
```diff
@@ -138,63 +91,109 @@ class RemediationWizard extends Component {
...
+                    [MANUAL_RESOLUTION]: true,
+                    [SELECTED_RESOLUTIONS]: {},
+                    [EXISTING_PLAYBOOK_SELECTED]: false
+                }}
+                componentMapper={{
+                    [componentTypes.WIZARD]: Wizard,
+                    ...mapperExtension
+                }}
+                onSubmit={(_, formOptions) => {
+                    onSubmit(formOptions.getState().values);
```
rvsia, line 184: You get values as the first argument (`_`)

packages/remediations/src/NewRemediationWizard.js
```diff
@@ -34,93 +36,44 @@ function createNotification (id, name, isNewSwitch) {
...
-    constructor() {
-        super();
-        this.container = React.createRef(document.createElement('div'));
+function reducer(state, action) {
+    switch (action.type) {
+        case 'update':
+            return {
+                ...state,
+                ...action.payload
+            };
```
rvsia, line 51: Not sure if using reducer like this makes sense when there is just on action called 'update'.

You can achieve the same by using:

```jsx
const [state, setState] = useState({ ... });

...

setState({ ...state, newValues })
```

Or please split the reducer to more actions. For example instead of 

`const closeWizard = () => dispatch({ type: 'update', payload: { isOpen: false } });`

create a new action

```jsx
        case 'close':
            return {
                ...state,
                isOpen: false
            };
```

`const closeWizard = () => dispatch({ type: 'close' });`

packages/remediations/src/newSteps/reviewActions.js
```diff
@@ -1,56 +1,34 @@
...
-                title: record.needsReboot ? <Fragment><RedoIcon/>{' Yes'}</Fragment> : <Fragment><CloseIcon/>{' No'}</Fragment>,
-                value: record.needsReboot
-            },
-            record.systemsCount
-        ]
-    }));
-
-    const onSort = (event, index, direction) => setSortByState({ index, direction });
+    const records = formOptions.getState().values[EXISTING_PLAYBOOK_SELECTED]
+        ? issuesMultiple.filter(issue => !formOptions.getState().values[EXISTING_PLAYBOOK].issues.some(i => i.id === issue.id))
```
rvsia, line 28: You can store values in a const to not run `getState` function 2x.

PanSpagetka (approved):
When you resolve Richards comments, I think this look good

packages/remediations/src/NewRemediationWizardHelper.js
fhlavac: It's there to keep the original API and the OpenWizard method. It is rather a temporary solution.

packages/remediations/src/NewRemediationWizard.js
fhlavac: sure, thanks for catching that :)

packages/remediations/src/NewRemediationWizard.js
fhlavac: But in the first argument, there are not all the values, just these mentioned in the schema :/

rvsia: Yep, there are only values that you visited in the wizard.

fhlavac: I don't see values that I set manually using formOptions.change()
Is it ok to use the second param or can I add it somehow to the first one?

rvsia: As we talked in DMs, the only concern is too handle values that users do not visit. If you have managed it elsewhere, then it's no issue to use the function.

rvsia (approved):

PR #904, opened 2021-01-21
Hot fix notifications by adding actions exporter

karelhala:
### Broken builds with older version

Since we want to maintain slow and incremental upgrade we shouldn't break how actions were previously imported. This PR fixes potentional problems by re-exporting redux partials under specific filenames.

### DIFF
--- a/packages/notifications/src/NotificationPortal/NotificationPortal.test.js
+++ b/packages/notifications/src/NotificationPortal/NotificationPortal.test.js
@@ -19,7 +19,7 @@ describe('Notification portal', () => {
         };
     });
 
-    it('should return no component when no notifications given', () => {
+    it.only('should return no component when no notifications given', () => {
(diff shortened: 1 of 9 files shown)


Hyperkid123 (commented):
@karelhala tests are dead. Could you also add a deprecation warning to these files? Ultimately we want people to use the new paths.

packages/notifications/src/NotificationPortal/NotificationPortal.test.js
```diff
@@ -19,7 +19,7 @@ describe('Notification portal', () => {
         };
     });
 
-    it('should return no component when no notifications given', () => {
+    it.only('should return no component when no notifications given', () => {
```
Hyperkid123, line 22: `only` leftover

Hyperkid123 (approved):

PR #908, opened 2021-01-25
Do not log error when copying styles to utilities folder

karelhala:


Hyperkid123:
@karelhala why would we want to hide the error?

karelhala:
@Hyperkid123 the copy was breaking and did not copyied over the files. We are not logging error for any of these folders and it should break the build if the copy fails instead of just logging it.

Hyperkid123:
Right, it was not throwing the error further. Gotcha.

Hyperkid123 (approved):

PR #917, opened 2021-01-28
chore(deps): Update apollo boost

bastilian:
Updates the dependency to match what is in compliance-frontend and also resolves the circular dependency issue rollup mentions.


akofink (approved):
Makes sense. Nice refactor :+1:

PR #919, opened 2021-01-28
feat: Introduce TableTools

bastilian:
This introduces a collection of hooks aka. TableTools to the frontend-components-utilities package to control a PatternFly table and/or the Insights Frontend Components `PrimaryToolbar` component.

It allows passing in an array of items and a set of columns as well as filters. Hooks are meant to be usable in combination with the `useTableTools` hook as well as separately (they may not fully yet). There is also a `TableToolsTable` component that is already set up to use the hooks, an example for a use of it can be found in compliance-frontend[0]. 
The resulting table looks and works like this:

https://user-images.githubusercontent.com/7757/116374906-0a8e3600-a80f-11eb-9b18-e5fc7d0d821f.mp4

[0] https://github.com/RedHatInsights/compliance-frontend/pull/1192/files#diff-042375a49456c92f5d5f42c19fdaf1e81fd7156fe94eb71e8691eadf78ff2d02

### DIFF
--- a/packages/utils/src/useTableTools/FilterConfigBuilder/ChipBuilder.js
+++ b/packages/utils/src/useTableTools/FilterConfigBuilder/ChipBuilder.js
@@ -1,0 +1,52 @@
+import { conditionalFilterType } from '@redhat-cloud-services/frontend-components/ConditionalFilter';
+
+class ChipBuilder {
+    constructor(filterConfig) {
+        this.filterConfig = filterConfig;
+        this.config = this.filterConfig.config;
+    }
+
+    textFilterChip = (category, currentValue) => (currentValue && currentValue !== '' ? {
+        category,
+        chips: [{ name: currentValue }]
+    } : null)
+
+    checkboxFilterChip = (category, currentValue) => (currentValue && currentValue.length > 0 ? {
+        category,
+        chips: currentValue.map((value) => (
+            { name: this.filterConfig.labelForValue(value, category) }
+        ))
+    } : null)
+
+    radioFilterChip = (category, currentValue) => (currentValue && currentValue !== '' ? {
+        category,
+        chips: [{ name: this.filterConfig.labelForValue(currentValue, category) }]
+    } : null)
+
+    chipFor = (filter, currentValue) => {
+        const categoryConfig = this.filterConfig.getCategoryForLabel(filter);
+        const { label, type } = categoryConfig ? categoryConfig : { label: filter, type: null };
+
+        switch (type) {
+            case conditionalFilterType.text:
+                return this.textFilterChip(label, currentValue);
+
+            case conditionalFilterType.checkbox:
+                return this.checkboxFilterChip(label, currentValue);
+
+            case conditionalFilterType.radio:
+                return this.radioFilterChip(label, currentValue);
+
+            default:
+                return null;
+        }
+    }
+
+    chipsFor = (filters) => (
+        Object.keys(filters).map((filter) => (
+            this.chipFor(filter, filters[filter])
+        )).filter((f) => (!!f))
+    )
+}
+
+export default ChipBuilder;
--- a/packages/utils/src/useTableTools/useFilterConfig.js
+++ b/packages/utils/src/useTableTools/useFilterConfig.js
@@ -1,0 +1,67 @@
+import { useState, useLayoutEffect } from 'react';
+import FilterConfigBuilder from './FilterConfigBuilder/FilterConfigBuilder';
+
+const filterConfigBuilder = new FilterConfigBuilder([]);
+
+const useFilterConfig = (initialConfig) => {
+    const [ activeFilters, setActiveFilters ] = useState(filterConfigBuilder.initialDefaultState(
+        initialConfig?.activeFilters || []
+    ));
+    const onFilterUpdate = (filter, value) => (
+        setActiveFilters({
+            ...activeFilters,
+            [filter]: value
+        })
+    );
+
+    const buildConfig = () => (
+        filterConfigBuilder.buildConfiguration(
+            onFilterUpdate,
+            activeFilters
+        )
+    );
+
+    const addConfigItem = (item) => {
+        filterConfigBuilder.addConfigItem(item);
+    };
+
+    const clearAllFilter = () => (
+        setActiveFilters(filterConfigBuilder.initialDefaultState())
+    );
+
+    const deleteFilter = (chips) => (
+        setActiveFilters(filterConfigBuilder.removeFilterWithChip(
+            chips, activeFilters
+        ))
+    );
+    const onFilterDelete = (_event, chips, clearAll = false) => (
+        clearAll ? clearAllFilter() : deleteFilter(chips[0])
+    );
+    const chipBuilder = filterConfigBuilder.getChipBuilder();
+    const filterChips = chipBuilder.chipsFor(activeFilters);
+
+    useLayoutEffect(() => {
+        initialConfig.forEach(addConfigItem);
+    }, []);
+
+    return {
+        filter: (items) => (
+            filterConfigBuilder.applyFilterToObjectArray(
+                items, activeFilters
+            )
+        ),
+        toolbarProps: {
+            filterConfig: buildConfig(),
+            activeFiltersConfig: {
+                filters: filterChips,
+                onDelete: onFilterDelete
+            }
+        },
+        addConfigItem,
+        filterConfigBuilder,
+        activeFilters,
+        buildFilterString: () => filterConfigBuilder.getFilterBuilder().buildFilterString(activeFilters)
+    };
+};
+
+export default useFilterConfig;
--- a/packages/utils/src/useTableTools/useTableTools.js
+++ b/packages/utils/src/useTableTools/useTableTools.js
@@ -1,0 +1,86 @@
+import { useEffect } from 'react';
+import useFilterConfig from './useFilterConfig';
+import useTableSort from './useTableSort';
+import { default as usePaginate } from './usePaginate';
+import useRowsBuilder from './useRowsBuilder';
+import useBulkSelect from './useBulkSelect';
+import useRowIdentify from './useRowIdentify';
+import useConditionalTableHook from './useConditionalTableHook';
+
+const useTableTools = (items = [], columns = [], options = {}) => {
+    const enableFilters = !!options.filterConfig;
+    const enableBulkSelect = !!options.onSelect;
+    const enablePagination = options?.pagination !== false;
+    const perPage = options?.perPage || 10;
+
+    const identify = useRowIdentify(options);
+    const identifiedItems = identify(items);
+
+    const {
+        tableProps: sortableTableProps, columns: sortableColumns, sorter
+    } = useTableSort(columns, options);
+
+    const {
+        filter, toolbarProps: conditionalFilter, activeFilters, addConfigItem: addFilterConfigItem
+    } = useConditionalTableHook(enableFilters, useFilterConfig, options.filterConfig);
+
+    const paginate = useConditionalTableHook(enablePagination, usePaginate, {
+        ...options,
+        perPage,
+        itemCount: items.length
+    });
+
+    const {
+        transformer: selectItem, toolbarProps: bulkSelectToolbarProps,
+        tableProps: bulkSelectTableProps, clearSelection
+    } = useConditionalTableHook(enableBulkSelect, useBulkSelect, {
+        ...options,
+        items: identifiedItems,
+        filter,
+        addFilterConfigItem,
+        paginate,
+        sorter
+    });
+
+    const transformers = [
+        selectItem
+    ];
+
+    const rowBuilder = useRowsBuilder(columns, {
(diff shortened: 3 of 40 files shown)


packages/inventory-compliance/src/Utilities/hooks/TableTools/useTableTools.js
```diff
@@ -0,0 +1,70 @@
+import { useEffect } from 'react';
+import useFilterConfig from './useFilterConfig';
+import useTableSort from './useTableSort';
+import usePaginate from './usePaginate';
+import useRowsBuilder from './useRowsBuilder';
+import useBulkSelect from './useBulkSelect';
+
+export const useIdentify = (items, options = {}) => {
```
bastilian, line 8: I wonder if this would make sense to be implemented as/with and/or even expose it to have other hooks could rely on that to manage items.

packages/inventory-compliance/src/Utilities/hooks/TableTools/useBulkSelect.js
```diff
@@ -0,0 +1,128 @@
...
+        setPageItems,
+        tableProps: {
+            onSelect: selectOne,
+            canSelectAll: false
+        },
+        toolbarProps: {
+            bulkSelect: {
+                toggleProps: { children: [ title, ' ' ] },
+                // TODO: These should be updated when the page conditions change
+                // like perPage, page.selected or the item count change
```
bastilian, line 110: If `setCurrentPage` is called it should update this, should it not?

packages/inventory-compliance/src/Utilities/hooks/TableTools/useTableSort.js
```diff
@@ -0,0 +1,53 @@
...
+    })
+);
+
+export const orderByArray = (objectArray, orderProp, orderArray, direction) => (
+    (direction !== 'asc' ? orderArray.reverse() : orderArray).flatMap((orderKey) => (
+        objectArray.filter((item) => (item[orderProp] === orderKey))
+    ))
+);
+
+const useTableSort = (items, columns) => {
```
bastilian, line 27: This can probably be done or enhanced with using the react-table transform.

packages/inventory-compliance/src/Utilities/hooks/TableTools/TableToolTable.js
```diff
@@ -0,0 +1,63 @@
...
+};
+
+/* eslint-disable react/display-name */
+export const exampleColumns = [
+    {
+        title: 'Name',
+        prop: 'name',
+        transforms: [ sortable ],
+        sortByProperty: 'name',
+        renderFunc: (name) => (
```
bastilian, line 27: This is inspired by the Inventory table, but I thought about actually naming it `component` as in most cases where the Cell might have more complex content a component would be better to use and the naming would encourage that, while still allowing to pass a function (component).

akofink:
Could you please fix the linting issues so this can be built without disabling eslint? :pray:

bastilian:
@akofink fixed!

akofink (commented):
The code seems fine. It's nice to have all the concerns split into their separate files. It's pretty hard to review nearly 2k lines, though, without any easy way to test it or having it split into digestible sized commits. I tried to use it in https://github.com/RedHatInsights/compliance-frontend/pull/1053 (see screenshots on that PR) but couldn't quite figure out how it replaces our current SystemRulesTable (not quite a drop-in replacement, it seems, even though the props seem similar). Thanks for including the ExampleTable component. How difficult would it be to submit a compliance-frontend PR that swaps our current SystemRulesTable for this implementation? IMO that'd help tremendously with review, even if we don't end up merging that compliance-frontend PR.

bastilian:
@akofink It is already replacing fully replacing the `SystemsRulesTable`[0], so it should only be needed to (npm) link the inventory-compliance package in the compliance-frontend and it will use the new table implementation. 
I left out removing the SystemRulesTable component out of this PR for now, to not have even more lines and to still have it as a reference till we are sure we got everything in the reimplementation.

I'll try to add split it up into at least two commits one introducing the utilities and the other rebuilding the table.

[0] https://github.com/RedHatInsights/frontend-components/pull/919/files#diff-c631c298f11e6d6c0f10e6eac271e13c676385d9281cca190cce5ab6f829997dR2

packages/utils/src/useTableTools/FilterConfigBuilder/ChipBuilder.js
```diff
@@ -0,0 +1,52 @@
+import { conditionalFilterType } from '@redhat-cloud-services/frontend-components/ConditionalFilter';
```
bastilian, line 1: All of these "FilterConfigBuilder" classes are moved over from the inventory-compliance package, from where it is used in all tables in Compliance with filters: https://github.com/RedHatInsights/frontend-components/tree/master/packages/inventory-compliance/src/Utilities

packages/utils/src/useTableTools/useFilterConfig.js
```diff
@@ -0,0 +1,57 @@
+import { useState } from 'react';
```
bastilian, line 1: This hook is from compliance-frontend: https://github.com/RedHatInsights/compliance-frontend/blob/master/src/Utilities/hooks/useFilterConfig.js

packages/inventory-compliance/src/index.js
```diff
@@ -1,5 +1,5 @@
 export { default as default } from './Compliance';
-export { default as SystemRulesTable, selectColumns as selectRulesTableColumns } from './SystemRulesTable';
+export { default as SystemRulesTable, selectColumns } from './Components/RulesTable';
```
bastilian, line 2: This replaces the current `SystemRulesTable`, the PR does not for now remove the actual files from the old rules table.

packages/inventory-compliance/src/Components/RulesTable/RuleDetailsRow.js
```diff
@@ -0,0 +1,69 @@
...
+            <Stack id={ `rule-identifiers-references-${key}` } className='margin-bottom-lg'>
+                <Grid>
+                    { identifier && identifier.length > 0 && <GridItem span={ 2 }>
+                        <Text className='pf-c-form__label' component={ TextVariants.h5 }><b>Identifier</b></Text>
+                        <Text>
+                            {
+                                identifier.map((ident, idx) => (
+                                    <ConditionalLink
+                                        href={ ident.system }
+                                        target='_blank'
```
vkrizan, line 27: For the `_blank` target the `rel` has to be set to `noopener noreferrer` to prevent XSS. See [CWE-79](https://cwe.mitre.org/data/definitions/79.html).

packages/inventory-compliance/src/Components/RulesTable/index.js
```diff
@@ -0,0 +1,157 @@
...
+    loading: propTypes.bool,
+    hidePassed: propTypes.bool,
+    system: propTypes.object,
+    remediationsEnabled: propTypes.bool,
+    tailoringEnabled: propTypes.bool,
+    selectedRefIds: propTypes.array,
+    selectedFilter: propTypes.bool,
+    handleSelect: propTypes.func,
+    onSelect: propTypes.func,
+    columns: propTypes.arrayOf(
```
vkrizan, line 137: Could the columns be refactored to select columns only without any further configuration? It could prevent calling `selectColumns` by the component user.
```
columns: propTypes.arrayOf(propTypes.string)
```

packages/inventory-compliance/src/Components/RulesTable/index.js
```diff
@@ -0,0 +1,157 @@
...
+        sortByProperty: 'remediationAvailable',
+        renderFunc: (rule) => (
+            <Ansible { ...{ rule } } />
+        )
+    }
+];
+/* eslint-enable */
+
+export const selectColumns = (columnTitles) => (
+    columns.filter((column) => columnTitles.includes(column.original || column.title))
```
vkrizan, line 60: Observation: There is a column matching by `original` and `title` used through the code. Could it use one kind, having `original` as primary? Preferably the `original` could be named `name` to provide a programmatic naming to a column.

Currently the column matching is mostly coupled with a visual title. If the title is changed (as we've seen with "Rule" -> "Name") the whole usage has to change with it. Having a single and stable naming would prevent visual changes affecting the column selection/matching.

packages/utils/src/useTableTools/TableToolTable.js
```diff
@@ -0,0 +1,63 @@
...
+            <TableHeader />
+            <TableBody />
+        </Table>
+        <TableToolbar isFooter>
+            <Pagination { ...toolbarProps.pagination } variant={ PaginationVariant.bottom } />
+        </TableToolbar>
+    </React.Fragment>;
+};
+
+export default ExampleTable;
```
vkrizan, line 63: Observation: File could be renamed to `ExampleTable.js` to easily find it.

packages/utils/src/useTableTools/useTableTools.js
```diff
@@ -0,0 +1,83 @@
...
+        page,
+        perPage,
+        currentPageItemIds: []
+    });
+    const { items: identifiedItems } = useIdentify(items, options);
+    const {
+        items: selectedItems,
+        toolbarProps: bulkSelectToolbarProps,
+        tableProps: bulkSelectTableProps
+    } = options.selectable ? useBulkSelect(identifiedItems, options, [ tableState, setTableState ]) : { items: identifiedItems };
```
vkrizan, line 39: A hook within a condition. Wouldn't this crash in some conditions, breaking the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html)?

vkrizan (commented):
Great improvement to move the common stuff to a library!

As already mentioned, this contains too many changes all at once. It's hard to find any issues. I've tried my best to find potential gaps. It would be a good practice to separate it to a several smaller changes, i.e. splitting creation of the library (the new hook) and the actual usage changes.

packages/utils/src/useTableTools/useTableTools.js
bastilian: You are totally right, I forgot about that bit. I'm also not happy with the way the hooks are called here, since they somewhat depend on a certain order. I'll look over this again and try to refactor that part.

packages/utils/src/useTableTools/TableToolTable.js
bastilian: I'll go for `TableToolTable.js`. It could also be used as a base for starting a custom table component and importing ExampleTable, might discourage that.

packages/inventory-compliance/src/Components/RulesTable/index.js
bastilian: Yeah, @akofink pointed these out as well, somewhere similar, I can't remember where, but essentially we'll move from using a string to have cell object to import and pass in in the place where it is used, this will also allow to extend these "cell objects" as needed with custom props and `selectColumn` will be obsolete.

akofink:
@bastilian I did test this yesterday without the compliance-frontend test PR, and it behaved identically to the existing implementation, partly why I was confused :) So great work! I'll do another test today to be sure (after rebase), but this seems to work identically to the existing implementation. :clap:

bastilian:
@akofink It should not look identical but have the dropdown to select:

Screenshot 2021-03-17 at 14 56 00

The new rules table is now exported from and as `SystemsRulesTable`, which should allow to test the table properly with the compliance-frontend.

akofink:
Thanks for the clarification, @bastilian. I can see the new select all component now; it doesn't work the same as the existing select all: it only acts on the page rather than the entire filter, but that's consistent with the way it works on the inventory table. Still, I wonder why there's not a 'Select all across pages' or similar. Like 'Select all on page' and 'Select all 153'.

I also noticed the policy wizard rules are now no longer selectable:

image

akofink:
Sorry :P I do see a 'Select all X' after all, but the X is not the same as the filtered total:

image

Shouldn't it match the total number shown in the top right of the table?

bastilian:
> Shouldn't it match the total number (12 in this case) shown in the top right of the table?

Yeah, you are right that should actually be the filtered amount! I'll update that.

akofink:
The remediation button only considers the current page to enable/disable. The screenshot below shows 43 selected rules, some with remediations available, and none selected on the current page:

image

akofink:
The policy filter on the system details page does not appear anymore when there are > 1 policies for the system:

image

akofink:
The wizard does not show the selected canonical rule by default anymore. Also, this table used to 'filter by selected' by default and no longer seems to include that default filter anymore:

image

akofink (commented):
@bastilian I think we all like the design of this work and have mostly approved the code. Now it's just a matter of identifying and fixing any issues - sorry for the many screenshot comments. Let me know if you cannot reproduce something I'm seeing.

bastilian:
@akofink Thank you for testing and the screenshot! I shouldn't have any issue reproducing them and addressing them.

bastilian:
Closing. We'll be moving them into compliance-frontend. See https://github.com/RedHatInsights/compliance-frontend/pull/1208

PR #920, opened 2021-01-28
Structured docs nav

Hyperkid123:
Components nav is now grouped into sections based on the folder the file is placed in.

screenshot-localhost_3000-2021 01 28-10_40_46


packages/docs/docs-generator.js
```diff
@@ -190,13 +190,31 @@ ${extensiveProps.map((data) => {
...
+        if (!Object.prototype.hasOwnProperty.call(nav, groupName)) {
+            nav[groupName] = [];
+        }
+
+        nav[groupName].push(navName);
+    });
+    return Object.entries(nav).map(([ key, value ]) => ({
+        group: key,
+        items: value.sort((a, b) => a.localeCompare(b))
+    })).sort((a, b) => a.group.localeCompare(b.group));
```
karelhala, line 208: This looks overcomplicated, plus it looks just like reduce function
```JS
Object.keys(components).reduce((nav, path) => {
  let fragments = path.split('/');
  const navName = fragments.pop().replace('.js', '');
  const groupName = fragments.pop();
  return {
    ...nav,
    [groupName]: [
      ...nav[groupName] || [],
      navName
    ]
  }
}, {});
```

karelhala (approved):

PanSpagetka (approved):

PR #924, opened 2021-01-28
fix(components): removed unused PF css import

Hyperkid123:
`@patternfly/patternfly/components/FormControls/form-control.css` CSS file is no longer provided in the current PF version and it is killing builds when trying to build the CSS.


AllenBW (approved):
YESSSSSSSSSS 🆗

karelhala (approved):

PR #925, opened 2021-01-28
Inventory general info v3

Hyperkid123:
Again this will require manual release

### TODO
- [x] test in apps

### v2(left)/v3(right) side by side comparision
Screenshot from 2021-01-28 16-18-51


rvsia:
image

packages/inventory-general-info/.gitignore
```diff
@@ -1,6 +1,21 @@
...
+!.gitignore
+!babel.config.js
+!LICENSE
+!package.json
+!README.md
+!tsconfig.json
+!pre-publish.js
+
+# Incremental ts info
+tsconfig.tsbuildinfo
```
rvsia, line 21: I am started feeling this should be a common file. :thinking:

Hyperkid123: Yep. Once we migrate I will create a common gitignore for all packages. I am adding some entries in each migration. Same goes for the babel config and some scripts.

rvsia (approved):

PR #942, opened 2021-02-04
Update Utils usage in Sources to V3

rvsia:


packages/sources/package.json
```diff
@@ -26,7 +26,6 @@
     },
     "homepage": "https://github.com/RedHatInsights/frontend-components/tree/master/packages/sources#readme",
     "dependencies": {
-        "@redhat-cloud-services/frontend-components": "*",
```
Hyperkid123, line 29: Is this package not using frontend components at all? I think you should also run install in the root to update the lock file.

rvsia: @Hyperkid123  there is no change to package-lock.

Hyperkid123 (approved):

PR #943, opened 2021-02-04
Charts v3

Hyperkid123:
@karelhala I know this is a deprecated package. Should we also add a message for devs to stop using it?

### TODO
- [x] use v3 build
- [x] test in app


karelhala:
@Hyperkid123 no need to, it's astill a good package, we could potentially use this package as base for PDF charts. We'll see.

karelhala (approved):

Hyperkid123:
@karelhala ok. We can run autorelease on this.

Hyperkid123:
release major

PR #945, opened 2021-02-04
Add action type import example to read me

Hyperkid123:


karelhala (approved):

rvsia (approved):

rvsia:
@Hyperkid123 can you also fix `import NotificationPortal from '@redhat-cloud-services/frontend-components-notifications/NotificationPortal';` in the same file? I think it should be `import NotificationsPortal from '@redhat-cloud-services/frontend-components-notifications/NotificationPortal';`

Hyperkid123:
@rvsia I can't tell what is the difference between these two imports. :eyes:

rvsia:
@Hyperkid123  oh you are right, the example for v2 is wrong `import NotificationsPortal from '@redhat-cloud-services/frontend-components-notifications/esm/NotificationsPortal';` but that doesn't matter, sorry!

PR #952, opened 2021-02-05
fix(SystemPolicyCard): RHICOMPL-1162 - Use DateFormat for last scanned

bastilian:
**Before:**
Screenshot 2021-02-05 at 13 56 12

**After:**

Screenshot 2021-02-05 at 13 56 05

**Note:** Test via LOCAL_CHROME + (re)building + inventory-compliance package linked

### DIFF
--- a/packages/inventory-compliance/src/SystemPolicyCard/SystemPolicyCard.js
+++ b/packages/inventory-compliance/src/SystemPolicyCard/SystemPolicyCard.js
@@ -1,5 +1,5 @@
-import React, { Fragment } from 'react';
-import { FormattedRelative, FormattedRelativeTime } from 'react-intl';
+import React from 'react';
+import DateFormat from '@redhat-cloud-services/frontend-components/DateFormat';
 import { CheckCircleIcon, ExclamationCircleIcon, OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
 import Truncate from 'react-truncate';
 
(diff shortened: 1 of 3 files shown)


packages/inventory-compliance/src/SystemPolicyCard/SystemPolicyCard.js
```diff
@@ -1,5 +1,5 @@
-import React, { Fragment } from 'react';
-import { FormattedRelative, FormattedRelativeTime } from 'react-intl';
+import React from 'react';
+import DateFormat from '@redhat-cloud-services/frontend-components/DateFormat';
```
vkrizan, line 2: Observation: I'm not fully sure whether this could use the `/esm` variant.

/cc @karelhala

vkrizan (approved):
Looks good. Thank you.

packages/inventory-compliance/src/SystemPolicyCard/SystemPolicyCard.js
bastilian: @Hyperkid123 Is working on eliminating the need for that as far as i know.

Hyperkid123: If a package and app is using v3 of the components dependency, there is no need for the /esm partial.

PR #955, opened 2021-02-08
fix wording and colors in error state

ryelo:


epwinchell (approved):

packages/components/src/ErrorState/ErrorState.js
```diff
@@ -11,11 +11,12 @@ import {
     EmptyStateBody,
     Button
 } from '@patternfly/react-core';
+import './filter-chips.scss';
```
karelhala, line 14: This file does not exists.
```suggestion
import './error-state.scss';
```

PR #962, opened 2021-02-12
Change "Clear filters" to "Reset filters" according to UXD

katerinapat:
Change wording according to 
https://docs.google.com/document/d/1lwG_D41-KhEBjZPxJvbOFKkpwmTN9FmWC7krkMrYOZI/edit#

### DIFF
--- a/packages/components/src/FilterChips/FilterChips.js
+++ b/packages/components/src/FilterChips/FilterChips.js
@@ -40,7 +40,7 @@ const FilterChips = ({ className, filters, onDelete }) => {
                     </Chip>
                 </ChipGroup>
             )) }
-            { filters.length > 0 && <Button variant="link" onClick={ (event) => onDelete(event, filters, true) }>Clear filters</Button> }
+            { filters.length > 0 && <Button variant="link" onClick={ (event) => onDelete(event, filters, true) }>Reset filters</Button> }
         </span>
     );
 };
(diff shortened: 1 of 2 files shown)


bastilian:
Wouldn't "Reset" mean to go back to the initial/default filters rather than clearing them?

katerinapat:
> Wouldn't "Reset" mean to go back to the initial/default filters rather than clearing them?

@bastilian exactly. Based on the documents from UXD all the apps should adapt this behavior.

karelhala:
@katerinapat we've had a few days back discussion with UXers about it as well, the problem is that upon clicking on "Reset filters" the filter has to go back to original state. So apps has to adapt to this behavior once they start using the reset, moreover not all filters has default values. For instance global filter doesn't really have default state and there should still be "Remove filters". Hence I wouldn't just change the string in here, but rather pass in new prop `clearTitle` which would be a node propType and by default it would be `Remove filter` (we want to make it backward compatible).

mkholjuraev:
I am wondering if there is an agreement on what to do. We have a blocked ticket associated with changing 'Clear Filters' to 'Reset filters' which is done in this PR.

bastilian:
@mkholjuraev I think what needs to be done for this is to make sure that all behaviours align with the wording in all places. The `FilterChips` component is used in a few applications and other components where this might not be the case yet. 

Maybe the change in this PR should just allow to pass in or choose a wording?

karelhala:
Closing this in favor of https://github.com/RedHatInsights/frontend-components/pull/1035, we want to preserve current behavior and allow consumers to pass in custom delete string. This is because some tables are not ready to reset filters or will never be.

PR #963, opened 2021-02-12
chore(docs): ignore docs package from "start" script

Hyperkid123:
The docs "start" command should not trigger. It starts server instead of watching build files.


bastilian (approved):
Works for me! Thank you @Hyperkid123!

rvsia (approved):

PR #967, opened 2021-02-15
Add temporary beta rewrite.

Hyperkid123:
There is a bug with the Insights proxy. It serves HTML (and other static assets by extensions) to beta env, even though webpack deployment is not set to `beta`.

However, when making an XHR request for the `fed-mods.json` it responds as 404 (which it should). This PR adds a temporary proxy rewrite from beta to stable with a warning and link to an example of how to use beta env properly.

The proxy rewrite can be turned off.

part of: https://github.com/RedHatInsights/insights-chrome/issues/1199

### DIFF
--- a/packages/config/src/config.js
+++ b/packages/config/src/config.js
@@ -1,6 +1,7 @@
 /* eslint-disable camelcase */
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const path = require('path');
+let rewriteLineCounter = 0;
 
 module.exports = ({
     port,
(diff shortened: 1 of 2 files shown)


packages/config/src/config.js
```diff
@@ -142,7 +144,24 @@ module.exports = ({
             inline: true,
             disableHostCheck: true,
             historyApiFallback: true,
-            writeToDisk: true
+            writeToDisk: true,
+            proxy: enableBetaRewrite ? {
+                'https://ci.foo.redhat.com:1337/beta': {
```
karelhala, line 149: What about qa and prod? I think it would be more reasonable to pass in `betaEnv` which would default to `ci` and if it is defined use this proxy.

Hyperkid123: Forgot about the rest

karelhala (approved):

PR #968, opened 2021-02-15
Add group name in the pdf document

katerinapat:
image


karelhala (approved):

packages/pdf-generator/src/components/PDFDocument.js
```diff
@@ -38,6 +39,9 @@ const PDFDocument = ({
                         </Text>
                     </View>
                     <View style={appliedStyles.reportNameWrapper}>
+                        <Text style={ appliedStyles.groupName }>
```
AllenBW, line 42: Should this be conditionally rendered on the presence or absence of `groupName` ?

PR #969, opened 2021-02-15
Fix invalid prop types and invalid keys.

Hyperkid123:
Fixes multiple react warnings at runtime

A possible breaking change is replacing Fragment with span to remove the invalid prop warning. I have tried clicking in the chrome global filter and it was working. But if someone could double-check it would be grand.

### DIFF
--- a/packages/charts/src/SeverityLine/SeverityLine.js
+++ b/packages/charts/src/SeverityLine/SeverityLine.js
@@ -45,9 +45,9 @@ const SeverityLine = ({ title, value, className, tooltipMessage, config, chartPr
 
 SeverityLine.propTypes = {
     value: PropTypes.number.isRequired,
-    title: PropTypes.string.isRequired,
+    title: PropTypes.node.isRequired,
     className: PropTypes.string,
-    tooltipMessage: PropTypes.string,
+    tooltipMessage: PropTypes.node,
     config: PropTypes.shape({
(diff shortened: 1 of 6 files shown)


packages/charts/src/SeverityLine/SeverityLine.js
```diff
@@ -45,9 +45,9 @@ const SeverityLine = ({ title, value, className, tooltipMessage, config, chartPr
 
 SeverityLine.propTypes = {
     value: PropTypes.number.isRequired,
-    title: PropTypes.string.isRequired,
+    title: PropTypes.oneOfType([ PropTypes.node, PropTypes.arrayOf(PropTypes.node) ]).isRequired,
```
rvsia, line 48: Isn't PropTypes.node the same as an array of nodes?

Hyperkid123: Yes, it is. Nasty habit of mine :+1:

rvsia (approved):

karelhala (approved):

rvsia:
Oh, you have to update PF version to satisfy DDF needs! 😅

Hyperkid123:
@rvsia the slider component is not released in LTS of PF. I've locked ddf to prev version. You can fix it if you want but if you do scope the PF version only to sources package pls.

PR #971, opened 2021-02-16
Update html-webpack-plugin to latest

rvsia:
When updating Sources UI to the latest config, the HTML plugin was returning error:

```
Error: The loader "/Users/rvsiansk/insights-project/sources-ui/node_modules/html-webpack-plugin/lib/loader.  js!/Users/rvsiansk/insights-project/sources-ui/src/index.html" didn't return html.
```

The latest version fixed that.

Changelog: [here](https://github.com/jantimon/html-webpack-plugin/blob/master/CHANGELOG.md#-breaking-changes)

### DIFF
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -27,7 +27,7 @@
         "file-loader": "^6.2.0",
         "git-revision-webpack-plugin": "^3.0.6",
         "html-replace-webpack-plugin": "^2.6.0",
-        "html-webpack-plugin": "^4.5.0",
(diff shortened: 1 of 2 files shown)


karelhala (commented):
Good point! Can you go trough the tests and figure why they are failing with new version?

rvsia:
@karelhala They changed the name of attribute from options to userOptions. I tracked the change and it seems it's just a random refactoring.

Hyperkid123 (approved):

PR #975, opened 2021-02-17
Enable option to pass sass prefix

karelhala:
### Sass prefix missing

Some apps have different class on main element and styles are not applied. This PR fixes such issue by allowing to pass custom sass prefix when using shared config.

### DIFF
--- a/packages/config/src/config.js
+++ b/packages/config/src/config.js
@@ -92,7 +92,7 @@ module.exports = ({
                                  * Context class is equal to app name and that class ass added to root element via the chrome-render-loader.
                                  */
                                 if (relativePath.match(/^src/)) {
-                                    return `.${appName}${sassPrefix ? `, .${sassPrefix}` : ''}{\n${content}\n}`;
+                                    return `${sassPrefix || `.${appName}`}{\n${content}\n}`;
                                 }
 
                                 return content;
--- a/packages/config/src/config.js
+++ b/packages/config/src/config.js
@@ -12,7 +12,8 @@ module.exports = ({
     mode,
     appName,
     useFileHash = true,
-    betaEnv = 'ci'
+    betaEnv = 'ci',
+    sassPrefix
 } = {}) => {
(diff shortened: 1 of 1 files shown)


rvsia (commented):
@karelhala  Shouldn't this be documented somewhere?

packages/config/src/config.js
```diff
@@ -91,7 +92,7 @@ module.exports = ({
                                  * Context class is equal to app name and that class ass added to root element via the chrome-render-loader.
                                  */
                                 if (relativePath.match(/^src/)) {
-                                    return `.${appName}{\n${content}\n}`;
+                                    return `.${appName}${sassPrefix ? `, .${sassPrefix}` : ''}{\n${content}\n}`;
```
Hyperkid123, line 95: I though the `sassPrefix` will replace the `appName` instead of having both.

karelhala: Mhh, yeah, that can be replaced completely.

packages/config/src/config.js
```diff
@@ -92,7 +92,7 @@ module.exports = ({
                                  * Context class is equal to app name and that class ass added to root element via the chrome-render-loader.
                                  */
                                 if (relativePath.match(/^src/)) {
-                                    return `.${appName}${sassPrefix ? `, .${sassPrefix}` : ''}{\n${content}\n}`;
+                                    return `${sassPrefix || `.${appName}`}{\n${content}\n}`;
```
Hyperkid123, line 95: I think you are missing the `.` before the prefix. Unless we expect the user to input prefix with the dot.

karelhala: Yup, we expect the user to do that. The prefix can be any sass prefix for instance `#something .oneClass, .rbac`

Hyperkid123: :+1:

Hyperkid123 (approved):

PR #977, opened 2021-02-17
Add aria-label to description filter text input

gkarat:
All filters inside `RuleFilters/` with ` value: 'description'` that are rendered within `RuleTable` component (see `filters` property) generate error in the console: `Text input: Text input requires either an id or aria-label to be specified`.

This PR adds `aria-label` property to all of them (right now there is only one such filter -- `description`).


gkarat:
@rvsia, @karelhala, could you review, please?

rvsia (approved):

PR #979, opened 2021-02-18
Remove eslint plugin

rvsia:
Blocking builds because of eslint is never a good idea 😬


Hyperkid123 (commented):
@rvsia can you add an option so plugins can be added to the config? Just spreading a new array of user input plugins should do the trick.

Hyperkid123 (approved):

PR #981, opened 2021-02-18
fix(config): use jsVarName for moduleName

redallen:
Fixes empty fed-mods.json files. Tested against access-requests-frontend.


Hyperkid123:
@redallen are the any required changes in the main.yml config or the in-app webpack config for existing apps?

redallen:
@Hyperkid123 Nope. This just makes it so it's not longer required to pass `module` to `config` or `moduleName` to `federated-modules.js`. If apps were passing them they'll continue to be used.

Hyperkid123 (approved):

PR #984, opened 2021-02-22
Remove singletings from certain shared dependencies.

Hyperkid123:
Singleton config ignores version restrictions and causes runtime issues when a different version of the module was shared before the application requests its own.

Know issues
- runtime error for redux promise middleware between version 5 -> 6
- undefined context for react-redux
- undefined location value for react-router dom

Possible future issue but not fixed. 
- missing components in the lower version of PF (the API is not changing so it's possible that a new file will be shared when not found in webpack namespace)


redallen (approved):
This can be simplified a lot in the future with [a reducer](https://github.com/project-koku/koku-ui/blob/master/webpack.config.js#L180):
```js
...singletonDeps.reduce((acc, dep) => {
    acc[dep] = { singleton: true, requiredVersion: dependencies[dep] };
    return acc;
}, {})
```

Looks fine in the meantime.

packages/config/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@redhat-cloud-services/frontend-components-config",
-    "version": "4.0.8",
+    "version": "4.0.9-beta.1",
```
karelhala, line 3: No need to push this, release bot will override it.

karelhala (approved):

packages/config/package.json
Hyperkid123: The bot will make 4.0.9 by itself. It did that before.

PR #986, opened 2021-02-22
If last elem in location is not inv ID use secont to last

karelhala:
I don't really like using `.pop()`, because it mutates the original array, but I am afraid we have no other way in here.


packages/inventory/src/components/detail/InventoryDetail.js
```diff
@@ -12,6 +12,11 @@ import { addNotification } from '@redhat-cloud-services/frontend-components-noti
 import ApplicationDetails from './ApplicationDetails';
 import './InventoryDetail.scss';
 
+const fallbackInventoryId = () => {
+    const pathName = location.pathname.split('/');
+    return pathName.pop() || pathName.pop(); // if last partial is empty use second to last
```
rvsia, line 17: What about 

```
pathName[pathName.length - 1] || pathName[pathName.length - 2]
```

karelhala: That's even more confusing :(

Hyperkid123: @karelhala why don't we just strip replace `/\/$/` with `''` from the pathname?

karelhala: I try to avoid replacing, as it can lead to confusion. But This can be simplified to oneliner.

Hyperkid123 (approved):

packages/inventory/src/components/detail/InventoryDetail.js
rvsia: You can extract it to seperate functions `getlast(pathname) || getSecondLast(pathname)`

rvsia: Or make a function `getLastDefined(array)` and make it more generic.

PR #989, opened 2021-02-23
Migrate inventory to v3

karelhala:


Hyperkid123:
@karelhala the folder structure is not complete. In order for automatic env selection to work, we need each folder to include index.js which will point to desirable components. This file will be referenced by local `package.json` to switch between cjs/ems on demand.

This applies to any component/function that is "importable". It needs its folder with the same name as the JS file and index.js file that exports it.

Hyperkid123:
@karelhala will you create a manual release?

karelhala:
@Hyperkid123 yup, I'll do that tomorrow

Hyperkid123 (approved):
After this PR is merged, we should start with cleaning the imports in the whole repo and removing fallback options.

PR #992, opened 2021-02-25
fix(supported popover): RHICOMPL-1461 update docs link

akofink:
Corresponds with https://github.com/RedHatInsights/compliance-frontend/pull/1018

Since we have to change the link in both places, we may as well export it here so it's not such a pain in the future. The UnsupportedSSGVersion components (one in each code base) are not the same, but that seems like n obvious refactor for the future.


vkrizan (approved):
LGTM

PR #993, opened 2021-02-25
fix(RemediationsButton): RHICOMPL-1456 - Ensure supported is passed in

bastilian:


packages/inventory-compliance/src/ComplianceRemediationButton/ComplianceRemediationButton.js
```diff
@@ -42,8 +42,8 @@ class ComplianceRemediationButton extends React.Component {
     dataProvider = () => {
         const { allSystems, selectedRules } = this.props;
         const result = { systems: [], issues: [] };
-        allSystems.forEach(async (system) => {
-            system.supported === true && result.systems.push(system.id);
+        allSystems.filter((system) => (system.supported)).forEach(async (system) => {
```
vkrizan, line 45: Observation: Previously there was a strict comparison to `true`.

vkrizan (approved):
LGTM. Thanks.

Issue changed to RHICOMPL-1456.

packages/inventory-compliance/src/ComplianceRemediationButton/ComplianceRemediationButton.js
bastilian: Yeah?

packages/inventory-compliance/src/ComplianceRemediationButton/ComplianceRemediationButton.js
```diff
@@ -58,9 +59,15 @@ class ComplianceRemediationButton extends React.Component {
         });
     }
 
+    notEmptyData = async () => {
```
bastilian, line 62: It shouldn't happen (anymore), but just to ensure we don't pass on anything that would raise an error

packages/inventory-compliance/src/ComplianceRemediationButton/ComplianceRemediationButton.js
```diff
@@ -58,9 +59,15 @@ class ComplianceRemediationButton extends React.Component {
...
+        const data = await this.dataProvider();
```
vkrizan, line 63: Observation: Data processing would be called twice, once to check non-empy values and second to pass the data to via dataProvider to the button. Considering the amount of data flowing through it, it might decrease the performance of the page.

Ideally it might be precomputed/cached or handled down the stream within `RemediationButton`.

bastilian: I am aware of this. But for now, till we have time to refactor, this is the way.

packages/inventory-compliance/src/ComplianceRemediationButton/ComplianceRemediationButton.js
```diff
@@ -97,6 +103,7 @@ ComplianceRemediationButton.propTypes = {
         profiles: propTypes.arrayOf(propTypes.shape({
             refId: propTypes.string,
             name: propTypes.string,
+            supported: propTypes.bool.isRequired,
```
akofink, line 106: Shouldn't this be up one more level under `allSystems`?

bastilian: Yep, you were right. Moved it up. It should now be correct.

packages/inventory-compliance/src/SystemRulesTable/SystemRulesTable.js
```diff
@@ -328,6 +328,12 @@ class SystemRulesTable extends React.Component {
             this.state.activeFilters,
             { hideLabel: true }
         );
+        const systemToRemediate = {
+            id: system.id,
+            profiles: system.testResultProfiles,
+            ruleObjectsFailed: [],
+            supported: system.supported
```
akofink, line 335: This seems to break when creating a new policy in the rules step. `system` is `undefined`.

bastilian: @akofink Good catch! I wasn't considering the remediations disabled catch. It shouldn't break anymore.

packages/inventory-compliance/src/ComplianceRemediationButton/ComplianceRemediationButton.js
```diff
@@ -7,8 +7,8 @@ import { addNotification } from '@redhat-cloud-services/frontend-components-noti
 import { AnsibeTowerIcon } from '@patternfly/react-icons';
 
 class ComplianceRemediationButton extends React.Component {
-    formatRule = ({ title, refId }, profile, system) => ({
-        id: `ssg:rhel7|${profile.split('xccdf_org.ssgproject.')[1]}|${refId}`,
+    formatRule = ({ title, refId }, profile, system, majorOsVersion) => ({
+        id: `ssg:rhel${majorOsVersion}|${profile.split('xccdf_org.ssgproject.')[1]}|${refId}`,
```
akofink, line 11: I think we need some validation here to ensure `majorOsVersion` is not `undefined`.

image

packages/inventory-compliance/src/SystemRulesTable/__snapshots__/SystemRulesTable.test.js.snap
```diff
@@ -1364,9 +1366,11 @@ exports[`SystemRulesTable component should render filtered and search mixed resu
                       "title": "Disable At Service (atd)",
                     },
                   ],
+                  "supported": true,
                 },
               ],
               "ruleObjectsFailed": Array [],
+              "supported": undefined,
```
akofink, line 1373: I think the systems fixture needs to include this?

packages/inventory-compliance/src/ComplianceRemediationButton/ComplianceRemediationButton.js
bastilian: I opened a PR to request the OS version on the report details as well. It shouldn't be undefined anymore with that for this case: https://github.com/RedHatInsights/compliance-frontend/pull/1019

We could set a default in the buttons `formatRules` function, but the user would still get a similar error.

packages/inventory-compliance/src/Fixtures/Fixtures.js
```diff
@@ -6,6 +6,7 @@ export const remediationsResponse = {
 export const system = {
     id: 'aa9c4497-5707-4233-9e9b-1fded5423ef3',
     name: '3.example.com',
+    supported: true,
     testResultProfiles: [{
```
akofink, line 10: I guess we'll need to add majorOsVersion here after https://github.com/RedHatInsights/compliance-frontend/pull/1019

bastilian: Yeah, we should. Right now there are no tests for the `ComplianceRemediationsButton` component or any that invoke it to cause a failure. 

We could add tests now, but I'd rather fix the issues with it for now and revisit it later to refactor it also with more support from the backend to not have to construct much data here in the frontend.

akofink (approved):
Works as expected. Thanks for taking care of these regressions, @bastilian! :tada:

PR #1001, opened 2021-03-02
Add review systems step

fhlavac:
Fixes: https://issues.redhat.com/browse/RHCLOUD-9978

- added review systems step 
- both issue-level and top-level systems are added to the selection
- updated tests

also fixed wrong wizard description: https://issues.redhat.com/browse/RHCLOUD-12700 

systems01

systems02

@katierik in the systems table I had to omit filtering and the "OS" column from mockups because of missing API support. Please take look if it is ok for now.


jiridostal:
When a list of hosts is defined on issue level like this
```
issues: [{
                id: 'vulnerabilities:CVE-2019-3815',
                description: 'Bonding will not fail over to the backup link when bonding options are partially read',
                systems: ['34b9f7d9-fc81-4e0f-bef0-c4b402a1510e']
            }, {
                id: 'advisor:network_bond_opts_config_issue|NETWORK_BONDING_OPTS_DOUBLE_QUOTES_ISSUE',
                description: 'CVE-2017-17712',
                systems: ['34b9f7d9-fc81-4e0f-bef0-c4b402a1510e', '9dfe0269-b676-4e15-8bb1-b7003ea00718']
            }]
```
There is an error thrown after clicking the remediation button
image


image

fhlavac:
@jiridostal sorry about that, it should be working now :-)

packages/remediations/src/RemediationWizard/RemediationWizard.js
```diff
@@ -115,6 +129,7 @@ const RemediationWizard = ({
                 FormTemplate={(props) => <Pf4FormTemplate {...props} showFormControls={false} />}
                 initialValues={{
                     [HAS_MULTIPLES]: !!state.resolutions?.find(r => r.resolutions.length > 1),
+                    [SYSTEMS]: undefined,
```
Hyperkid123, line 132: I've noticed that there is a lot of accessors with fallback value to `[]` or length to `0`. Is it possible to use an empty array as an initial value or are there some difficulties with it?

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -0,0 +1,166 @@
...
+
+    let dispatch = useDispatch();
+    const inventory = useRef(null);
+    const { input } = useFieldApi(props);
+    const formOptions = useFormApi();
+
+    const [ inventoryApi, setInventoryApi ] = useState();
+
+    const { selected, loaded, rows } = useSelector(({ entities }) => ({
+        selected: entities?.selected || new Map(),
```
Hyperkid123, line 30: Map should not be used with reducers as it is a mutable construction. Can we use plain object instead?

packages/remediations/src/steps/selectPlaybook.js
```diff
@@ -27,6 +27,10 @@ const SelectPlaybook = (props) => {
     const { input } = useFieldApi(props);
     const formOptions = useFormApi();
     const values = formOptions.getState().values;
+    const allSystems = [ ...new Set(issues.reduce((acc, curr) => [
```
Hyperkid123, line 30: You have multiple `[... new Set(...)]` calls. I think we can create a helper function called something like `dedupeArray`.

Hyperkid123 (commented):
Looks good. I have just a few minor comments about the code.

packages/notifications/src/notificationsMiddleware/notificationsMiddleware.test.js
```diff
@@ -3,6 +3,7 @@ import configureStore from 'redux-mock-store';
 import promiseMiddleware from 'redux-promise-middleware';
 import { notificationsMiddleware } from '..';
 import { ADD_NOTIFICATION } from '../redux/actions/action-types';
+import { createPromise } from 'redux-promise-middleware';
```
karelhala, line 6: No need to import like this, you are already importing from `promiseMiddleware` on line 3
```JS
import promiseMiddleware, { createPromise } from 'redux-promise-middleware';
```

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -0,0 +1,166 @@
...
+                    <Router>
+                        <InventoryTable
+                            hideFilters={{ all: true }}
+                            noDetail
+                            variant="compact"
+                            showTags
+                            onRefresh={onRefresh}
+                            ref={inventory}
+                            getEntities={async (_i, config) => {
+                                const invData = await inventoryApi?.getEntities(allSystems, {
```
karelhala, line 89: This will break if `allSystems` has more systems than 50, might want to change this to
```JS
const systems = (allSystems || []).slice((config.page -1) * config.per_page, config.page * config.per_page);
const data = await inventoryApi?.getEntities(allSystems, { ..config, hasItems: true}, true);
```

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -0,0 +1,166 @@
...
+                                    ...config,
+                                    hasItems: true
+                                });
+                                const data = await inventoryApi?.mapTags(invData);
```
karelhala, line 93: I've checked the inv api and looks like you don't have to fetch the tags if you include 3rd argument `true` to `getEntities`

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -0,0 +1,166 @@
...
+                {
+                    <Router>
+                        <InventoryTable
+                            hideFilters={{ all: true }}
+                            noDetail
+                            variant="compact"
+                            showTags
+                            onRefresh={onRefresh}
+                            ref={inventory}
+                            getEntities={async (_i, config) => {
```
karelhala, line 88: You should move this to a function defined above `return` and wrapped in `useCallback`
```JS
const fetchSystemsInfo = useCallback(async (_i, config) => {
  return await inventoryApi?.getEntities(
    (allSystems || []).slice((config.page -1) * config.per_page, config.page * config.per_page),
    { ..config, hasItems: true},
    true
  );
}, [allSystems]);
```

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -0,0 +1,166 @@
...
+                                items: [{
+                                    title: 'Select none (0)',
+                                    onClick: () => onSelectRows(false)
+                                },
+                                {
+                                    ...loaded && rows && rows.length > 0 ? {
+                                        title: `Select page (${ rows.length })`,
+                                        onClick: () => {
+                                            onSelectRows(true);
+                                        }
```
karelhala, line 113: Redundant function wrapper
```suggestion
                                        onClick: () => onSelectRows(true)
```

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -0,0 +1,166 @@
...
+                                    } : {}
+                                }],
+                                checked: selected.size > 0,
+                                onSelect: (value) => {
+                                    onSelectRows(value);
+                                }
```
karelhala, line 119: Redundant function wrapper
```suggestion
                                onSelect: (value) => onSelectRows(value)
```

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -0,0 +1,166 @@
...
+
+    return (
+        <Fragment>
+            {
+                registry && registry?.store  && <Provider store={registry?.store}>
+                    <ReviewSystems {...props} registry={registry} />
+                </Provider>
+            }
+        </Fragment>
+    );
```
karelhala, line 163: No need to wrap in fragment and extra check for `registry`
```suggestion
    return registry?.store  && <Provider store={registry.store}>
      <ReviewSystems {...props} registry={registry} />
    </Provider>;
```

jiridostal:
@fhlavac Thanks! Looks good to me, unfortunately, the second step randomly fails in API call to 
`/api/inventory/v1/hosts/` with Bad Request (no more details provided) and Wizard keeps loading. Usually, it happens when I select a larger number of systems for remediation >50.

Even when I'm trying the same set of systems/issues, it fails randomly. Don't have any idea what that could be...

packages/remediations/src/steps/reviewSystems.js
jiridostal: This might be the cause of the issue I posted below...

katierik:
Missing OS is ok since I don't think it is currently showing in the Inventory table.
Why is filtering unavailable? One of the UX standards for our tables is that they must have filtering and sorting.

package.json
```diff
@@ -33,7 +33,7 @@
         "react-router-dom": "^5.2.0",
         "react-truncate": "^2.4.0",
         "redux": "^3.7.2",
-        "redux-promise-middleware": "^5.1.1",
+        "redux-promise-middleware": "^6.1.2",
```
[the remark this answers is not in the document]
fhlavac: Moved update of the package to local package.json, thank you

karelhala:
@katierik the problem with filters is that some systems comes from selected table and some from remediations, they are then joined together. However there is no real API filtering available I don't think it's possible to have even system name filter on frontend as we are getting this information from inventory. @fhlavac can we somehow receive the system name from selected systems and remediations API?

packages/remediations/src/RemediationWizard/RemediationWizard.js
fhlavac: @Hyperkid123  Undefined value is used for the automatic selection of all items on the initial load of the step. It needs to be different from empty array to avoid this selection happening when loading data on paginating. I can't think of a better way to solve this yet.

packages/remediations/src/steps/reviewSystems.js
fhlavac: Changed to a basic array, Map is actually not needed

packages/remediations/src/steps/reviewSystems.js
fhlavac: oh, good point, thank you :-)

katierik:
@karelhala Ok -- The big issue with the systems table being unfilterable is that once this stretches past 10 systems or so, it becomes impossible for users to browse the systems to make sure they are or are not including anything in particular. 

_[Suggestion below is based on nothing and I am not a software engineer and I'm speaking based on not much knowledge]_
Could we somehow take what the remediations API is saying the systems are and filter the large systems list down to a subset and then show that subset only? Because the large systems list has filter/sort/etc and that is the same functionality expected for this table.

karelhala:
@katierik I just spoke with @fhlavac and looks like we should be able to show at least the Filter by name. It will require going trough all apps that use Remediations button and pass ID and name as well, but it will be much better than having no filter.

fhlavac:
@katierik fetching all the systems and filtering on the client side would be really slow. As @karelhala I'll enable at least filtering by name, but preferably as a follow-up PR to avoid blocking other teams waiting for this PR to be merged. At the same time, I'm verifying if there isn't a chance to add filtering support to the API.

packages/remediations/src/steps/reviewSystems.js
fhlavac: @jiridostal yes, this should solve it

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -0,0 +1,160 @@
...
+
+const ReviewSystems = (props) => {
+    const { issues, systems, registry } = props;
+
+    let dispatch = useDispatch();
+    const inventory = useRef(null);
+    const { input } = useFieldApi(props);
+    const formOptions = useFormApi();
+
+    const [ inventoryApi, setInventoryApi ] = useState();
```
karelhala, line 27: Looking at the usage, it should be more performant to use `useRef` here.

```JS
const inventoryApi = useRef({});
```

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -0,0 +1,160 @@
...
+                            hideFilters={{ all: true }}
+                            noDetail
+                            variant="compact"
+                            showTags
+                            onRefresh={onRefresh}
+                            ref={inventory}
+                            getEntities={(_i, config) => fetchSystemsInfo(_i, config)}
+                            onLoad={({ mergeWithEntities, api }) => {
+                                registry.register(mergeWithEntities(entitiesReducer(onSelect, input.value)));
+                                setInventoryApi(() => api);
```
karelhala, line 102: To assign to a inventoryApi ref call
```JS
inventoryApi.current = api;
```

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -0,0 +1,160 @@
...
+        const systems = (allSystems || []).slice((config.page - 1) * config.per_page, config.page * config.per_page);
+        const data = await inventoryApi?.getEntities(systems, { ...config, hasItems: true, page: 1 }, true);
+        return {
+            ...data,
+            total: allSystems.length,
+            page: config.page,
+            // eslint-disable-next-line camelcase
+            per_page: config.per_page
+        };
+    }, [ allSystems ]);
```
karelhala, line 78: Looking at the dependency list it might be better to move this function out of this component
```JS
const fetchSystemsInfo = (config, allSystem, { getEntities } = {}) => {
  // ... rest of the function
}
```

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -0,0 +1,160 @@
...
+                {
+                    <Router>
+                        <InventoryTable
+                            hideFilters={{ all: true }}
+                            noDetail
+                            variant="compact"
+                            showTags
+                            onRefresh={onRefresh}
+                            ref={inventory}
+                            getEntities={(_i, config) => fetchSystemsInfo(_i, config)}
```
karelhala, line 99: To use the function that you moved from the component simply call
```JSX
getEntities={(_i, config) => fetchSystemsInfo(config, allSystems, inventoryApi.current)}
```

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -0,0 +1,160 @@
...
+                            }
+                            bulkSelect={{
+                                id: 'select-systems',
+                                count: selected.length,
+                                items: [{
+                                    title: 'Select none (0)',
+                                    onClick: () => onSelectRows(false)
+                                },
+                                {
+                                    ...loaded && rows && rows.length > 0 ? {
```
karelhala, line 113: You don't have to check for `rows`
```suggestion
                                    ...loaded && rows.length > 0 ? {
```

fhlavac: @karelhala thank you, updated :-)

katierik:
@karelhala @fhlavac Thank you! I appreciate your work on this :)

karelhala (approved):

PR #1004, opened 2021-03-02
chore(InventoryDetail): display custom app children components

mkholjuraev:
We need to add more info to the Patch system details page header and I am able to add custom components to Inventory Header after this change in InventoryDetail.js.  Please feel free to ping me if my PR needs some changes :).

image

### DIFF
--- a/packages/inventory/src/components/detail/InventoryDetail.js
+++ b/packages/inventory/src/components/detail/InventoryDetail.js
@@ -25,7 +25,8 @@ const InventoryDetail = ({
     onBackToListClick,
     showDelete,
     appList,
-    showInventoryDrawer
+    showInventoryDrawer,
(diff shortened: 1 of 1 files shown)


packages/inventory/src/components/detail/InventoryDetail.js
```diff
@@ -83,7 +85,8 @@ InventoryDetail.propTypes = {
         pageId: PropTypes.string
     })),
     onTabSelect: PropTypes.func,
-    onBackToListClick: PropTypes.func
+    onBackToListClick: PropTypes.func,
+    children: PropTypes.element
```
rvsia, line 89: Shouldn't it be a `node`?

rvsia (approved):

PR #1010, opened 2021-03-03
Docs nav groups

Hyperkid123:
### Changes
- enable eslint in docs package
- update docs FEC dependnecy
- remove single item groups from docs nav

IF a nav group has only one item in it, the now renders only that one item, instead of a group with one item in it

screenshot-0 0 0 0_8080-2021 03 03-15_58_31

### DIFF
--- a/packages/docs/components/navigation/index.js
+++ b/packages/docs/components/navigation/index.js
@@ -11,8 +11,73 @@ const useStyles = createUseStyles({
         '& > button': {
             textTransform: 'capitalize'
         }
+    },
+    groupLink: {
+        'font-size': 14
     }
 });
+
+const NavigationGroup = ({ group, items, packageName }) => {
+    const classes = useStyles();
+    const { pathname } = useRouter();
+    if (items.length === 1) {
+        const  item = items[0];
+        const title = typeof item === 'object' ? item.title : item;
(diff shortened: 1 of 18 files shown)


packages/docs/components/navigation/index.js
```diff
@@ -11,9 +11,74 @@ const useStyles = createUseStyles({
...
+                                </a>
+                            </Link>
+                        )}
+                    >
+                        {title}
+                    </NavItem>
+                );})}
+        </NavGroup>
+    );
+
```
rvsia, line 79: empty line

packages/docs/pages/_app.js
```diff
@@ -2,9 +2,9 @@ import Layout from '../components/layout';
 import { MDXProvider } from '@mdx-js/react';
 import components from '../components/layout/mdx-provider-components';
 import '@patternfly/react-styles/css/components/Table/table.css';
+import '@redhat-cloud-services/frontend-components/index.css';
```
rvsia, line 5: Do we need to include the css file when we are using v3?

Hyperkid123: Yeah. I need do to more investigation. The CSS imports are weird with the SSR.

rvsia (approved):

PR #1011, opened 2021-03-04
chore(docs): add dockerfile to run docs

Hyperkid123:


packages/docs/docker.package.json
```diff
@@ -0,0 +1,45 @@
...
+    "next": "^10.0.7",
+    "prism-react-renderer": "^1.1.1",
+    "react": "^17.0.1",
+    "react-dom": "^17.0.1",
+    "react-json-view": "^1.19.1",
+    "react-redux": "^7.2.2",
+    "react-router-dom": "^5.2.0",
+    "redux": "^4.0.5"
+  },
+  "devDependencies": {
```
rvsia, line 29: I don't think there is a difference between devDeps and deps. 🤔  Maybe we could merge them together.

packages/docs/docker.package.json
```diff
@@ -0,0 +1,45 @@
...
+    "concurrently": "^5.3.0",
+    "file-loader": "^6.2.0",
+    "jsdoc": "^3.6.6",
+    "jsdoc3-parser": "^2.0.0",
+    "next-transpile-modules": "^6.0.0",
+    "raw-loader": "^4.0.2",
+    "react-docgen": "^5.3.1",
+    "react-jss": "^10.5.0",
+    "url-loader": "^4.1.1",
+    "webpack": "^4.0.0"
```
rvsia, line 43: Do you need to specify webpack? Isn't it a part of Next?

Hyperkid123: Yes, we do. There is an issue with webpack 5 and one of the CSS decorators. I don't know what is it exactly so we need webpack 4 to be installed.

packages/docs/docker.package.json
Hyperkid123: Nah keeping it separate is good for our sanity.

rvsia (approved):

ryelo (approved):

PR #1012, opened 2021-03-04
fix(docs): wrap mdlinks in router Link

Hyperkid123:
Links generated from MD files used normal HTML `a` tag which caused full page refreshes. It has to be wrapped in next `Link` component to trigger the client-side routing.

### DIFF
--- a/packages/docs/components/layout/mdx-provider-components.js
+++ b/packages/docs/components/layout/mdx-provider-components.js
@@ -1,6 +1,7 @@
 import { Title, Card, CardBody } from '@patternfly/react-core';
 import { createUseStyles } from 'react-jss';
 import classnames from 'classnames';
+import Link from 'next/link';
 
 const useTableStyles = createUseStyles({
     card: {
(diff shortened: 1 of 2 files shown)


rvsia (approved):

packages/docs/components/layout/mdx-provider-components.js
```diff
@@ -1,13 +1,15 @@
...
 import classnames from 'classnames';
+import Link from 'next/link';
 
 const useTableStyles = createUseStyles({
     card: {
         overflowY: 'auto'
     }
 });
 
+const A = ({ children, target,  ...props }) => <Link {...props}><a target={target}>{children}</a></Link>;
```
rvsia, line 12: Shouldn't you also provide a `href` to the `a` element? 🤔

Hyperkid123: No, the href is injected by the link. However, it seems that it consumes the target property, maybe others aswell

rvsia (approved):

PR #1022, opened 2021-03-09
fix(compliance): RHICOMPL-1488 Rules table header

vkrizan:
s/Rule/Name/

Continuation at https://github.com/RedHatInsights/compliance-frontend/pull/1035


akofink (commented):
Works as expected. We also need the same thing on the system details rules table. The UX reasoning given in the jira is that 'filter by name' should match the name column.

image

akofink (approved):
Oops, sorry about that. Thanks, @vkrizan!

PR #1023, opened 2021-03-10
Add settings for proxy

rvsia:
## useProxy

The config provide a way to run your applications without using insights-proxy. Just add `useProxy: true` to your configuration.

```jsx
const { config: webpackConfig, plugins } = config({
  rootFolder: resolve(__dirname, '../'),
  debug: true,
  useFileHash: false,
  deployment: process.env.BETA ? 'beta/apps' : 'apps',
  useProxy: true,
});
```

And run your application with following command:

```shell
NODE_ENV=development BETA=true webpack serve --config config/dev.webpack.config.js
```

*Path to config can be different*

Then you will find your application running on `https://ci.foo.redhat.com:1337/beta/...`. This works only with **Chrome 2** ready applications.

You can change the environment by setting `betaEnv`, but unless the targeted environment is using Chrome 2.0, it won't work correctly.

### localChrome

You can also easily run you application with a local build of Chrome by adding `localChrome: <absolute_path_to_chrome_build_folder>`.

```jsx
  rootFolder: resolve(__dirname, '../'),
  debug: true,
  useFileHash: false,
  deployment: process.env.BETA ? 'beta/apps' : 'apps',
  useProxy: true,
  localChrome: process.env.INSIGHTS_CHROME,
});
```

```shell
INSIGHTS_CHROME=/Users/rvsiansk/insights-project/insights-chrome/build/
```

To check what the proxy is doing with your local chrome settings you can set `proxyVerbose: true`.

### Custom proxy settings

You can add any additional custom proxy configuration. See [Webpack documentation](https://webpack.js.org/configuration/dev-server/#devserverproxy). Please, provide your configuration as an array of object containing `context` property.

*Example*:

```jsx
const { config: webpackConfig, plugins } = config({
  rootFolder: resolve(__dirname, '../'),
  debug: true,
  useFileHash: false,
  deployment: process.env.BETA ? 'beta/apps' : 'apps',
  useProxy: true,
  localChrome: process.env.INSIGHTS_CHROME,
  customProxy: [
    {
      context: (path) => path.includes('/api/'),
      target: 'https://qa.cloud.redhat.com',
      secure: false,
      changeOrigin: true,
      autoRewrite: true,
      ws: true,
    },
  ],
});
```

This configuration will redirect all API requests to QA environment, so you can check CI UI with QA data.

### DIFF
--- a/packages/config/README.md
+++ b/packages/config/README.md
@@ -14,3 +14,79 @@ You'll also have to update your `webpack-cli` dependency to `>=4.2.0`, this pack
 ### Removed features with webpack 5
 
 The new version of webpack 5 changed plyfills and plugin configs, some packages are outdated, one example is `lodash-webpack-plugin` this plugin is no longer maintain anyways. You should be just fine by installing `lodash` directly, imports can stay the same as before `import get from 'lodash/get`.
+
+
+## useProxy
+
+The config provide a way to run your applications without using insights-proxy. Just add `useProxy: true` to your configuration.
+
+```jsx
+const { config: webpackConfig, plugins } = config({
+  rootFolder: resolve(__dirname, '../'),
+  debug: true,
+  useFileHash: false,
+  deployment: process.env.BETA ? 'beta/apps' : 'apps',
+  useProxy: true,
+});
+```
+
+And run your application with following command:
+
+```shell
+NODE_ENV=development BETA=true webpack serve --config config/dev.webpack.config.js
+```
+
+*Path to config can be different*
+
+Then you will find your application running on `https://ci.foo.redhat.com:1337/beta/...`. This works only with **Chrome 2** ready applications.
+
+You can change the environment by setting `betaEnv`, but unless the targeted environment is using Chrome 2.0, it won't work correctly.
+
+### localChrome
+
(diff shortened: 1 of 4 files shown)


karelhala (commented):
Awesome effort! Thank you for rasing this up. What about option to consume additional paths? For instance custom cloud-services-config would need custom proxy, which is kinda dumb. How about consuming a config similiar to insights proxy?

rvsia:
@karelhala Providing custom routes is possible via `customProxy` attribute. We could make some function, that will take the spandx config and convert it to the configuration that webpack accepts, it should be pretty simple - however serving local files is not so straightforward. However, I would like to focus on the basic use-case now, which should be sufficient for frontend development.

karelhala:
@rvsia yeah I am not refering to serving local files, I'd still imagine users would spin new http server in the config folder and just have option to pass new config to the webpack config.

Hyperkid123:
@karelhala I think we should merge this since it's an opt-in feature and add additional configuration is second run. It gives an opportunity to test out the config in everyday use.

Hyperkid123 (approved):

PR #1027, opened 2021-03-15
Add option to pass routes config or use direct routes

karelhala:
### Option to re-use old proxy config

Since we provide a new way of using proxy we should also allow passing old proxy config with routes. This PR adds 2 options
* pass in list of routes - simply use config and pass `routes` to it
```JS
  rootFolder: resolve(__dirname, '../'),
  debug: true,
  useFileHash: false,
  deployment: process.env.BETA ? 'beta/apps' : 'apps',
  useProxy: true,
  routes: {
      '/config/main.yml': { host: 'http://127.0.0.1:8889' }
  },
```
* pass in path to config - if you have old config stored on disk you can now pass `routesPath` to shared config
```JS
  rootFolder: resolve(__dirname, '../'),
  debug: true,
  useFileHash: false,
  deployment: process.env.BETA ? 'beta/apps' : 'apps',
  useProxy: true,
  routesPath: process.env.CONFIG_PATH
```

And when running start you can do `CONFIG_PATH=/home/khala/Documents/git/RedHatInsights/spandx.config.js npm start` the spandx.config.js is just regular config we used before.

### DIFF
--- a/packages/config/src/proxy.js
+++ b/packages/config/src/proxy.js
@@ -75,6 +75,7 @@ function createInsightsProxy({ betaEnv, rootFolder, localChrome, customProxy = [
             }] : []),
             ...proxyRoutes ? Object.entries(proxyRoutes).map(([ route, redirect ]) => {
                 const currTarget = redirect.host || redirect;
+                delete redirect.host;
                 return {
                     context: (path) => path.includes(route),
                     target: currTarget === 'PORTAL_BACKEND_MARKER' ? target : currTarget,
@@ -82,7 +83,7 @@ function createInsightsProxy({ betaEnv, rootFolder, localChrome, customProxy = [
                     changeOrigin: true,
                     autoRewrite: true,
                     ws: true,
-                    ...redirect
+                    ...typeof redirect === 'object' ? redirect : {}
                 };
             }) : [],
             ...customProxy
--- a/packages/config/src/proxy.js
+++ b/packages/config/src/proxy.js
@@ -4,15 +4,25 @@ const { sync } = require('glob');
 function createInsightsProxy({ betaEnv, rootFolder, localChrome, customProxy = [], publicPath, proxyVerbose, https = true, port, routes, routesPath }) {
     const target = betaEnv === 'prod' ? 'https://cloud.redhat.com/' : `https://${betaEnv}.cloud.redhat.com/`;
 
-    let proxyRoutes = routes || (routesPath && require(routesPath));
-    proxyRoutes = proxyRoutes && (proxyRoutes.routes || proxyRoutes);
+    let proxyRoutes = routes;
+
+    if (routesPath) {
+        proxyRoutes = require(routesPath);
+    }
+
+    if (proxyRoutes) {
+        proxyRoutes = proxyRoutes.routes || proxyRoutes;
+    }
 
     // eslint-disable-next-line no-console
     proxyVerbose && proxyRoutes && console.log(`Using proxy routes: ${JSON.stringify(proxyRoutes, null, 2)}`);
 
     const isNotCustomContext = (path) => {
-        const isRoutes = Object.keys(proxyRoutes || {}).some((route) => path.includes(route));
-        return customProxy.length > 0 ? customProxy.reduce((acc, curr) => acc && !curr.context(path), !isRoutes) : !isRoutes;
+        if (Object.keys(proxyRoutes || {}).some((route) => path.includes(route))) {
+            return false;
+        }
+
+        return customProxy.length > 0 ? customProxy.reduce((acc, curr) => acc && !curr.context(path), true) : true;
     };
 
     return {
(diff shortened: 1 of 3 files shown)


packages/config/src/proxy.js
```diff
@@ -54,6 +63,14 @@ function createInsightsProxy({ betaEnv, rootFolder, localChrome, customProxy = [
...
                 }
             }] : []),
+            ...proxyRoutes ? Object.entries(proxyRoutes).map(([ route, redirect ]) => ({
+                context: (path) => path.includes(route),
+                target: redirect.host || redirect === 'PORTAL_BACKEND_MARKER' ? target : redirect.host || redirect,
+                secure: false,
+                changeOrigin: true,
+                autoRewrite: true,
+                ws: true
+            })) : [],
```
rvsia, line 73: If the route is an object, I would spread it here.

packages/config/src/proxy.js
```diff
@@ -54,6 +63,14 @@ function createInsightsProxy({ betaEnv, rootFolder, localChrome, customProxy = [
                     }
                 }
             }] : []),
+            ...proxyRoutes ? Object.entries(proxyRoutes).map(([ route, redirect ]) => ({
+                context: (path) => path.includes(route),
+                target: redirect.host || redirect === 'PORTAL_BACKEND_MARKER' ? target : redirect.host || redirect,
```
rvsia, line 68: The last OR does not make sense - you already checking `redirect.host` at the beggining.

packages/config/src/proxy.js
```diff
@@ -1,10 +1,19 @@
...
-    const isNotCustomContext = (path) => customProxy.length > 0 ? customProxy.reduce((acc, curr) => acc && !curr.context(path), true) : true;
+    let proxyRoutes = routes || (routesPath && require(routesPath));
+    proxyRoutes = proxyRoutes && (proxyRoutes.routes || proxyRoutes);
+
+    // eslint-disable-next-line no-console
+    proxyVerbose && proxyRoutes && console.log(`Using proxy routes: ${JSON.stringify(proxyRoutes, null, 2)}`);
+
+    const isNotCustomContext = (path) => {
+        const isRoutes = Object.keys(proxyRoutes || {}).some((route) => path.includes(route));
+        return customProxy.length > 0 ? customProxy.reduce((acc, curr) => acc && !curr.context(path), !isRoutes) : !isRoutes;
```
rvsia, line 15: If `isRoute === true` we can skip the whole reduce section.

packages/config/src/proxy.js
```diff
@@ -1,10 +1,19 @@
 const { readFileSync } = require('fs');
 const { sync } = require('glob');
 
-function createInsightsProxy({ betaEnv, rootFolder, localChrome, customProxy = [], publicPath, proxyVerbose, https = true, port }) {
+function createInsightsProxy({ betaEnv, rootFolder, localChrome, customProxy = [], publicPath, proxyVerbose, https = true, port, routes, routesPath }) {
     const target = betaEnv === 'prod' ? 'https://cloud.redhat.com/' : `https://${betaEnv}.cloud.redhat.com/`;
 
-    const isNotCustomContext = (path) => customProxy.length > 0 ? customProxy.reduce((acc, curr) => acc && !curr.context(path), true) : true;
+    let proxyRoutes = routes || (routesPath && require(routesPath));
```
rvsia, line 7: What about to change this and the line below to old good `if`, seems too complex to me.

packages/config/src/proxy.js
karelhala: Good point!

packages/config/src/proxy.js
karelhala: I forgot to wrap it in parenthesis, it should be `(redirect.host || redirect === 'PORTAL_BACKEND_MARKER') ? target : (redirect.host || redirect)`

rvsia: Then it's the same situation - if redirect.host is true, then it's target.

packages/config/src/proxy.js
karelhala: Seems a bit too wordy, but I am fine with that
```JS
    let proxyRoutes;
    if (routes) {
        proxyRoutes = routes;
    } else if (routesPath) {
        proxyRoutes = require(routesPath);
    }

    if (proxyRoutes) {
        proxyRoutes = proxyRoutes.routes || proxyRoutes;
    }
```

rvsia: It doesn't have to be.

```
let proxyRoutes = routes;

if(routesPath) {
   const fileRoutes = require(routesPath);
   proxyRoutes = fileRoutes.routes || fileRoutes;
}
```

packages/config/src/proxy.js
karelhala: Wrong brackets again! `(redirect.host || redirect) === 'PORTAL_BACKEND_MARKER' ? target : (redirect.host || redirect)` might be better to assign `redirect.host || redirect` to a variable to avoid confusion.

packages/config/src/proxy.js
karelhala: I'd still keep the second if though.
```JS
let proxyRoutes = routes;

if(routesPath) {
   proxyRoutes = require(routesPath);
}

if (proxyRoutes) {
  proxyRoutes = proxyRoutes.routes || proxyRoutes;
}
```

rvsia: I think it would be better to fail instead of pretending that nothing wrong happened if users want to use a file that does not exist.

This is just a specific developer tool, it doesn't have to be fail-proof.

packages/config/src/proxy.js
```diff
@@ -54,6 +73,18 @@ function createInsightsProxy({ betaEnv, rootFolder, localChrome, customProxy = [
...
+            ...proxyRoutes ? Object.entries(proxyRoutes).map(([ route, redirect ]) => {
+                const currTarget = redirect.host || redirect;
+                return {
+                    context: (path) => path.includes(route),
+                    target: currTarget === 'PORTAL_BACKEND_MARKER' ? target : currTarget,
+                    secure: false,
+                    changeOrigin: true,
+                    autoRewrite: true,
+                    ws: true,
+                    ...redirect
```
rvsia, line 85: add a check if it's a object (otherwise the string will be spread) and we should also omit the host property from it.

rvsia (approved):

PR #1029, opened 2021-03-16
Allow hiding cards and fields in general information

karelhala:
### Hide cards on general info page

Since we have a way to fetch just a few fields from API we should also allow hiding certain parts of general info detail. This PR allows to do that by simply adding `Wrapper` prop for each card and if consumer passes `false` in it the card won't be shown. There is also option to pass in custom Wrapper for each card to hide certain fields or to render something completely different.

```JSX
<GeneralInformation
  store={useStore()}
  writePermissions={writePermissions}
  ConfigurationCardWrapper={false}
  SystemCardWrapper={(props) => (
    <Suspense>
      <SystemCard {...props} hasSAP={false} />
    </Suspense>
  )}
  OperatingSystemCardWrapper={(props) => (
    <Suspense fallback="">
      <OperatingSystemCard {...props} hasKernelModules={false} />
    </Suspense>
  )}
/>
```

### DIFF
--- a/packages/inventory-general-info/src/GeneralInformation/GeneralInformation.js
+++ b/packages/inventory-general-info/src/GeneralInformation/GeneralInformation.js
@@ -58,30 +58,41 @@ class GeneralInformation extends Component {
 
     render() {
         const { isModalOpen, modalTitle, cells, rows, expandable, filters, modalVariant } = this.state;
-        const { store, writePermissions } = this.props;
+        const {
+            store,
+            writePermissions,
+            SystemCardWrapper,
+            OperatingSystemCardWrapper,
+            BiosCardWrapper,
+            InfrastructureCardWrapper,
+            ConfigurationCardWrapper,
+            CollectionCardWrapper,
+            children
+        } = this.props;
         const Wrapper = store ? Provider : Fragment;
         return (
             <Wrapper {...(store && { store })}>
                 <div className="ins-c-general-information">
                     <Grid sm={ 12 } md={ 6 } hasGutter>
-                        <GridItem>
-                            <SystemCard handleClick={ this.handleModalToggle } writePermissions={writePermissions} />
-                        </GridItem>
-                        <GridItem>
-                            <OperatingSystemCard handleClick={ this.handleModalToggle } />
-                        </GridItem>
-                        <GridItem>
-                            <BiosCard handleClick={ this.handleModalToggle } />
-                        </GridItem>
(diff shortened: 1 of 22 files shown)


packages/inventory-general-info/src/GeneralInformation/GeneralInformation.js
```diff
@@ -111,10 +122,22 @@ GeneralInformation.propTypes = {
...
     loadSystemDetail: PropTypes.func,
     store: PropTypes.any,
-    writePermissions: PropTypes.bool
+    writePermissions: PropTypes.bool,
+    SystemCardWrapper: PropTypes.oneOfType([ PropTypes.node, PropTypes.bool ]),
+    OperatingSystemCardWrapper: PropTypes.oneOfType([ PropTypes.node, PropTypes.bool ]),
+    BiosCardWrapper: PropTypes.oneOfType([ PropTypes.node, PropTypes.bool ]),
+    InfrastructureCardWrapper: PropTypes.oneOfType([ PropTypes.node, PropTypes.bool ]),
+    ConfigurationCardWrapper: PropTypes.oneOfType([ PropTypes.node, PropTypes.bool ]),
+    CollectionCardWrapper: PropTypes.oneOfType([ PropTypes.node, PropTypes.bool ])
```
rvsia, line 131: Why `bool`? 🤔

karelhala: So consumers can rutn off the wrapper.
```JSX
<GeneralInformation
  store={useStore()}
  OperatingSystemCardWrapper={false}
  BiosCardWrapper={false}
  InfrastructureCardWrapper={false}
  ConfigurationCardWrapper={false}
  CollectionCardWrapper={false}
/>
```

Will show just bios card

rvsia: hmmm I would rather use `null` (= `Node`)

karelhala: Yeah, that works as well.

rvsia (approved):

packages/inventory-general-info/src/GeneralInformation/GeneralInformation.js
rvsia: Yeah, I understand, I just don't like the false/bool here.. 😄  but it doesn't matter too much.

PR #1033, opened 2021-03-17
refactor(external): RHICOMPL-1478 no longer handle nil policy

akofink:
External policies (i.e. profiles with a nil policy) are no longer
present in the backend and thus no longer need to be handled by the
frontend.

Counterpart to https://github.com/RedHatInsights/compliance-frontend/pull/1056, though the changes are not dependent on each other.


packages/inventory-compliance/src/SystemRulesTable/SystemRulesTable.js
```diff
@@ -187,7 +187,7 @@ class SystemRulesTable extends React.Component {
     updatePolicyFilterConfig = () => {
         let policies = this.props.profileRules.filter(({ profile }) => !!profile).map(({ profile }) => (
             {
-                id: profile.policy ? profile.policy.id : profile.id,
+                id: profile.policy.id,
```
bastilian, line 190: This seems to raise a `SystemRulesTable.js?afee:276 Uncaught (in promise) TypeError: Cannot read property 'id' of undefined` Error.  Should it maybe just be `profile.id`?

bastilian: The issue appears on the Rules tab on a Policy Details page.

bastilian:
@akofink Thank you! Only the one issue on the rules tab, but other than that all appears to still be working as expected.

packages/inventory-compliance/src/SystemRulesTable/SystemRulesTable.js
akofink: Nice find! Ideally, we'll ensure `profile.policy` exists here. Clearly some query isn't passing the policy info with the profile. I'll look into it. Thanks, @bastilian!

akofink: Ok! So, this table was used at one point to filter the system rules table by policy. We disabled it for now, but the code still remains here. At any rate, we also use this table in places that we don't need policy filtering, such as the policy rules table, where it doesn't make sense to [request the nested policy](https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/PolicyDetails/PolicyDetails.js#L42). That means this shouldn't break if `profile.policy` is `unassigned`, but it should use `profile.policy.id` for the filter if it's provided. I've updated it to `profile?.policy?.id`.

akofink: If the user doesn't define `profile.policy` properly, this filter won't work correctly, but this code runs even if the filter is disabled, so it needs to be able to handle this case when it's not defined. Using `profile.id` is definitely not correct (that was the case for external policies). I'm definitely open to other ideas here.

As I mentioned, this filter is currently not even used anywhere, though we may use it again in the future.

vkrizan: Isn't it used on System Details page to filter rules by a policy? Note, that the filter shows only if there is more than one policy reporting.

akofink: Ah yes, that's correct. It's used on system details page.

akofink: I've tested this change now with the policy filter on system details (with > 1 policy), and it seems to work as expected. Are there any concerns or regressions here that you are seeing?

vkrizan: Awesome. No concerns.

vkrizan (approved):
Sorry I thought that it already had an ack. Here it is...

Thanks.

PR #1034, opened 2021-03-19
Allow using onClick to show modals for extra items

karelhala:
### Onclick for extra items

Since we allow passing extra items to general information we should also reuse our modals if consumer wants to show tabled data. This PR introduces a check for onClick and if such callback is present it calls it and consumes array off of it.

### DIFF
--- a/packages/inventory-general-info/src/ConfigurationCard/ConfigurationCard.test.js
+++ b/packages/inventory-general-info/src/ConfigurationCard/ConfigurationCard.test.js
@@ -113,7 +113,7 @@ describe('ConfigurationCard', () => {
         const store = mockStore(initialState);
         const wrapper = render(<ConfigurationCard store={ store } extra={[
             { title: 'something', value: 'test' },
-            { title: 'with click', value: '1 tests', onClick: () => [ 'Something', {}, 'small' ] }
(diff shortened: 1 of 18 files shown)


packages/inventory-general-info/src/ConfigurationCard/ConfigurationCard.test.js
```diff
@@ -112,7 +112,8 @@ describe('ConfigurationCard', () => {
     it('should render extra', () => {
         const store = mockStore(initialState);
         const wrapper = render(<ConfigurationCard store={ store } extra={[
-            { title: 'something', value: 'test' }
+            { title: 'something', value: 'test' },
+            { title: 'with click', value: '1 tests', onClick: () => [ 'Something', {}, 'small' ] }
```
rvsia, line 116: I think it's very confusing to call it `onClick` when it's a function that returns an array - onClick should be just a function where consumers can do anything what they want to do.

I have two suggestions:

1. rename it to `items` (or whatever) and change it to just an `array`

2. provide handleClick to be consumed `onClick: (e, handleClick) => handleClick([ 'Something', {}, 'small' ])`

karelhala: I like the second approach!

rvsia (approved):

PR #1039, opened 2021-03-23
feat(RulesTable): RHICOMPL-1549 remediationAvailableFilter

akofink:
Add a filter to the rules table for rules with remediations available.

Test with https://github.com/RedHatInsights/compliance-frontend/pull/1065

### DIFF
--- a/packages/inventory-compliance/src/SystemRulesTable/SystemRulesTable.js
+++ b/packages/inventory-compliance/src/SystemRulesTable/SystemRulesTable.js
@@ -29,8 +29,11 @@ export const selectColumns = (columnTitles) => (
 );
 
 class SystemRulesTable extends React.Component {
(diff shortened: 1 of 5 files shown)


vkrizan (approved):
LGTM and works as expected. Thank you.

packages/inventory-compliance/src/SystemRulesTable/SystemRulesTable.js
```diff
@@ -29,8 +29,11 @@ export const selectColumns = (columnTitles) => (
 );
 
 class SystemRulesTable extends React.Component {
+    remediationAvailableFilter = this.props.remediationAvailableFilter || this.props.remediationsEnabled;
```
akofink, line 32: @vkrizan Updated. I'm not sure this is the 'right' way to do this, but it seems to work OK.

vkrizan: :+1:

PR #1042, opened 2021-03-25
Fix review steps by moving selector to bottom

karelhala:
### Error while loading ReviewSystems step

When rendering review system step there is error about accessing store too soon this is caused by how store is created and when reducer is registered. This PR fixes such issue by moving the selector right before the render which rearranges the hooks and the selector is correctly used.


fhlavac (approved):
Thank you @karelhala

PR #1043, opened 2021-03-25
Re-add async inventory so apps can update to newest version without the need for chrome 2

karelhala:


Hyperkid123 (approved):
I hoped at this point chrome 2 will be everywhere :crying_cat_face:

packages/remediations/src/tests/steps/reviewSystems.test.js
```diff
@@ -31,7 +37,7 @@ const createSchema = () => ({
     }]
 });
 
-describe('ReviewSystems', () => {
+describe.only('ReviewSystems', () => {
```
Hyperkid123, line 40: Only left over

karelhala:
release

PR #1046, opened 2021-03-26
Remove side effects from review systems steps

karelhala:
### No side effects in reducers

There is an error in remediation wizard with reducers having side effects by calling `onSelect`. This PR fixes it by using ref to selected systems and when this ref differs from actual selected systems send them to form input.

### DIFF
--- a/packages/remediations/src/steps/reviewSystems.js
+++ b/packages/remediations/src/steps/reviewSystems.js
@@ -29,14 +29,11 @@ const ReviewSystems = ({ issues, systems, registry, ...props }) => {
     const inventory = useRef(null);
     const { input } = useFieldApi(props);
     const formOptions = useFormApi();
-    const selectedRef = useRef();
     const inventoryApi = useRef({});
 
-    const { selected, loaded, rows } = useSelector(({ entities }) => ({
-        selected: entities?.selected || [],
-        loaded: entities?.loaded,
-        rows: entities?.rows || []
-    }));
+    const rowsLength = useSelector(({ entities }) => (entities?.rows || []).length);
+    const selected = useSelector(({ entities }) => entities?.selected || []);
+    const loaded = useSelector(({ entities }) => entities?.loaded);
 
     const formValues = formOptions.getState().values;
     const error = formOptions.getState().errors?.systems;
@@ -67,10 +64,11 @@ const ReviewSystems = ({ issues, systems, registry, ...props }) => {
         });
     };
 
-    if (!isEqual(selectedRef.current, selected)) {
-        selectedRef.current = selected;
-        input.onChange(selected);
-    }
+    useEffect(() => {
+        if (!isEqual(input.value, selected)) {
+            input.onChange(selected);
+        }
+    });
 
     return (
         <Stack hasGutter>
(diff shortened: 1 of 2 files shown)


packages/remediations/src/steps/reviewSystems.js
```diff
@@ -21,15 +21,23 @@ import { Provider, useDispatch, useSelector } from 'react-redux';
...
     const formOptions = useFormApi();
-
+    const selectedRef = useRef();
     const inventoryApi = useRef({});
 
+    const { selected, loaded, rows } = useSelector(({ entities }) => ({
+        selected: entities?.selected || [],
+        loaded: entities?.loaded,
+        rows: entities?.rows || []
+    }));
```
rvsia, line 39: shallow equal

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -52,22 +60,17 @@ const ReviewSystems = ({ issues, systems, registry, ...props }) => {
...
     };
 
-    const { selected, loaded, rows } = useSelector(({ entities }) => ({
-        selected: entities?.selected || [],
-        loaded: entities?.loaded,
-        rows: entities?.rows || []
-    }));
+    if (!isEqual(selectedRef.current, selected)) {
+        selectedRef.current = selected;
+        input.onChange(selected);
```
rvsia, line 72: This should be definitely in useEffect

Hyperkid123 (approved):

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -52,22 +60,17 @@ const ReviewSystems = ({ issues, systems, registry, ...props }) => {
...
             payload: value
         });
     };
 
-    const { selected, loaded, rows } = useSelector(({ entities }) => ({
-        selected: entities?.selected || [],
-        loaded: entities?.loaded,
-        rows: entities?.rows || []
-    }));
+    if (!isEqual(selectedRef.current, selected)) {
```
rvsia, line 70: maybe you don't have to have ref, just use the comparison array in useEffect hook. 🤔

rvsia (commented):
I would rather to have it refactored - so the select function is called separately where it should be called, because this could get very tricky (especially without proper tests). But if that is a huge task for now, I left some suggestions.

packages/remediations/src/steps/reviewSystems.js
karelhala: Not really, the `selected` is array we are checking is matching to previously selected systems. This actually has to be calculated on each render and if previously selected items are different from current one, the `input.onChange` has to tick.

packages/remediations/src/steps/reviewSystems.js
karelhala: This one will be tricky, `rows` are complicated value so that wouldn't really help. `selected` on other hand is array so `shallowEqual` would not help. But splitting this to specific selectors and using some comparators would not hurt.

packages/remediations/src/steps/reviewSystems.js
rvsia: So just move the computation into the effect - useEffect is called on each render. You should not do sideeffects in render function.

karelhala: Good point, I'll use empty dependency for the effect and check in the body. Or we could use `useMemo` on the selected systems, but I'm not sure if that would help.

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -52,22 +57,18 @@ const ReviewSystems = ({ issues, systems, registry, ...props }) => {
...
-    const { selected, loaded, rows } = useSelector(({ entities }) => ({
-        selected: entities?.selected || [],
-        loaded: entities?.loaded,
-        rows: entities?.rows || []
-    }));
+    useEffect(() => {
+        if (!isEqual(input.value, selected)) {
+            input.onChange(selected);
+        }
+    });
```
rvsia, line 71: oh this is perfect! thank you!

rvsia (approved):

packages/remediations/src/steps/reviewSystems.js
karelhala: Just for the record, using no dependency in the effect causes 1 render per select
```JS
useEffect(() => {
        if (!isEqual(input.value, selected)) {
            input.onChange(selected);
        }
    });
```
However using `isEqual(input.value, selected)` in the dependency list causes 2 renders per select action.
```JS
useEffect(() => {
  input.onChange(selected);
}, [ isEqual(input.value, selected) ]);
```

PR #1048, opened 2021-03-29
feat(RulesTable): RHICOMPL-1550 selected filter switch

akofink:
Add a Switch for selected rules vs. filter + filter chip

According to UX, this fixes an inconsistency with the selected filter:
>"selected" isn't really an aspect of the object (its a state) so the filter is not 100% appropriate.

This work is not behind any feature flag.

image

### DIFF
--- a/packages/inventory-compliance/src/SystemRulesTable/SystemRulesTable.js
+++ b/packages/inventory-compliance/src/SystemRulesTable/SystemRulesTable.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import propTypes from 'prop-types';
-import { Pagination, PaginationVariant, ToolbarItem } from '@patternfly/react-core';
+import { Pagination, PaginationVariant, Switch, ToolbarItem } from '@patternfly/react-core';
 import { CheckCircleIcon, ExclamationCircleIcon, AnsibeTowerIcon } from '@patternfly/react-icons';
 import { Table, TableHeader, TableBody, sortable, fitContent } from '@patternfly/react-table';
 import { PrimaryToolbar } from '@redhat-cloud-services/frontend-components/PrimaryToolbar';
@@ -32,7 +32,6 @@ class SystemRulesTable extends React.Component {
     remediationAvailableFilter = this.props.remediationAvailableFilter || this.props.remediationsEnabled;
 
     config = buildFilterConfig({
-        selectedFilter: this.props.selectedFilter,
         remediationAvailableFilter: this.remediationAvailableFilter,
         showPassFailFilter: (this.props.columns.filter((c) => (c.title === 'Passed')).length > 0)
     });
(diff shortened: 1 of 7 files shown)


packages/inventory-compliance/src/SystemRulesTable/SystemRulesTable.js
```diff
@@ -158,7 +157,7 @@ class SystemRulesTable extends React.Component {
         )
     )
 
-    selectedRules = (rules) => {
+    addIsSelected = (rules) => {
```
akofink, line 160: This function was a bit misleadingly named, and it conflicted with one I wanted to add.

packages/inventory-compliance/src/SystemRulesTable/SystemRulesTable.js
```diff
@@ -182,8 +185,10 @@ class SystemRulesTable extends React.Component {
     getRules = () => {
         const rules = toRulesArray(this.props.profileRules);
         return this.sortedRules(
-            this.filteredRules(
-                this.selectedRules(this.openedRules(rules))
+            this.selectedRules(
+                this.filteredRules(
```
vkrizan, line 189: What's the aim of the inversion of method calls?
Shouldn't `selectedRules` be called before filtering?

IIRC `selectedRules` sets the selected boolean.

The filtering does not work for me as expected. It always yields 0 results even if there are some rules selected. Could this be the reason?

vkrizan (changes requested):
Does not work for me. :confused:

packages/inventory-compliance/src/SystemRulesTable/SystemRulesTable.js
akofink: `selectedRules` was renamed to `addIsSelected`. As you've said, `addIsSelected` adds the field `isSelected` to the rules (aka sets the selected boolean). This new function, `selectedRules` returns only those rules which are selected, if `showOnlySelected` is `true`. The order of these functions doesn't really matter, since `filteredRules` doesn't use `isSelected`.

`selectedRules`, however, is the only function now that requires the `isSelected` boolean to be set, so it calls `addIsSelected` as a dependency.

You may have linked frontend-components incorrectly, perhaps? It works fine for me, and the snapshots seem to be updating as expected in the tests as well. Would you mind trying again please?

packages/inventory-compliance/src/SystemRulesTable/SystemRulesTable.js
```diff
@@ -172,6 +171,10 @@ class SystemRulesTable extends React.Component {
         });
     }
 
+    selectedRules = (rules) => this.addIsSelected(
+        this.state.selectedOnly ? rules.filter(rule => rule.isSelected) : rules
```
vkrizan, line 175: Is this correct to call `addIsSelected` prior to filtering?

akofink: It shouldn't affect any behavior. Filtering used to use `isSelected` but no longer needs it after this change.

akofink: btw, `addIsSelected` is called _after_ `filteredRules`, not before.

vkrizan: Lets rewrite it to this:
```
selectedRules = (rules) => {
    const filtered = this.state.selectedOnly ? rules.filter(rule => rule.isSelected) : rules;
    const selectAdded = this.addIsSelected(filtered);
    return selectAdded;
}
```

The `filtered` filters the rules by `isSelected` boolean which is added after the `filter()` is called.

What am I missing?

akofink: Hmm... yes, I see your point :laughing: this does not seem correct at all.

PR #1050, opened 2021-03-30
Add display_name filter to reviewSystems

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-13177
Added filtering by `display_name` to the new Remediation wizard "Review systems" step.

Systems filtering

@katierik


katierik:
AWESOME

packages/remediations/src/RemediationWizard/RemediationWizard.js
```diff
@@ -67,7 +77,16 @@ const RemediationWizard = ({
         }
     };
 
+    const fetchHostNames = (systems = []) => {
+        const perChunk = 50;
+        const chunks = splitArray(systems, perChunk);
+        chunks.map(chunk => {
```
Hyperkid123, line 83: No need for `map` you don't consume the return value. `forEach` is good enough.

packages/remediations/src/RemediationWizard/RemediationWizard.js
```diff
@@ -159,7 +180,20 @@ RemediationWizard.propTypes = {
         systems: propTypes.arrayOf(propTypes.string),
         onRemediationCreated: propTypes.func
     }).isRequired,
-    basePath: propTypes.string
+    basePath: propTypes.string,
+    registry: propTypes.object.isRequired
```
Hyperkid123, line 184: ReducerRegistry is not `object` but class. Use `PropTypes.instanceOf(ReducerRegistry)`

Hyperkid123 (commented):
I'd like to see some additional tests. Functions `fetchSystemsInfo` have some pretty complex conditions.

fhlavac:
@Hyperkid123 sure, I'll add the tests in a while, thank you

fhlavac:
@Hyperkid123 enhanced utils.js tests, all complex logic should be covered

Hyperkid123:
@fhlavac should or is?

Hyperkid123:
This bad boy definitely deserves some attention. It should be easy enough to test. Just mount the step and spy on the `fetchSystemsInfo` and its arguments.

```jsx
const playbookSystems = formValues && formValues[EXISTING_PLAYBOOK] && uniqWith(formValues[EXISTING_PLAYBOOK].issues?.reduce((acc, curr) => [
        ...acc,
        ...(curr.systems.map(system => ({ id: system.id, display_name: system.display_name })))
    ], []), isEqual) || [];
```

fhlavac:
@Hyperkid123 finally, I decided to move the logic to utils, created a test for it, and also made some more refactoring.

packages/remediations/src/api/inventory.js
```diff
@@ -0,0 +1,10 @@
+import instance from '@redhat-cloud-services/frontend-components-utilities/interceptors';
+import { HostsApi } from '@redhat-cloud-services/host-inventory-client';
+
+export const INVENTORY_API_BASE = '/api/inventory/v1';
+export const hosts = new HostsApi(undefined, INVENTORY_API_BASE, instance);
+
+export function getHostsById(systems, { page, perPage }) {
+    return hosts.apiHostGetHostById(systems, undefined, perPage, page)
+    .then((response) => response);
```
karelhala, line 9: Extra `then` response.

packages/remediations/src/helpers/host-helper.js
```diff
@@ -0,0 +1,9 @@
+import instance from '@redhat-cloud-services/frontend-components-utilities/interceptors';
```
karelhala, line 1: Duplicate file, it's the same as `packages/remediations/src/api/inventory.js`

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -31,23 +34,40 @@ const ReviewSystems = ({ issues, systems, registry, ...props }) => {
...
-    const playbook = formValues && formValues[EXISTING_PLAYBOOK];
+    const allSystems = useRef([]);
+    const [ allSystemsNamed, setAllSystemsNamed ] = useState([]);
 
-    const allNewSystems = dedupeArray(issues.reduce((acc, curr) => [
-        ...acc,
-        ...(curr.systems || [])
-    ], [ ...systems ]));
+    useEffect(() => {
+        const playbookSystems = formValues && formValues[EXISTING_PLAYBOOK_SELECTED]
```
karelhala, line 52: Safe accessor can be used here
```suggestion
        const playbookSystems = formValues?.[EXISTING_PLAYBOOK_SELECTED]
```

packages/remediations/src/steps/reviewSystems.js
```diff
@@ -31,23 +34,40 @@ const ReviewSystems = ({ issues, systems, registry, ...props }) => {
...
+            : [];
+        setAllSystemsNamed(unionWith(playbookSystems, newSystemsNamed, isEqual));
+        const allNewSystems = dedupeArray(issues.reduce((acc, curr) => [
+            ...acc,
+            ...(curr.systems || [])
+        ], [ ...systems ]));
+        allSystems.current = dedupeArray([
+            ...allNewSystems,
+            ...(playbookSystems.map(system => system.id))
+        ]);
```
karelhala, line 63: You can theoretically merge these two together, but you are sacrificing readability at that point. So your call @fhlavac 
```JS
        allSystems.current = dedupeArray([
            ...dedupeArray(issues.reduce((acc, curr) => [
              ...acc,
              ...(curr.systems || [])
            ], [ ...systems ])),
            ...(playbookSystems.map(system => system.id))
        ]);
```

Hyperkid123 (approved):

packages/remediations/src/utils/utils.js
```diff
@@ -177,18 +179,45 @@ const changeBulkSelect = (state, action) => {
...
 
+export const splitArray = (inputArray, perChunk) => inputArray.reduce((resultArray, item, index) => {
+    const chunkIndex = Math.floor(index / perChunk);
+    if (!resultArray[chunkIndex]) {
+        resultArray[chunkIndex] = [];
+    }
+
+    resultArray[chunkIndex].push(item);
+    return resultArray;
+}, []);
```
karelhala, line 214: This is too complex, you can achieve the same result by using `slice`, you are making too much iterations
```JS
export const splitArray = [...new Array(Math.ceil(inputArray.length/perChunk))].map((_item, key) => inputArray.slice(key * perChunk, (key + 1) * perChunk))
```

You can test it on https://jsben.ch/cD5xN, the results are quite impressive. The sollution with `split` is almost twice as fast for medium arrays. For small ones the difference is just a few percentage, but once we enter the large sized arrays the difference is huge (20 % for the one in this PR).

packages/remediations/src/helpers/host-helper.js
fhlavac: Oh, good catch, thanks for reporting :-)

packages/remediations/src/steps/reviewSystems.js
fhlavac: @karelhala I think we can change it, the names of the variables make it still understandable, thank you

packages/remediations/src/utils/utils.js
fhlavac: @karelhala wow, really cool! Thank you :)

karelhala (approved):

PR #1056, opened 2021-04-01
Visibility prop of delete button in filter chips

leSamo:
This change is backwards compatible.
Added `showDeleteButton` prop, which is necessary for [reset filters functionality](https://issues.redhat.com/browse/VULN-1512) in case no filters applied is not the default state.

Example in Vulnerability -- default filter state is `{ systemsExposed: '1 or more' }` so delete (reset) button is shown only if chips are not in a default state
resetFilters

### DIFF
--- a/packages/components/src/FilterChips/FilterChips.js
+++ b/packages/components/src/FilterChips/FilterChips.js
@@ -4,7 +4,7 @@ import { Badge, Chip, ChipGroup, Button } from '@patternfly/react-core';
 import classNames from 'classnames';
 import './filter-chips.scss';
 
-const FilterChips = ({ className, filters, onDelete, deleteTitle, onDeleteGroup }) => {
+const FilterChips = ({ className, filters, onDelete, deleteTitle, showDeleteButton, onDeleteGroup }) => {
     const groups = filters.filter(group => group.category);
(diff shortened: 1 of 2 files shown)


packages/components/src/FilterChips/FilterChips.js
```diff
@@ -78,13 +79,15 @@ FilterChips.propTypes = {
     ),
     onDelete: PropTypes.func,
     onDeleteGroup: PropTypes.func,
-    deleteTitle: PropTypes.node
+    deleteTitle: PropTypes.node,
+    showDeleteButton: PropTypes.oneOf([ 'default', true, false ])
```
karelhala, line 83: It's kinda strange mixing boolean values and string. Can you use `null` or `undefined` instead of this? I think that the `undefined` will be just awesome! Because we won't have to introduce new default prop and the line 53 would change to:
```JS
{ (showDeleteButton === true || (showDeleteButton === undefined && filters.length > 0)) &&
```

karelhala (approved):
Looking good! Just one minor update and we are golden!

leSamo:
@karelhala, thank you for the review! I have fixed it.

karelhala (approved):

PR #1061, opened 2021-04-07
Advisor report table to have sticky header

AllenBW:
fixes https://issues.redhat.com/browse/ADVISOR-1677


Hyperkid123:
@AllenBW did you tried it in a stable env? I remember other apps had issues with the sticky header in old chrome. You might have to add a little chrome 1 workaround if it does not.

AllenBW:
prod stable has it for advisor recommendations, topics, and systems, havent seen any issues

Hyperkid123 (approved):

AllenBW:
Geez, ok I mispoke, or misread, yer right @Hyperkid123 only workin in beta, but workin in all da beta, chrome safari ff n since that env will be released to stable soon, m ok just leaving it as is, without the shim thing, n seein this show up when the change happens 🤗

Hyperkid123:
Yeah, the "workaround" has to be done in apps anyway.

PR #1062, opened 2021-04-08
Fix broken inv table with title as cells

karelhala:


packages/inventory/src/components/table/EntityTable.js
```diff
@@ -12,6 +12,7 @@ import {
 } from '@patternfly/react-table';
 import { SkeletonTable } from '@redhat-cloud-services/frontend-components/SkeletonTable';
 import { createRows, createColumns } from './helpers';
+import { map } from 'd3-array';
```
Hyperkid123, line 15: I suspect accidental import?

Hyperkid123:
@karelhala PF table destroying the root element eh? What a nice feature.

packages/inventory/src/components/table/EntityTable.js
karelhala: Ah, yeah!

Hyperkid123 (approved):

PR #1065, opened 2021-04-12
Fallback to index.html when file not found

rvsia:


packages/config/README.md
```diff
@@ -165,6 +165,8 @@ const { config: webpackConfig, plugins } = config({
 
 This settings will redirect all requests to `appUrl` to your local `index.html` file and will replace ESI tags for Chrome.
 
+By default, all subroutes are (`/beta/settings/sources/new`) will be redirected to `index.html`, you can disable this behavior by setting `disableFallback: true`.
```
Hyperkid123, line 168: Does this consume xhr requests as well? THis rule should only apply to document-type requests.

Hyperkid123 (approved):

PR #1074, opened 2021-04-16
Make Remediation wizard open faster

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-13626

Made the Remediation wizard appear immediately after clicking on the remediation button by fetching resolutions in the background. Also did some more code cleaning and updated tests.

Before:
gif-before

After:
gif-after


packages/remediations/src/redux/actions/resolution-actions.js
```diff
@@ -0,0 +1,26 @@
+import * as api from '../../api';
+import { FETCH_RESOLUTIONS } from '../action-types';
+import transform from 'lodash/transform';
+
+export const fetchResolutions = (issues = []) => ({
+    type: FETCH_RESOLUTIONS,
+    payload: async () => {
```
karelhala, line 7: The payload shouldn't be a function, can you make this into a const above this action and create the action as
```JS
export const fetchResolutions = (issues = []) => ({
  type: FETCH_RESOLUTIONS,
  payload: batchResolution(issues)
});
```

packages/remediations/src/steps/selectPlaybook.js
```diff
@@ -43,6 +50,8 @@ const SelectPlaybook = (props) => {
     const [ selectedPlaybook, setSelectedPlaybook ] = useState(values[EXISTING_PLAYBOOK]);
     const [ isLoadingRemediation, setIsLoadingRemediation ] = useState(false);
 
+    const { errors = [], warnings = [], resolutions = [], isLoading } = useSelector(({ resolutionsReducer }) => resolutionsReducer);
```
karelhala, line 53: This selector is too complex. I'd break it down to multiple consts and used `shallowEqual`

```JS
const errors = useSelector(({ resolutionsReducer }) => resolutionsReducer?.errors || [], shallowEqual);
const warnings = useSelector(({ resolutionsReducer }) => resolutionsReducer?.warnings || [], shallowEqual);
const resolutions = useSelector(({ resolutionsReducer }) => resolutionsReducer?.resolutions || [], shallowEqual);
const isLoading = useSelector(({ resolutionsReducer }) => resolutionsReducer?.isLoading);
```

I would also be carefull with the `equalityFn` as `shallowEqual`, for errors and warning that might be fine, but for `resolutions` it might be worth writing custom one.

packages/remediations/src/steps/selectPlaybook.js
```diff
@@ -52,115 +61,124 @@ const SelectPlaybook = (props) => {
         fetchData();
     }, []);
 
+    useEffect(() => {
+        if (differenceWith(resolutions, values[RESOLUTIONS], isEqual)?.length > 0) {
```
karelhala, line 65: Yup, as I suspected. The `resolutions` selector pulls the data from redux everytime and hence triggering the re-render multiple times. Although this is for triggering new updated to `formOptions` we could be just fine with the `shallowEqual` in `useSelector` and keeping check here as well (we'll shave off a couple re-renders and make this more performant)

karelhala (commented):
Looking good! Just a few performance updates and this will be blazingly fast!

karelhala (approved):

PR #1075, opened 2021-04-16
feat(compliance): RHICOMPL-1652 sticky header for rules table

vkrizan:
Currently only works within the wizard and modals. Modals need some tweak (will submit it in a bit). The regular pages might need more investigation, which we ultimately fix via RHICOMPL-1628.

Here is the wizard:
Screenshot_2021-04-16 SCAP policies - Compliance Red Hat Insights


akofink (approved):
Bravo! :1st_place_medal:

PR #1078, opened 2021-04-19
Add OS release to inventory table

rvsia:
Adds OS release column to table

**After**

Screenshot 2021-04-20 at 12 57 26

@karelhala Are there any mockups how it should look? 🤔  👀

### DIFF
--- a/packages/inventory/src/api/api.js
+++ b/packages/inventory/src/api/api.js
@@ -92,7 +92,7 @@ export async function getEntities(items, {
                     undefined,
                     {
                         cancelToken: controller && controller.token,
-                        query: { ...generateFilter(fields, 'fields') }
+                        query: generateFilter(fields, 'fields')
                     }
                 );
 
@@ -142,8 +142,8 @@ export async function getEntities(items, {
             {
                 cancelToken: controller && controller.token,
                 query: {
-                    ...generateFilter(options.filter),
-                    ...generateFilter(fields, 'fields')
+                    ...(options.filter && Object.keys(options.filter).length && generateFilter(options.filter)),
+                    ...(fields && Object.keys(fields).length && generateFilter(fields, 'fields'))
                 }
             }
         )
--- a/packages/inventory/src/api/api.js
+++ b/packages/inventory/src/api/api.js
@@ -3,7 +3,7 @@ import flatMap from 'lodash/flatMap';
 export const INVENTORY_API_BASE = '/api/inventory/v1';
 
 import instance from '@redhat-cloud-services/frontend-components-utilities/interceptors';
-import { generateFilter } from '@redhat-cloud-services/frontend-components-utilities/helpers';
+import { generateFilter, mergeArraysByKey } from '@redhat-cloud-services/frontend-components-utilities/helpers';
 import { HostsApi, TagsApi } from '@redhat-cloud-services/host-inventory-client';
 import { defaultFilters } from '../shared/constants';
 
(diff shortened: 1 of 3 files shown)


packages/inventory/src/api/api.js
```diff
@@ -78,6 +78,35 @@ export function getEntities(items, {
 }, showTags) {
     if (hasItems && items.length > 0) {
         return hosts.apiHostGetHostById(items, undefined, perPage, page, undefined, undefined, { cancelToken: controller && controller.token })
+        .then(async (data) => {
```
karelhala, line 81: I would add `fields` field into config, if this field is undefined I'd use default value
```JS
{
  // rest of config
  fields: {
    system_profile: [ 'os_release' ]
  }
}
```

And if this fields is defined and has some values I'd use this `then` clause.

karelhala, line 81: No need to use `then`if we are using `async/wait`
```JS
const invData = await hosts.apiHostGetHostById(items, undefined, perPage, page, undefined, undefined, { cancelToken: controller && controller.token });
return {
   ...invData,
   ...Object.keys(fields).length > 0 && {
    results: mergeArraysByKey([
     invData.results,
     (await hosts.apiHostGetHostSystemProfileById(
       items,
       perPage,
       page,
       orderBy,
       undefined,
       undefined,
       undefined,
       {
         cancelToken: controller && controller.token,
         query: { ...generateFields(fields) }
       }
     })).results
    ])
   }
}
```

packages/inventory/src/api/api.js
```diff
@@ -78,6 +78,35 @@ export function getEntities(items, {
...
+                    items,
+                    perPage,
+                    page,
+                    orderBy,
+                    undefined,
+                    undefined,
+                    undefined,
+                    {
+                        cancelToken: controller && controller.token,
+                        query: { 'fields[system_profile]': [ 'os_release' ]
```
karelhala, line 95: I would either change the `generateFilter` to be named as `queryFromObject` or add another `generateFields` (the function should be generic enough to handle how fields are generated as well)

packages/inventory/src/redux/entities.js
```diff
@@ -36,6 +36,13 @@ export const defaultColumns = [
         title: 'Name',
         renderFunc: TitleColumn
     },
+    {
+        key: 'system_profile',
+        title: 'OS release',
+        // eslint-disable-next-line react/display-name
+        renderFunc: (systemProfile) => systemProfile.os_release || 'Not available',
+        props: { isStatic: true }
```
karelhala, line 44: `width` prop set at least to `30` should be enough to hold the full col name.

karelhala (commented):
I guess that in follw up PR a new filter for OS release will be added, right?

packages/inventory/src/redux/entities.js
rvsia: We can make the column name wrappable

packages/inventory/src/api/api.js
rvsia: I do not understand, to what config?

packages/inventory/src/api/api.js
rvsia: This is so simple I would rather to avoid using external functions from dependencies.

packages/inventory/src/api/api.js
rvsia: Ok, I didn't want to make so big refactor here 😅

rvsia:
@karelhala Yeah, I would update the filter in an another run - but the question is, how the filter should look? I suppose checkboxes would be the best for this case, but I am not sure what variants can be here... (and I am not sure if the api is capable to do this...)

packages/inventory/src/api/api.js
karelhala: In the function definition, we are using first argument as list of IDs, second as config/options and 3rd for showing tags. So when we call this function from consumer side the signature would be `getEntities([], { fields: {some: 'value'} }, false)`

rvsia: alright! make sense, thanks!

packages/inventory/src/api/api.js
karelhala: Yeah, but I was more thinking about future plans, more specifically if some app will use `getEntities` and they'd want to pull in custom fields (there is already a mock fot this in remediations, cloud-connector and edge)

packages/inventory/src/redux/entities.js
karelhala: That could work as well.

rvsia:
@karelhala I updated the code according to all of your comments.

packages/inventory/src/redux/entities.js
```diff
@@ -36,6 +37,14 @@ export const defaultColumns = [
         title: 'Name',
         renderFunc: TitleColumn
     },
+    {
+        key: 'system_profile',
+        title: 'OS release',
+        // eslint-disable-next-line react/display-name
+        renderFunc: (systemProfile) => systemProfile.os_release || 'Not available',
+        props: { isStatic: true },
```
karelhala, line 45: Looking at the screenshot the width of the cell is not that fancy. I'd add the width prop here anyway.
```suggestion
        props: { width: 25, isStatic: true },
```

karelhala (approved):

PR #1082, opened 2021-04-21
Show operating system in label

rvsia:
Screenshot 2021-04-21 at 14 05 39
Screenshot 2021-04-21 at 14 05 48

### DIFF
--- a/packages/inventory/src/redux/entities.js
+++ b/packages/inventory/src/redux/entities.js
@@ -19,6 +19,7 @@ import groupBy from 'lodash/groupBy';
 import TitleColumn from '../components/table/TitleColumn';
 import InsightsDisconnected from '../shared/InsightsDisconnected';
 import { wrappable } from '@patternfly/react-table';
+import OperatingSystemFormatter from '../shared/OperatingSystemFormatter';
 
 export const defaultState = {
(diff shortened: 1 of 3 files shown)


bastilian:
Would it be possible to move the Operating Systems Label component to frontend-components?

We added a similar component to compliance[0]. If we had a common label component we could make sure the OS version colors match across the platform. 

[0] https://github.com/RedHatInsights/compliance-frontend/blob/master/src/PresentationalComponents/OperatingSystemBadge/OperatingSystemBadge.js

karelhala:
@bastilian if you are using inventory component you should show the OS release from inventory table, no need to write custom one. Unless you want to show the Os release badge in some other table.

bastilian:
@karelhala We also show the Label in other none InventoryTable places.

karelhala:
@bastilian oooh, yeah we can def move it to components repo as well. For now we're going to keep it in inventory so this feature is released ASAP and we can move this feature to FEC afterwards.

karelhala (approved):

rvsia:
I would also say this is still a WIP design and it could mess other applications designs. Maybe we should do some iterations before we make it shared.

PR #1083, opened 2021-04-21
Handle duplicate IDs in issue resolution step

fhlavac:
Fixes: https://issues.redhat.com/browse/RHCLOUD-13743

- This fixes the inability to select a resolution in case of two resolutions with the same ID occurring.
- Also fixed `pluralize` utils function for cases when count is 0 and hidden message about other issues in `reviewActions` if there are no other issues.
- Updated tests.

duplicates1
duplicates2


karelhala:
@fhlavac can you take a look at the conflicts?

karelhala (approved):

PR #1084, opened 2021-04-21
Use extension to test (not whole name)

rvsia:
https://webpack.js.org/plugins/source-map-dev-tool-plugin/ 

test is using to test only **extensions**


karelhala (approved):

akofink (approved):
Thank you!!

PR #1086, opened 2021-04-22
fix(compliance): RHICOMPL-1741 filter chip remove

vkrizan:
Reproducible on any non-group types on all Compliance tables.


bastilian (approved):
Nice catch! Thank you @vkrizan!

PR #1087, opened 2021-04-22
Allow Passing custom no results component

bastilian:
In compliance we used to have a warning in the InventoryTable on specific pages under certain conditions, which we injected via the rows in the reducer.
With a recent refactor we removed this. To bring it back and not again inject it via the reducer, it would be nice to be able and pass a custom No Systems Table for the Entity Table to display.

Is this a good approach or could this be done differently?

See this for a usage example and screenshot of what the empty state with the warning should look like: https://github.com/RedHatInsights/compliance-frontend/pull/1170


packages/inventory/src/components/table/__snapshots__/EntityTable.test.js.snap
```diff
@@ -4,7 +4,7 @@ exports[`EntityTable DOM should render correctly - compact 1`] = `
 <table
   aria-label="Host inventory"
   class="ins-c-entity-table pf-c-table pf-m-grid-md pf-m-compact"
-  data-ouia-component-id="OUIA-Generated-Table-20"
+  data-ouia-component-id="OUIA-Generated-Table-22"
```
bastilian, line 7: I'm unsure why these change and if this is a valid change. I only pushed the snapshot with the results of the test added here, because I thought it might be a different patternfly version difference or something.

packages/inventory/src/components/table/helpers.js
```diff
@@ -18,11 +18,11 @@ export const buildCells = (item, columns, extra) => {
     });
 };
 
-export const createRows = (rows = [], columns = [], { actions, expandable, ...extra } = {}) => {
+export const createRows = (rows = [], columns = [], { actions, expandable, customNoSystemsTable, ...extra } = {}) => {
     if (rows.length === 0) {
         return [{
             cells: [{
-                title: <NoSystemsTable />,
+                title: customNoSystemsTable ? customNoSystemsTable : <NoSystemsTable />,
```
rvsia, line 25: The same as in the previous PR, we can skip `custom` from props and make the default component as the default prop.

packages/inventory/src/components/table/__snapshots__/EntityTable.test.js.snap
rvsia: Yeah, this is valid. PF generates ids automatically and deterministically, and because you added new test cases, there are now more components rendered in the same “instance” so the global ID pool is now bigger. If you run this test seperately, the number will be 1.

... and that’s one of the reasons I don’t like snapshot tests. 😄

rvsia (commented):
Looks great! 🍹

packages/inventory/src/components/table/__snapshots__/EntityTable.test.js.snap
bastilian: :D oh, ok. yeah, the snapshot diffs can get excessive.

packages/inventory/src/components/table/EntityTable.js
```diff
@@ -139,7 +142,8 @@ EntityTable.propTypes = {
         [PropTypes.string]: PropTypes.any
     }),
     onRowClick: PropTypes.func,
-    showTags: PropTypes.bool
+    showTags: PropTypes.bool,
+    customNoSystemsTable: PropTypes.node
```
rvsia, line 146: Old name

rvsia (approved):

packages/inventory/src/components/table/helpers.test.js
```diff
@@ -42,15 +42,11 @@ describe('createRows', () => {
     const cells = [{ key: 'first' }, { key: 'second.dot' }, { key: 'third' }];
     it('should create empty table', () => {
         const data = createRows();
-        const wrapper = mount(data[0].cells[0].title);
```
bastilian, line 45: The component is now outside the helpers, the tests still pass and the actual rendering of the `NoSystemsResults` component is tested in the `InventoryTable` tests.

karelhala (approved):
Looking good!

PR #1089, opened 2021-04-23
fix(RulesTable): RHICOMPL-978 - Change passed filter to checkboxes

bastilian:


packages/inventory-compliance/src/Utilities/FilterBuilderConfigBuilder.js
```diff
@@ -43,13 +43,13 @@ const BASE_FILTER_CONFIGURATION = [
 ];
 
 const PASS_FILTER_CONFIG = {
-    type: conditionalFilterType.radio,
+    type: conditionalFilterType.checkbox,
     label: 'Passed',
     items: [
         { label: 'Show passed rules', value: 'passed' },
         { label: 'Hide passed rules', value: 'failed' }
```
akofink, line 50: Slight discrepancy with the mocks here:

```suggestion
        { label: 'Passed rules', value: 'passed' },
        { label: 'Failed rules', value: 'failed' }
```

akofink (approved):
Works great! Just one small text change to align with the mocks.

Also, note that I asked a question to PM on the Jira, since the mocks and the PM comment there don't match, so we may want to wait to merge this until we get a response.

bastilian:
@akofink Checkbox are ✅ from PM. Thank you for testing and reviewing!

PR #1090, opened 2021-04-23
Transform jwt cookie to x-hr-identity header

rvsia:
### cookieTransform
For running local services you can use `cookieTransform` ([original function](https://github.com/RedHatInsights/insights-standalone/blob/1eef6cfc21f96304275683d090c6b8178a4d386f/index.js#L8), you can also check [insights-proxy implementation](https://github.com/RedHatInsights/insights-proxy/blob/1cdbc597681eac51998d8c2dd2dd6b5a2d4d03d6/spandx.config.js#L101)) in `onProxyReq` function. This function transform `jwt` cookie to `x-hr-identity` header.
```jsx
const cookieTransform = require('@redhat-cloud-services/frontend-components-config/src/cookieTransform');
onProxyReq: (...args) => {
    cookieTransform(...args);
},
```

Routes passed via `routes` or `routesFile` attributes are using this transform automatically. If you override the `onProxyReq` function, you have to add it back manually.

#### custom entitlements

By default, the same entitlements as in insights-proxy are provided. You can rewrite them via the options object:

```jsx
cookieTransform(proxyReq, req, res, { entitlements });
```

You can also modify the whole identity object:

```jsx
cookieTransform(proxyReq, req, res, { entitlements, identity, user, internal });
```

### DIFF
--- a/packages/config/src/cookieTransform.js
+++ b/packages/config/src/cookieTransform.js
@@ -11,7 +11,7 @@ const defaultEntitlements = {
     ansible: { is_entitled: true }
 };
 
-function cookieTransform(proxyReq, req, _res, { entitlements = defaultEntitlements }) {
+function cookieTransform(proxyReq, req, _res, { entitlements = defaultEntitlements, user, internal, identity: customIdentity }) {
     const cookie = req.headers.cookie;
     const match = cookie && cookie.match(/cs_jwt=([^;]+);/);
     if (match) {
@@ -24,6 +24,7 @@ function cookieTransform(proxyReq, req, _res, { entitlements = defaultEntitlemen
                 type: 'User',
                 auth_type: 'basic-auth',
                 account_number: payload.account_number + '',
+                ...customIdentity,
                 user: {
                     username: payload.username,
                     email: payload.email,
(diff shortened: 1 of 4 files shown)


packages/config/src/cookieTransform.js
```diff
@@ -0,0 +1,46 @@
...
+        const cs_jwt = match[1];
+        const { payload } = jws.decode(cs_jwt);
+        const identity = {
+            entitlements: {
+                insights: { is_entitled: true },
+                smart_management: { is_entitled: true },
+                openshift: { is_entitled: true },
+                hybrid: { is_entitled: true },
+                migrations: { is_entitled: true },
+                ansible: { is_entitled: true }
```
Hyperkid123, line 18: I think we should make these objects configurable. If somebody wants to use different identity than "super user".

rvsia: added customization

Hyperkid123:
@rvsia makes me think if we should allow the user data configuration as well. You know if, you want all permissions but not be org admin or have a different username, etc,. I think we could do that. As it is an opt-in config nothing will break if users keep the default values.

Hyperkid123 (approved):

packages/config/src/proxy.js
```diff
@@ -165,6 +166,9 @@ function createInsightsProxy({
                     changeOrigin: true,
                     autoRewrite: true,
                     ws: true,
+                    onProxyReq: (...args) => {
+                        cookieTransform(...args);
```
karelhala, line 170: Once consumer uses `proxyRoutes` they can't really change what are the custom entitlements. I would call `redirect.onProxyReq` if it is a function.
```suggestion
                        typeof redirect.onProxyReq === 'function' ? redirect.onProxyReq(...args) : cookieTransform(...args);
```

karelhala (commented):
Consumers can't really change what `onProxyReq` is called when using the proxyRoutes. Actually, it might be reasonable to spread the entire `redirect` after our config, so consumers can pass custom proxy config to each route.

karelhala (approved):

PR #1093, opened 2021-04-23
fix(RulesTable): Fix Ansible sorting with variable columns

bastilian:
The `property` property comes from PatternFly, which, due to the title being a React node, uses just `column-INDEX` for it.
Since the columns can be different the index of the column can vary, this works around the issue und fixes sorting for the Ansible column in rules tables.


akofink (approved):
Works great, Thanks!

PR #1098, opened 2021-04-27
Refactor ContextInventoryList to functional component

rvsia:


rvsia:
@karelhala  I updated the code, the function is now passed to ref directly. It seems working the same as before:

Kapture 2021-04-28 at 09 58 02

packages/inventory/src/components/table/InventoryList.js
```diff
@@ -11,87 +11,87 @@ import AccessDenied from '../../shared/AccessDenied';
...
+            page,
+            perPage,
+            items,
+            hasItems,
+            sortBy,
+            activeFilters,
+            ...customFilters,
+            ...options
+        }, showTags);
+    };
```
karelhala, line 74: Might be reasonable to combine these 2 functions together. The `loadEntities` function is not used anywhere else.

karelhala (approved):
Looking good! Just one minor nitpick, but this can be ignored.

PR #1100, opened 2021-04-28
Allow passing custom props to BulkSelect items

bastilian:
This is just a nicety. When working on the BulkSelect integration for compliance (via https://github.com/RedHatInsights/frontend-components/pull/919) I thought it would be nice to be able and disable the "Select none" if none are selected, but there wasn't a way yet to pass in the necessary prop. 

Screenshot 2021-04-28 at 05 15 57

### DIFF
--- a/packages/components/src/BulkSelect/BulkSelect.js
+++ b/packages/components/src/BulkSelect/BulkSelect.js
@@ -72,6 +72,7 @@ visible unless you update it.');
                             component="button"
                             key={ oneItem.key || key }
                             onClick={ (event) => oneItem.onClick && oneItem.onClick(event, oneItem, key) }
+                            { ...oneItem?.props }
                         >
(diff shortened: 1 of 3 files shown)


packages/components/src/BulkSelect/BulkSelect.js
```diff
@@ -72,6 +72,7 @@ visible unless you update it.');
                             component="button"
                             key={ oneItem.key || key }
                             onClick={ (event) => oneItem.onClick && oneItem.onClick(event, oneItem, key) }
+                            { ...oneItem?.props && oneItem.props }
```
rvsia, line 75: `{ ...oneItem?.props }` should be enough I think.

rvsia (approved):

Hyperkid123 (approved):

PR #1103, opened 2021-04-29
Refactor onRefreshData to be controlled from one place

rvsia:
onRefreshData is declared only on one place. This makes the function consistent (there should be no bugs that some attributes are missing when the function is called from different places) and it makes all future changes much more manageable.

### DIFF
--- a/packages/inventory/src/components/table/InventoryList.js
+++ b/packages/inventory/src/components/table/InventoryList.js
@@ -1,10 +1,9 @@
 import React, { useEffect, useRef } from 'react';
-import { useDispatch, useSelector } from 'react-redux';
+import { useSelector } from 'react-redux';
 import InventoryEntityTable from './EntityTable';
 import { Grid, GridItem } from '@patternfly/react-core';
 import PropTypes from 'prop-types';
 import './InventoryList.scss';
-import { loadSystems } from '../../shared';
 import isEqual from 'lodash/isEqual';
 import AccessDenied from '../../shared/AccessDenied';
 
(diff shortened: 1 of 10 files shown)


packages/inventory/src/components/table/InventoryList.js
```diff
@@ -41,44 +40,18 @@ const ContextInventoryList = ({ showHealth, onRefreshData, ...props }) => {
...
-                    ...options
-                },
-                showTags,
-                getEntities
-            )
-        );
-    };
-
-    if (ref) { ref.current = { onRefreshData }; }
+    if (ref) { ref.current = { onRefreshData: (params, disableRefresh = true) => onRefreshData(params, disableRefresh) }; }
```
karelhala, line 54: I would span this over a couple lines
```suggestion
    if (ref) {
      ref.current = {
        onRefreshData: (params, disableRefresh = true) => onRefreshData(params, disableRefresh)
      };
    }
```

packages/inventory/src/components/table/InventoryTable.js
```diff
@@ -56,6 +58,50 @@ const InventoryTable = forwardRef(({
     const sortBy = useSelector(({ entities: { sortBy: invSortBy } }) => (
         hasItems ? propsSortBy : invSortBy
     ), shallowEqual);
+    const dispatch = useDispatch();
+    const store = useStore();
+
+    /**
+     * If conumer wants to change data they can call this function via component ref.
```
karelhala, line 65:
```suggestion
     * If consumer wants to change data they can call this function via component ref.
```

karelhala (approved):
Looking good!

packages/inventory/src/components/table/InventoryList.js
rvsia: I want to move this logic to the upper component in the follow-up pr.

(And also, we should set eslint if really want to change formatting. 😄  )

PR #1104, opened 2021-04-29
Add ErrorState to InventoryTable

bastilian:
This adds an `ErrorState` for when a `LOAD_ENTITIES_REJECTED` is dispatched and allows to provide a custom component as an ErrorState if necessary.

https://user-images.githubusercontent.com/7757/116590329-e79e7780-a91d-11eb-91ef-bf0775def4a3.mp4

### DIFF
--- a/packages/inventory/src/components/table/InventoryTable.js
+++ b/packages/inventory/src/components/table/InventoryTable.js
@@ -2,6 +2,7 @@ import React, { Fragment, forwardRef } from 'react';
 import { useSelector, shallowEqual } from 'react-redux';
 import EntityTableToolbar from './EntityTableToolbar';
 import { TableToolbar } from '@redhat-cloud-services/frontend-components/TableToolbar';
+import { ErrorState } from '@redhat-cloud-services/frontend-components/ErrorState';
 import InventoryList from './InventoryList';
 import Pagination from './Pagination';
 import AccessDenied from '../../shared/AccessDenied';
@@ -30,9 +31,13 @@ const InventoryTable = forwardRef(({
     getEntities,
     hideFilters,
     paginationProps,
+    errorState = <ErrorState />,
     ...props
 }, ref) => {
     const hasItems = Boolean(items);
+    const error = useSelector(({ entities }) => (
+        entities?.error
+    ));
     const page = useSelector(({ entities: { page: invPage } }) => (
         hasItems ? propsPage : (invPage || 1)
     )
@@ -65,7 +70,7 @@ const InventoryTable = forwardRef(({
                     To view the content of this page, you must be granted a minimum of inventory permissions from your Organization Administrator.
                 </div>}
             /> :
-            <Fragment>
+            !error ? <Fragment>
                 <EntityTableToolbar
                     { ...props }
                     customFilters={customFilters}
@@ -118,7 +123,7 @@ const InventoryTable = forwardRef(({
                         paginationProps={paginationProps}
                     />
                 </TableToolbar>
-            </Fragment>
+            </Fragment> : errorState
     );
 });
 
--- a/packages/inventory/src/redux/entities.js
+++ b/packages/inventory/src/redux/entities.js
@@ -243,7 +243,7 @@ export default {
     [ACTION_TYPES.LOAD_ENTITIES_REJECTED]: loadingRejected,
     [ACTION_TYPES.LOAD_TAGS_PENDING]: showTagsPending,
     [ACTION_TYPES.LOAD_TAGS_FULFILLED]: showTags,
-    [ACTION_TYPES.LOAD_TAGS_REJECTED]: loadingRejected,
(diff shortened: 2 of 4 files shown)


packages/inventory/src/components/table/InventoryTable.js
```diff
@@ -118,7 +123,7 @@ const InventoryTable = forwardRef(({
                         paginationProps={paginationProps}
                     />
                 </TableToolbar>
-            </Fragment>
+            </Fragment> : (customErrorState ? customErrorState : <ErrorState />)
```
rvsia, line 126: You can put ErrorState as the default prop for customErrorState so there is no need for conditional logic.

Also, what about callimg it just `errorState`, it’s prop, so it’s custom by default. 😄

packages/inventory/src/redux/entities.js
```diff
@@ -233,6 +240,7 @@ export default {
     [ACTION_TYPES.ALL_TAGS_PENDING]: (state) => ({ ...state, allTagsLoaded: false }),
     [ACTION_TYPES.LOAD_ENTITIES_PENDING]: entitiesPending,
     [ACTION_TYPES.LOAD_ENTITIES_FULFILLED]: entitiesLoaded,
+    [ACTION_TYPES.LOAD_ENTITIES_REJECTED]: entitiesLoadingRejected,
```
rvsia, line 243: Should we also add it for `LOAD_TAGS` action? 🤔

rvsia (commented):
Looks great! 🏆  just two suggestions

packages/inventory/src/components/table/InventoryTable.js
bastilian: Nice suggestion! Updated.

rvsia (approved):

packages/inventory/src/redux/entities.js
bastilian: Should we? In compliance we don't use the global filter yet, so I'm not sure how that deals with rejection, but if it also causes the inventory to get stuck in the loading state it should definitely be added.

packages/inventory/src/components/table/InventoryTable.js
```diff
@@ -30,9 +31,13 @@ const InventoryTable = forwardRef(({
     getEntities,
     hideFilters,
     paginationProps,
+    errorState = <ErrorState />,
     ...props
 }, ref) => {
     const hasItems = Boolean(items);
+    const error = useSelector(({ entities: { error } }) => (
+        error
+    ));
```
karelhala, line 40: I would use safeAccessor here. It could happen that entities are not an object if app missregistered the reducder
```suggestion
    const error = useSelector(({ entities }) => (
        entities?.error
    ));
```

packages/inventory/src/redux/entities.js
karelhala: Yup, good point. If consumer wants to show tags in table and the request fails it would cause the table to be stuck in infinit loading.

karelhala (commented):
Looking good! Just one minor nitpick.

packages/inventory/src/redux/entities.js
rvsia: So also for `ALL_TAGS_REJECTED`.

bastilian:
@rvsia @karelhala Thank you both for taking a look at my PRs! I updated this on to add `LOAD_TAGS_REJECTED` and named the function more generic.

packages/inventory/src/redux/entities.js
```diff
@@ -233,8 +240,10 @@ export default {
     [ACTION_TYPES.ALL_TAGS_PENDING]: (state) => ({ ...state, allTagsLoaded: false }),
     [ACTION_TYPES.LOAD_ENTITIES_PENDING]: entitiesPending,
     [ACTION_TYPES.LOAD_ENTITIES_FULFILLED]: entitiesLoaded,
+    [ACTION_TYPES.LOAD_ENTITIES_REJECTED]: loadingRejected,
     [ACTION_TYPES.LOAD_TAGS_PENDING]: showTagsPending,
     [ACTION_TYPES.LOAD_TAGS_FULFILLED]: showTags,
+    [ACTION_TYPES.LOAD_TAGS_REJECTED]: loadingRejected,
```
karelhala, line 246: This should be `ALL_TAGS_REJECTED`
```suggestion
    [ACTION_TYPES.ALL_TAGS_REJECTED]: loadingRejected,
```

karelhala (approved):

packages/inventory/src/redux/entities.js
bastilian: 🤦Thank you! :D

rvsia: Shouldn't it be both?

karelhala: @rvsia not really, the `LOAD_TAGS` is for single system tags. `ALL_TAGS` are for filter tags, if this one fails the table is still present.

PR #1105, opened 2021-04-30
Allow to pass custom columns as a prop

rvsia:
- also allow to disable the default ones

## columns

*array*

An array of columns definitions. They are merged with default columns by their keys.

## disableDefaultColumns

*boolean | string[]*

You can disable default columns. If you set this prop to `true`, then all the default columns are disabled. Or you can provide an array of keys to disable just specific ones: `[ 'display_name', 'last_seen' ]`.

**After**

```jsx
      disableDefaultColumns={['display_name']}
      columns={
          [{
              key: 'display_name_1',
              title: 'Different name',
              renderFunc: (name, id, { display_name }) => <h1>{display_name}</h1>
          }
          ]
      }
```

Screenshot 2021-04-30 at 11 41 33

### DIFF
--- a/packages/inventory/src/components/table/EntityTable.js
+++ b/packages/inventory/src/components/table/EntityTable.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo, useRef } from 'react';
 import PropTypes from 'prop-types';
 import { selectEntity, setSort } from '../../redux/actions';
 import { useDispatch, useSelector, shallowEqual } from 'react-redux';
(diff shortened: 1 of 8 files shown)


rvsia:
cc @bastilian

bastilian:
Would it maybe make sense to just completely override the default columns when the columns prop is set?

Default columns could be made available to import and allow mixing the columns and also dictate their order.
We have had instances where UX/PM asked for a specific order or columns in tables including the systems/inventory table.

There is something along those lines implemented here: https://github.com/RedHatInsights/compliance-frontend/pull/1005/files#diff-e0494b62ef3cc7ba3308fa54b1933e5d3c299e030a4ac5adaef91ef23630a08dR10

rvsia:
@bastilian So this is currently possible. Just set `disableDefaultFilters` and then import the default ones (they are available as export) and change them as you need in the columns.

packages/inventory/src/components/table/EntityTable.js
```diff
@@ -62,6 +67,16 @@ const EntityTable = ({
...
 
+    const disabledColumns = Array.isArray(disableDefaultColumns) ? disableDefaultColumns : [];
+    const defaultColumnsFiltered = defaultColumns.filter(({ key }) =>
+        (key === 'tags' && showTags) || (key !== 'tags' && !disabledColumns.includes(key))
+    );
+
+    const columns = mergeArraysByKey([
+        typeof disableDefaultColumns === 'boolean' && disableDefaultColumns ? [] : defaultColumnsFiltered,
+        columnsProp || columnsRedux || []
+    ], 'key');
```
karelhala, line 78: This is quite a lot of heavy computing, might be reasonable to move all of this `useEffect` where you could use the check from `useSelector` of `columnsRedux` and store the value in `ref`. We don't want to trigger yet another re-render, but the calculation could be usefull to moize this way. Or we can use `useMemo`. But the memoize function would be a bit complex.

Either way it should be reasonable to somehow limit the amount of calculations for this piece of code as it uses `filter` and `mergeArraysByKey` at each re-render.

karelhala (commented):
Looking hella good! Just a bit of optimization would be nice.

packages/inventory/src/components/table/EntityTable.js
rvsia: done!

karelhala (approved):

AllenBW:
Any update on when we might see this work in prod beta/stable?

karelhala:
@AllenBW we are releasing chrome to all envs this sprint, so it should land in prod-stable by end of this week in best case scenario.

AllenBW:
ah! great news, thanks for the update 🤗

PR #1106, opened 2021-04-30
Do not ignore doc folders

bastilian:
A few packages have a doc folder, it should probably not be ignored.


bastilian:
Closing, missed that https://github.com/RedHatInsights/frontend-components/pull/1105/files already also addressed that. :D

PR #1111, opened 2021-05-05
Set tagsModalLoaded when fetching all tagts

rvsia:
Loading logic was changed in https://github.com/RedHatInsights/frontend-components/commit/f79e6966fadb63112eae90ae00f1f565da4abb05

This sets `tagsModalLoaded` also when loading allTags.

**Before**

Screenshot 2021-05-05 at 8 35 21

**After**

ezgif com-gif-maker

We need this to backport to older environments as well @karelhala


rvsia:
cc @gmcculloug

Hyperkid123 (approved):

PR #1112, opened 2021-05-05
Add SelectOption for a group

bastilian:
This is implementing the option to enable a `SelectOption` for a whole group to be selectable. 
It should only be enabled if the group has a `groupSelectable` option set. 
The event handler will be passed all items in the group as selected.

**Example:**

https://user-images.githubusercontent.com/7757/118270239-0c6b2100-b4c0-11eb-97ce-0be8045dc8f3.mp4

### DIFF
--- a/packages/components/src/ConditionalFilter/GroupFilter.js
+++ b/packages/components/src/ConditionalFilter/GroupFilter.js
@@ -173,19 +173,60 @@ class Group extends Component {
         };
     }
 
-    onSelect = (event, group, item, groupKey, itemKey) => {
-        let newSelection = this.calculateSelected(group, groupKey, itemKey);
-        const { onChange } = this.props;
-
+    setNewSelection = (onChange, event, newSelection, group, item, groupKey, itemKey) => {
         if (onChange) {
-            onChange(event, newSelection, group, item, groupKey, itemKey);
+            if (group) {
+                onChange(event, newSelection, group, item, groupKey, itemKey);
+            } else {
+                onChange(event, newSelection);
+            }
+
             this.setState({ selected: {} });
         }
 
         this.setState({
             selected: newSelection
         });
+    }
+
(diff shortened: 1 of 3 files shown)


packages/components/src/ConditionalFilter/GroupFilter.js
```diff
@@ -267,21 +309,37 @@ class Group extends Component {
...
+                                            } }>
+                                            <Checkbox
+                                                isChecked={ this.isGroupChecked(groupValue || groupKey, items) }
+                                                label={ groupLabel } />
+                                        </SelectOption>
+                                    }
+                                    <SelectGroup
+                                        {...group}
+                                        className='pf-u-pl-sm'
+                                        label={ !groupSelectable && groupLabel }
```
karelhala, line 337: I'm kinda sceprical about this field. PF has hard check for `props.label.replace` without safe accessors. We've had problems with it in the past. The worst thing is that you can't have nodes as label (so no translations or the select option from above in here).

packages/components/src/ConditionalFilter/GroupFilter.js
```diff
@@ -267,21 +309,37 @@ class Group extends Component {
...
-                            return (<SelectGroup
-                                {...group}
-                                key={groupId || groupValue || groupKey}
-                                value={groupId || groupValue || groupKey}
-                                label={groupLabel || ''}
-                                id={groupId || `group-${groupValue || groupKey}`}
-                            >{filteredItems}</SelectGroup>);
+
+                            return (
+                                <div key={groupId || groupValue || groupKey}>
```
karelhala, line 323: Let's hope that this won't break the UI (PF select has ways of breaking stuff). With the new menu component we should sink in some time and switch this component to the new approach. That way we won't have all these problems.

karelhala (approved):
:crossed_fingers:

PR #1124, opened 2021-05-07
Split docs into sections

Hyperkid123:
### Changes
- Move FEC generated docs under /fec route
- Added new sections to the root page (not final we just added some data so we can see it)
- Added breadcrumbs
- Most importantly, added a proper favicon

Screenshot from 2021-05-07 12-03-35

screenshot-localhost_3000-2021 05 07-12_16_00

screenshot-localhost_3000-2021 05 07-12_14_09


packages/docs/components/navigation/components-navigation.js
```diff
@@ -0,0 +1,106 @@
...
+                        <a className={classnames('pf-c-nav__link', {
+                            'pf-m-current': props.href === pathname
+                        })}>
+                            {children}
+                        </a>
+                    </Link>
+                )}
+            >
+                {title}
+            </NavItem>
```
karelhala, line 44: This is fairly the same as is lines 55-76. Might want to be good to make this into a component

karelhala (commented):
I really like the idea of sections! Can we generate the config file from folder structure? This way if new section will be needed it will be automatically generated.

Hyperkid123:
> I really like the idea of sections! Can we generate the config file from folder structure? This way if new section will be needed it will be automatically generated.

I think we could. I would like to wait with any "advanced" stuff until we have a clear outline of the structure of the doc and also the content itself. I don't want to end up in a trap as we did with the CSC and the re-do it all gain.

PR #1127, opened 2021-05-10
Fix systems seletion in Remediation Wizard

fhlavac:
Fixes: https://issues.redhat.com/browse/VULN-1656 & https://issues.redhat.com/browse/RHCLOUD-13891

Fixed issues caused by forgetting which systems were assigned to which issues and applying all issues to all systems. Now issues are being applied just to corresponding systems reflecting previously set systems. Also added expandable table to `reviewActions` step and `review` step showing names of NEW (not previously selected systems in case of enhancing existing playbook) systems on which the remmediation will be applied to. Finally, fixed broken sorting in tables because of wrong indexes in sortByState and updated tests.

@katierik I will create a separate PR to use InventoryTable (allowing to display Tags and so on) in the expandable section with systems. This needs some more changes.

reviewActions_final
review

@karelhala

### DIFF
--- a/packages/remediations/src/RemediationWizard/RemediationWizard.js
+++ b/packages/remediations/src/RemediationWizard/RemediationWizard.js
@@ -73,12 +73,14 @@ const RemediationWizard = ({
         'select-playbook': {
             component: SelectPlaybook,
             issues: data.issues,
+            systems: data.systems,
             allSystems: allSystems.current
         },
         'review-systems': {
             component: ReviewSystems,
             issues: data.issues,
             systems: data.systems,
+            allSystems: allSystems.current,
             registry
         },
         'review-actions': {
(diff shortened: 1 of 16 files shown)


packages/remediations/src/RemediationWizard/RemediationWizard.js
```diff
@@ -97,7 +99,7 @@ const RemediationWizard = ({
 
     const validatorMapper = {
         'validate-systems': () => (value) => (
-            value && value.length > 0
+            value && Object.keys(value).length > 0
```
Hyperkid123, line 102: Here I would probably look at `Object.values(value).filter(value => typeof value !== undefined).length`. If its possible that value can be `undefined` or `null` or whatever is the value after `select value -> de select it -> select other one`.

karelhala (approved):

PR #1134, opened 2021-05-13
chore(config): cleanup

redallen:
Simplify some config and fix typos. This is in preparation of porting `insights-standalone` to `frontend-components-config`.
- Single source of truth for `isProduction`
- Find webpack entry file by globbing and allow single entrypoint file or customizing it. 
- Remove unused `koa-connect`
- Replace `style-loader` with `MiniCSSExtract`
  - Greatly reduces number of `<style>` tags in `<head>`
  - MiniCSSExtract added dev mode 2 years ago
- Remove hot module replacement
  - Setup is quite involved to work with React. Can only be used with additional packages and config. Even when that is configured it still won't work on class components.
- Cleanup federated module package sharing to use a list
- Allow React 17 in peer deps (monorepo is using React 17 for tests and builds, so it's working!)
- Fix Dart SASS deprecation warning


packages/config/src/config.js
```diff
@@ -25,34 +26,39 @@ module.exports = ({
...
+    isProd
 } = {}) => {
     const filenameMask = `js/[name]${useFileHash ? '.[chunkhash]' : ''}.js`;
     return {
-        mode: mode || (process.env.NODE_ENV === 'production' ? 'production' : 'development'),
+        mode: mode || (isProd ? 'production' : 'development'),
         devtool: false,
         optimization: {
-            minimize: (process.env.NODE_ENV || mode) === 'production',
+            runtimeChunk: true,
```
Hyperkid123, line 37: @redallen have tried this with federated modules? I know that using option `single` will rip out the webpack init code from the federated module entry point so it's impossible to use in a different container. I am not sure about `multiple` though.

EDIT: It seems to have broken the `ChunkMapper` used for generating the `fed-mods.json`. 
```
TypeError: Cannot read property 'modules' of undefined
    at /home/martin/insights/insights-inventory-frontend/node_modules/@redhat-cloud-services/frontend-components-config/chunk-mapper.js:28:69

```

When added a condition to leave the function if the module is not found, no output was generated.

packages/config/src/config.js
```diff
@@ -167,7 +168,6 @@ module.exports = ({
             disableFallback
         }) : {
             contentBase: `${rootFolder || ''}/dist`,
-            hot: true,
```
Hyperkid123, line 170: Can you leave it as an option?

packages/inventory/package.json
```diff
@@ -26,8 +26,8 @@
     },
     "homepage": "https://github.com/RedHatInsights/frontend-components/tree/master/packages/inventory#readme",
     "peerDependencies": {
-        "react": "^16.14.0",
-        "react-dom": "^16.14.0",
+        "react": ">=16.5.1 || >=17.0.0",
+        "react-dom": ">=16.5.1 || >=17.0.0",
```
Hyperkid123, line 30: I don't think we can go below 16.8.x, the package is using react hooks.

packages/pdf-generator/src/components/PDFDocument.js
```diff
@@ -7,7 +7,8 @@ import {
     Document
 } from '@react-pdf/renderer';
 import RHLogo from './Logo';
-import { DateFormat } from '@redhat-cloud-services/frontend-components/DateFormat';
+// eslint-disable-next-line rulesdir/disallow-fec-relative-imports
+import { DateFormat } from '@redhat-cloud-services/frontend-components';
```
Hyperkid123, line 11: Did some build errors occur?

packages/translations/package.json
```diff
@@ -33,7 +33,7 @@
     },
     "peerDependencies": {
         "prop-types": "^15.6.2",
-        "react": "^16.5.1"
+        "react": "^16.5.1 || ^17.0.0"
```
Hyperkid123, line 36: Can you bump the minimum to 16.8.0?

packages/pdf-generator/src/components/PDFDocument.js
redallen: Yes there are errors with the UMD build on node v14.16.0:
> @redhat-cloud-services/frontend-components-pdf-generator: Error: 'DateFormat' is not exported by ../components/DateFormat/index.js, imported by src/components/PDFDocument.js

packages/config/src/config.js
redallen: Removed.

packages/config/src/config.js
redallen: Have you found a case where this works and is useful? With a minimal config I've never seen this work with React components.

packages/inventory/package.json
redallen: Fixed.

packages/translations/package.json
redallen: Looks like you already did in #1147

Hyperkid123: Yeah sorry about that, I was upgrading one repo to npm 7 and it was actually throwing installation errors.

Hyperkid123:
@redallen one test file needs an update: https://travis-ci.com/github/RedHatInsights/frontend-components/builds/226831570#L2286

Hyperkid123 (approved):

PR #1137, opened 2021-05-24
de-duplicate config utils package

Hyperkid123:


packages/config/package.json
```diff
@@ -17,6 +17,7 @@
     },
     "homepage": "https://github.com/RedHatInsights/frontend-components/tree/master/packages/config#readme",
     "dependencies": {
+        "@redhat-cloud-services/frontend-components-config-utilities": "^1.0.0",
```
karelhala, line 20: I would use `>=` in here.
```suggestion
        "@redhat-cloud-services/frontend-components-config-utilities": ">=1.0.0",
```

karelhala (approved):

packages/config/package.json
Hyperkid123: I dunno. I would assume that 2.x will have breaking changes which may mess up the end user.

PR #1159, opened 2021-06-01
Fixing404 wording

arivepr:
This card is to close issue RHCLOUD-14098


rvsia:
@arivepr you have included `.DS_STORE` file. Please, remove it. You can also set up [global gitignore](https://docs.github.com/en/github/getting-started-with-github/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) to ignore it.

packages/components/src/Components/Dark/Dark.d.ts
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react';
```
rvsia, line 1: I think all these declaration type files are here by mistake, aren’t they?

arivepr: I'm not entirely sure tbh, haven't touched these files myself.

arivepr:
@rvsia tookout the DS_store file. It shows up on the local gitIgnore, don't know how it nudged itself in.

packages/components/src/Components/Dark/Dark.d.ts
rvsia: Yep, then it should not be here too.

Hyperkid123 (commented):
@arivepr I think you had a very very old version of the repository locally and you did not clean it up after the update. None of the `*.d.ts` should be there. In fact the folder `/src/Components/*` no longer exists.

PR #1154, opened 2021-05-31
Use InventoryTable in RemediationWizard expandable tables 

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-14348

- moved InventoryTable to a separate reusable SystemsTable component
- added the SystemsTable component to reviewSystems step
- added the SystemsTable component to the reviewAction and review steps' expandable tables
- updated and enhanced tests

systems02

systems01

@katierik @karelhala

PR #1160, opened 2021-06-02
Update expandable SystemsTables styling

fhlavac:
UX changes from https://github.com/RedHatInsights/frontend-components/pull/1154

**Before:**
before

**After:**
after

@katierik is it correct this way?

@epwinchell stole a bit of your code from https://github.com/RedHatInsights/insights-remediations-frontend/pull/249 it probably needs to be added there separately


epwinchell (approved):
> @epwinchell stole a bit of your code from RedHatInsights/insights-remediations-frontend#249 it probably needs to be added there separately

@fhlavac Looks good!

rvsia (approved):

PR #1166, opened 2021-06-08
Update all config dependencies

rvsia:
This updates dependencies for the config package, only minor versions. There are some major versions, but they are requiring node 12+ (ts-loader, sass-loader)


Hyperkid123:
@rvsia I would not be afraid to update dependencies that require node 12+. Node 10 support has ended in April. Now we have node 12 in maintenance and node 14 as current lts. I would a node engine requirement of >=12 to the package.json and update mentioned dependencies as well.

rvsia:
@Hyperkid123 I would rather keep a change like that for minor version. And we should probably tell application teams to prepare before we do that.

Hyperkid123 (approved):

PR #1168, opened 2021-06-09
Fix DateFormat import in PDF generator

leSamo:
Without this I was getting
`Cannot find module '@redhat-cloud-services/frontend-components/components/cjs/DateFormat.js' from 'node_modules/@redhat-cloud-services/frontend-components-pdf-generator/dist/esm/index.js'`


rvsia (approved):

Hyperkid123:
The Travis failure is unrelated. Restarted the build. We can merge after it passes.

PR #1169, opened 2021-06-10
Fix typos

jeromemarc:
Fix 'from from' and add period.


josejulio:
@karelhala could you take a look?

rvsia (approved):

PR #1170, opened 2021-06-11
feat(config): add standalone

redallen:
A way to run cloud.redhat.com apps from `localhost` offline. Ported from https://github.com/redhatinsights/insights-standalone which will be deprecated. See changed packages/config/README.md for usage.

Should be backwards compatible, see comments.


packages/config/src/config.js
```diff
@@ -1,35 +1,36 @@
...
+    port = 1337,
     publicPath,
     appEntry,
     rootFolder,
     https,
     mode,
     appName,
     useFileHash = true,
-    betaEnv = 'ci',
+    betaEnv,
```
redallen, line 15: This isn't enough to determine which build repos to checkout, so I've deprecated it in favor of `env`.

packages/config/src/config.js
```diff
@@ -1,35 +1,36 @@
...
+    env = 'ci-beta',
     sassPrefix,
-    deployment,
```
redallen, line 19: Don't think this is used anymore, cc @Hyperkid123 .

packages/config/src/config.js
```diff
@@ -1,35 +1,36 @@
...
     skipChrome2 = false,
     useProxy,
     localChrome,
     customProxy,
-    proxyVerbose,
     routes,
     routesPath,
-    appUrl,
-    exactUrl,
-    disableFallback,
```
redallen, line 29: I'm happy to test out `useProxy` with a Chrome 1 app if you can link one, but these aren't needed for Chrome 2.

packages/config-utils/package.json
```diff
@@ -17,9 +17,6 @@
     },
     "homepage": "https://github.com/RedHatInsights/frontend-components#readme",
     "peerDependencies": {
-        "jws": "^4.0.0",
-        "glob": "^7.1.7",
-        "nodesi": "^1.15.0",
-        "webpack": ">=5.0.0"
+        "webpack": "^5.0.0"
```
Hyperkid123, line 20: We can't remove these dependencies. Applications that are using the webpack proxy (and there are quite a few of them now) will stop working if they update their config.

Hyperkid123 (changes requested):
I see that you have actually removed/changed a lot of stuff belonging to the current webpack proxy and its configuration. We have to keep them so it won't break applications that are using the current webpack proxy. We want devs to be able to pick proxy or standalone. Not replace the proxy. No config/code changes should be required when updating to a new version of this config. At least not before we decide to release new major version.

If the current proxy settings are not usable for the standalone, create separate config files and enable it based on some flag same as with `useProxy`.

packages/config-utils/package.json
redallen: These are peer dependencies in the config-utils package (aka nothing gets installed). The actual dependencies are in the config package and I've added them there (glob was missing for example).

The peer dependency list should technically be a lot longer, it's just going to be annoying to keep the packages and their versions in sync with the config-utils package. I'll add them like you suggest.

Hyperkid123: You are right. I've thought you have changed "dependencies" not "peerDependencies". Eyes don't work. :facepalm:

redallen:
> I see that you have actually removed/changed a lot of stuff belonging to the current webpack proxy and its configuration.

Yes, I had to for a few reasons.
1. To allow `standalone: true` to work with `useProxy: true`
2. To allow `standalone: true` to work with `localChrome: path`
3. To allow `localChrome: repoUrl`
4. To fix `localChrome` making requests to *.cloud.redhat.com

> We want devs to be able to pick proxy or standalone.

I think they can choose to use `standalone: ...` and `useProxy` which should use some local services while proxying everything else.

> No config/code changes should be required when updating to a new version of this config.
I agree. The reason for `appUrl`, `exactUrl`, and `disableFallback` I believe are for Chrome 1 apps. Are there any apps using these I can test against? I believe I've tidied up the config to a single rule which should work great:
```js
{
  secure: false,
  changeOrigin: true,
  autoRewrite: true,
  context: () => true,
  target,
  router: router(target)
}
```

Hyperkid123:
> I agree. The reason for appUrl, exactUrl, and disableFallback I believe are for Chrome 1 apps. Are there any apps using these I can test against?

You can try the tower analytics frontend repository. They use the new config but have chrome2 rendering disabled for now. https://github.com/RedHatInsights/tower-analytics-frontend

Other repositories that are using the proxy (that I know of) ate for example:
- sources: https://github.com/RedHatInsights/sources-ui
- rbac: https://github.com/RedHatInsights/insights-rbac-ui
- insights dashboard: https://github.com/RedHatInsights/insights-dashboard
All above use chrome 2

> I believe I've tidied up useProxy to a single rule which should work great:

I am not against cleaning it up but we can't break the dev environment within a minor update.

redallen:
1. Tower-analytics is currently broken on ci/ci-beta and qa/qa-beta, not sure how to test. https://ci.cloud.redhat.com/ansible/automation-analytics gives:
image

2. Sources works with new config (without `appUrl`) on ci/ci-beta, qa/qa-beta, and prod/prod-beta:
image

3. Rbac works with new config (without `appUrl`) on ci/ci-beta, qa/qa-beta, and prod/prod-beta:
image

4. Insights-dashboard works with new config (if you remove `customProxy`) on ci/ci-beta, qa/qa-beta, and prod/prod-beta:
image

For insights-dashboard you also have to remove `customProxy` with the old config, so I don't believe this is a regression.

Hyperkid123:
> Rbac works with new config (without appUrl) on ci/ci-beta, qa/qa-beta, and prod/prod-beta:
@redallen have you also tried accessing the `/my-user-access` app? They use the same module.

packages/config-utils/proxy.js
```diff
@@ -1,185 +0,0 @@
-/* eslint-disable no-console */
```
Hyperkid123, line 1: This file canno't be deleted. Some applications (cost) are using only the proxy config from utils package. Can you keep it and only re-export the configuration? I believe that also might be the case for the rest of the deleted files. For example, the `serveLocalFile` is a separate plugin that can be added to the configuration.

redallen:
@Hyperkid123 You're correct `appUrl` is required or else the proxy proxies everything. I've added `proxy.js` under the old path for cost management but I'd like to not add back the unused [cookieTransform](https://github.com/search?q=org%3ARedHatInsights+cookieTransform&type=code) and [serveLocalFile](https://github.com/search?q=org%3ARedHatInsights+serveLocalFile&type=code) unless there's another app in the wild using them. There's a better way to serve local files and cookie transform modifications (if necessary) should be a part of the `config` API.

Hyperkid123:
> I'd like to not add back the unused cookieTransform and serveLocalFile unless there's another app in the wild using them.

@redallen those are supposed to be used as plugins to the config and they are not used internally. There are applications that modify their cookies. and that use the `serveLocalFile` to server the `main.yml` for example. If you feel like there is a better one, feel free to add it but those files have to stay. We can deprecate them with a warning but we can't just delete them.

The thing is that not everybody fully understands webpack and modifying the configuration with methods like these makes it much easier.

Hyperkid123:
@redallen I've just tested it in the sources UI repository. I've tried using the webpack proxy and it does not work without any changes.

I've noticed that the server runs on `https://0.0.0.0:1337/` which is not going to work with remote SSO and APIs. Those only allow the `(ci|qa|prod).foo.cloud.redhat.com` origin.

Also, it was not serving me the correct HTML template I just got this:
```html
<!doctype html><html lang="en-US"><head><meta charset="UTF-8"><title>Home - cloud.redhat.com</title>
```

redallen:
@Hyperkid123 You still have to navigate to `(ci|qa|prod).foo.cloud.redhat.com`. `0.0.0.0` is a fancy localhost.

Hyperkid123:
> @Hyperkid123 You still have to navigate to `(ci|qa|prod).foo.cloud.redhat.com`. `0.0.0.0` is a fancy localhost.

@redallen  I am pretty sure I did that but I got 404. I will double-check though.

Hyperkid123:
@redallen my bad it is working!

packages/config/src/cookieTransform.js
```diff
@@ -1,3 +0,0 @@
-const cookieTransform = require('@redhat-cloud-services/frontend-components-config-utilities/cookieTransform');
```
Hyperkid123, line 1: We are almost good to merge. But re-export also has to stay for now. Can you add a deprecated console warning to import these directly from the `@redhat-cloud-services/frontend-components-config-utilities` package? These re-exports are leftover before the config split happened.

Hyperkid123 (commented):
One last thing, then we can merge :100:

Hyperkid123 (approved):

PR #1172, opened 2021-06-14
[WIP]Append PF vendor to the top of the head.

Hyperkid123:
We have CSS loading order issues in the ansible automation UI. The PF vendor gets loaded after the chrome which causes major graphical issues in chrome. By adding the `pfVendor` to the top of the head we give higher priority to the chrome CSS PF styles. Anything else will be still appended to the end of the head tag as it should be.

### Before
screenshot-cloud redhat com-2021 06 14-16_24_11

### After
screenshot-ci foo redhat com_1337-2021 06 14-16_24_53

### DIFF
--- a/packages/config/src/plugins.js
+++ b/packages/config/src/plugins.js
@@ -29,7 +29,14 @@ module.exports = ({
     new MiniCssExtractPlugin({
         chunkFilename: 'css/[name].[contenthash].css',
         filename: 'css/[name].[contenthash].css',
-        ignoreOrder: true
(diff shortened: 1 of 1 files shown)


Hyperkid123:
I got some updates from the tower team which may explain why this started happening. I would like to first check if this can be fixed from the outside.

Hyperkid123:
Turns out the issue was caused by one of the application libraries. No need for these changes. :man_cartwheeling:

PR #544, opened 2020-06-04
Added ability to reset vote in rule-components feedback

bond95:
@karelhala please, take a look

PR #1175, opened 2021-06-17
RuleFeedback: correct prop name in comment

cben:
Tiny followup to #401.  I don't really know what's going on here, just noticed the disrepancy :smile:  cc @karelhala 

BTW, #544 mentions preferring `feedback` consts rather than -1, 0, 1, maybe you want to update the doc comment accordingly.


karelhala (approved):

karelhala:
> BTW, #544 mentions preferring feedback consts rather than -1, 0, 1, maybe you want to update the doc comment accordingly.

@cben well it's a bit tricky these values `-1`, `0` and `1` comes from the API, the usage of `feedback` const is only if you are using it directly without the API. For instance if API sends different values and you'd want to map them, or when you'd want to manage the values directly.

PR #1176, opened 2021-06-18
Handle missing onRemediationCreated fn

fhlavac:
Handle undefined onRemediationCreated function


Hyperkid123 (approved):

packages/remediations/src/utils/utils.js
```diff
@@ -156,7 +156,7 @@ export const submitRemediation = (formValues, data, basePath, setState) => {
     api.patchRemediation(id, { add, auto_reboot: formValues[AUTO_REBOOT] }, basePath) ||
     api.createRemediation({ name: formValues[SELECT_PLAYBOOK], add, auto_reboot: formValues[AUTO_REBOOT] }, basePath)).then(({ id }) => {
         setState({ id, percent: 100 });
-        data.onRemediationCreated({
+        data.onRemediationCreated && data.onRemediationCreated({
```
karelhala, line 159: Optional chaining mught be better.
```suggestion
        data?.onRemediationCreated({
```

fhlavac: @karelhala is it really applicable in this case with a function?

Hyperkid123: @karelhala that does not work on functions

karelhala: Oh, I forgot the other conditonal one as well
```suggestion
        data?.onRemediationCreated?.({
```

fhlavac: @karelhala oh, great! Didn't know it works like this as well, thank you :-)

PR #1180, opened 2021-06-22
Add console routes to proxy config utils

karelhala:
### New console routes

Since we will be migrating to console.redhat.com let's migrate our paths in proxy as well. We want to have as smooth transition as possible so I've added new config option to use cloud paths. This way apps which start to use the new config can pass this option untill console paths are available.

### DIFF
--- a/packages/config-utils/proxy.js
+++ b/packages/config-utils/proxy.js
@@ -25,15 +25,16 @@ module.exports = ({
     localChrome,
     appUrl = [],
     publicPath,
-    proxyVerbose
+    proxyVerbose,
+    useCloud = false
 }) => {
     const proxy = [];
     const registry = [];
     const majorEnv = env.split('-')[0];
     const minorEnv = majorEnv === 'prod' ? '' : `${majorEnv}.`;
     const target = env === 'prod-stable'
-        ? 'https://cloud.redhat.com/'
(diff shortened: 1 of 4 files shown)


rvsia (approved):

packages/config-utils/proxy.js
```diff
@@ -25,15 +25,16 @@ module.exports = ({
     localChrome,
     appUrl = [],
     publicPath,
-    proxyVerbose
+    proxyVerbose,
+    useCloud = false
```
redallen, line 29: In order to make adopting the webpack proxy easier and smooth out the transition can we please set this to `true`? Currently it's **required** to pass `useCloud = true` since console.redhat.com is not working.

karelhala: Well the transition to console.redhat.com should happen in a few days, we'd have to open another PR to flip this value, apps would then have to update their config version once again. This is safer way, if you are going to use newer version add `useCloud: true` to config, once the console is deployed you can just remove that from the config. This approach means less steps for applications (they'd have to update the package in order for them to work without any problem).

redallen: It currently means more steps for applications I have to deal with, but I'll trust you.

PR #1181, opened 2021-06-22
Add useChrome hook to return chrome global context.

Hyperkid123:
part of: https://issues.redhat.com/browse/RHCLOUD-14898

Enables sharing a global context via `useChrome` hook. Eventually will replace the current chrome API available on window.

### DIFF
--- a/packages/components/src/useChrome/useChrome.js
+++ b/packages/components/src/useChrome/useChrome.js
@@ -1,0 +1,17 @@
+import { useScalprum } from '@scalprum/react-core';
+
+const useChrome = (selector) => {
+    const state = useScalprum();
+    let chrome = state.api?.chrome || {};
+    chrome = {
+        ...chrome,
+        initialized: state.initialized
+    };
+    if (typeof selector === 'function') {
+        return selector(chrome);
+    }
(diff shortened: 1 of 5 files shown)


packages/components/src/useChrome/useChrome.js
```diff
@@ -0,0 +1,17 @@
+import { useScalprum } from '@scalprum/react-core';
+
+const useChrome = (selector) => {
+    const state = useScalprum();
+    let chrome = state?.api?.chrome || {};
+    chrome = {
+        ...chrome,
+        initialized: state.initialized
```
rvsia, line 8: what if `state` is undefined?

Hyperkid123: It can't be, by default it's always an empty object. It could only be undefined if you were using the hook outside of the context.

rvsia: see line number 5 then 👀

Hyperkid123: :smiling_imp:

rvsia (approved):

PR #1189, opened 2021-06-22
Add empty state for Insights not connected

leSamo:
Ticket: https://issues.redhat.com/browse/ESSNTL-12
Mockup: https://marvelapp.com/prototype/geh911a/screen/79982935

Add component to make `Not connected` empty state consistent. Default text can be overwritten with props to enable intl.

notConnected

### DIFF
--- a/packages/components/src/NotConnected/NotConnected.js
+++ b/packages/components/src/NotConnected/NotConnected.js
@@ -1,0 +1,40 @@
+import React from 'react';
+import propTypes from 'prop-types';
+import { EmptyState, EmptyStateIcon, EmptyStateBody, Title, Button } from '@patternfly/react-core';
+import { DisconnectedIcon } from '@patternfly/react-icons';
+
+const NotConnected = ({ titleText, bodyText, buttonText }) => (
+    <EmptyState>
+        <EmptyStateIcon icon={DisconnectedIcon} />
+        <Title headingLevel="h5" size="lg">
+            {titleText}
+        </Title>
+        <EmptyStateBody>
+            {bodyText}
+        </EmptyStateBody>
+        <Button
+            variant="primary"
+            component="a"
+            href="http://access.redhat.com/products/cloud_management_services_for_rhel#getstarted"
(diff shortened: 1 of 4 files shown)


packages/components/src/NotConnected/NotConnected.js
```diff
@@ -0,0 +1,40 @@
...
+                {buttonText}
+            </Button>
+        </EmptyState>
+    );
+};
+
+NotConnected.propTypes = {
+    titleText: propTypes.string,
+    bodyText: propTypes.string,
+    buttonText: propTypes.string
```
rvsia, line 31: I think we should change this to `node` to enable elements too (for example, translatable texts)

packages/components/src/NotConnected/NotConnected.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import propTypes from 'prop-types';
+import { EmptyState, EmptyStateIcon, EmptyStateBody, Title, Button } from '@patternfly/react-core';
+import { DisconnectedIcon } from '@patternfly/react-icons';
+
+const NotConnected = ({ titleText, bodyText, buttonText }) => {
+    return (
```
rvsia, line 7: no need for a return block

```jsx
=> (<EmptyState 
...
```

packages/components/src/NotConnected/NotConnected.js
```diff
@@ -0,0 +1,40 @@
...
+            <Title headingLevel="h5" size="lg">
+                {titleText}
+            </Title>
+            <EmptyStateBody>
+                {bodyText}
+            </EmptyStateBody>
+            <Button
+                variant="primary"
+                component="a"
+                href="http://access.redhat.com/products/cloud_management_services_for_rhel#getstarted"
```
rvsia, line 19: We should also open this link a new tab:

```jsx
target="_blank"
rel="noopener noreferrer"
```

rvsia (approved):

PR #1190, opened 2021-06-23
fix(config): fix devServer port

Hyperkid123:
We need port 8002 for insights proxy to work. 1337 is only for webpack/standalone settings.


rvsia (approved):

packages/config/src/config.js
```diff
@@ -34,6 +34,7 @@ module.exports = ({
         console.warn('betaEnv is deprecated in favor of env');
     }
 
+    const devServerPort = typeof port === 'number' ? port : useProxy || standalone ? 1337 : 8002;
```
rvsia, line 37: Would it be possible to wrap `: (useProxy || standalone) ? ...` so the priority is clear?

PR #1196, opened 2021-06-24
Fix/standalone proxy

redallen:
Koku verified working with the following patch:
```
+    devServer: {
+      stats,
+      contentBase: false,
+      historyApiFallback: {
+        index: `${publicPath}index.html`,
+      },
+      host: '0.0.0.0',
+      hot: false, // default is true, which currently does not work with Insights and federated modules?
+      port,
+      disableHostCheck: true,
+      headers: {
+        'Access-Control-Allow-Origin': '*',
+        'Access-Control-Allow-Headers': 'Origin, X-Requested-With, Content-Type, Accept, Authorization',
+      },
+      https: useProxy,
+      ...proxy({
+        port,
+        env: 'qa-stable',
+        useProxy,
+        appUrl: [`/${isBeta ? 'beta/' : ''}openshift/cost-management`],
+        proxyVerbose: true,
+        publicPath,
+        useCloud: true
+      })
```

The fix was to override the `end` function in addition to the `write` function. I think webpack-dev-server version differences are to blame.


dlabrecq (commented):
Martin recently updated Cost Management via https://github.com/project-koku/koku-ui/pull/2001. Any chance we can continue with our existing configuration?

redallen:
I really think it's best to use this new config which extends `devServer` for a few reasons:

1. It allows adding more than just adding `proxy` entries which allows extending webpack's `express` dev server.
2. It also removes the need for the `exactUrl` and `disableFallback` options
3. It resolves ESI tags without having to make web requests for each page which is slow and have led to me hitting the cloud.redhat.com ratelimit
4. It standardizes a more familiar `env="ci-beta"` vs `BETA=true ... { betaEnv="ci" }`

dlabrecq:
We're using the current config because it resolves a few issues for us; a Google Chrome debugging issue, refreshes the page upon an update, and we don't need to run Docker. We weren't able to get that patch to work in the same way, and had to run with Docker, so hesitant to move away from what is currently working well for us.

redallen:
> a Google Chrome debugging issue
Not sure what you mean.
> refreshes the page upon an update
The patch does this for me.
> we don't need to run Docker
You shouldn't need to run Docker. If Docker is running something is wrong.

Let's schedule some time for you to test out these changes.

ddonahue007 (approved):
LGTM

dlabrecq (approved):

Hyperkid123 (commented):
@redallen I have noticed one regression.

When using the `localChrome` config the local version of chrome is only served if you access the correct URL. For example, if you run webpack proxy in RBAC and use the local chrome setting, the local chrome will be only served under RBAC routes. If you make a hard refresh on let's say `/settings/sources` you receive assets from CI instead.

Before when using local chrome, you would get the local chrome assets no matter which route you access. Can we fix that?

Hyperkid123:
@rvsia  has also found that the `exactUrl` is missing: https://github.com/RedHatInsights/frontend-components/pull/1204
That is required for the landing page.

redallen:
I've fixed the `localChrome` issue by stripping the SHA and using `glob` to find a similar local file. I've verified this works with both `npm run build` and `npm run watch` in https://github.com/RedHatInsights/insights-chrome.

I've fixed the `exactUrl` by allowing regexes in `appUrl`. The following config works with https://github.com/RedHatInsights/landing-page-frontend:
```js
  useProxy: true,
  env: process.env.BETA ? 'ci-beta' : 'ci-stable',
  appUrl: process.env.BETA ? /^\/beta$/ : /^\/$/,
  useCloud: true,
  proxyVerbose: true
```

Hyperkid123 (approved):

PR #1213, opened 2021-07-20
fix(proxy): fix building stage url

redallen:
Only care about `majorEnv === 'prod'` if `useCloud: false` in order to map to https://console.redhat.com or https://console.stage.redhat.com .

Allow custom `target` to override for urls like https://cloud.stage.redhat.com/

### DIFF
--- a/packages/config-utils/proxy.js
+++ b/packages/config-utils/proxy.js
@@ -27,6 +27,7 @@ module.exports = ({
     publicPath,
     proxyVerbose,
     useCloud = false,
+    useStage = false,
     target = ''
 }) => {
     const proxy = [];
@@ -34,17 +35,19 @@ module.exports = ({
     const majorEnv = env.split('-')[0];
     if (target === '') {
         target += 'https://';
-        if (useCloud && majorEnv !== 'prod') {
+        if (!useStage && majorEnv !== 'prod') {
             target += majorEnv + '.';
         }
 
         target += useCloud ? 'cloud' : 'console';
-        if (!useCloud && majorEnv !== 'prod') {
+        if (useStage && majorEnv !== 'prod') {
             target += '.stage';
         }
 
         target += '.redhat.com/';
     }
+
+    console.log('Proxing to', target);
 
     if (!Array.isArray(appUrl)) {
         appUrl = [ appUrl ];
--- a/packages/config-utils/proxy.js
+++ b/packages/config-utils/proxy.js
@@ -26,15 +26,26 @@ module.exports = ({
     appUrl = [],
     publicPath,
     proxyVerbose,
-    useCloud = false
+    useCloud = false,
+    target = ''
 }) => {
     const proxy = [];
     const registry = [];
     const majorEnv = env.split('-')[0];
-    const minorEnv = majorEnv === 'prod' ? '' : `${majorEnv}.`;
-    const target = env === 'prod-stable'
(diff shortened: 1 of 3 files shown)


dlabrecq (approved):

rvsia (commented):
Left a comment, maybe I am looking wrongly 🙈

packages/config-utils/proxy.js
```diff
@@ -26,15 +26,26 @@ module.exports = ({
...
-        ? `https://${useCloud ? 'cloud' : 'console'}.redhat.com/`
-        : `https://${minorEnv}${useCloud ? 'cloud' : 'console'}.redhat.com/`;
+    if (target === '') {
+        target += 'https://';
+        if (useCloud && majorEnv !== 'prod') {
+            target += majorEnv + '.';
+        }
+
+        target += useCloud ? 'cloud' : 'console';
+        if (!useCloud && majorEnv !== 'prod') {
```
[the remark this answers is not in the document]
redallen: Yes, that's correct since https://ci.console.redhat.com doesn't currently exist.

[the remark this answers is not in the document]
Hyperkid123: It does not exist not but It will exist.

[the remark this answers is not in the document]
redallen: Okay, I'm not sure _when_ it will exist but I've updated the config to point to it unless `useStage: true` is passed.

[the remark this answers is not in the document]
Hyperkid123: It should exist after July 29th. We are also quite nervous about the date.

packages/config-utils/proxy.js
```diff
@@ -26,15 +26,29 @@ module.exports = ({
     appUrl = [],
     publicPath,
     proxyVerbose,
-    useCloud = false
+    useCloud = false,
+    useStage = false,
+    target = ''
```
rvsia, line 31: I guess we should add these to the documentation (readme).

And config does not pass all attributes to proxy automatically so add it there too https://github.com/RedHatInsights/frontend-components/blob/8ffea0df0c366922c180a4f3a210d3e472a91e56/packages/config/src/config.js#L180

and shouldn't be stage indicated just by `env=` `stage-stable` or `stage-beta`? 🤔

redallen: I'll add to the docs and make sure the props are passed.

The `env` is used to determine which build repo branches to checkout. `stage-stable` and `stage-beta` don't exist, but I suppose I could use them and then replace `stage` with `qa`.

rvsia: That would be awesome! 🍹

rvsia: But I think stage branches can exist, if repos configure them.. @Hyperkid123 ? 🤔 

(Not for Chrome though)

Hyperkid123: It does not really matter from which branch you run the proxy. And chrome does not have a stage branch in the build repo. Stage env usually use QA builds. For chrome and every other application.

redallen:
Addressed feedback, CI passing, please rereview!

packages/config-utils/proxy.js
rvsia: Yeah, but I think if application create an stage branch, it would be served from that, wouldn't it?

rvsia (approved):

PR #1225, opened 2021-08-03
fix(standalone): ignore esi query params

redallen:
Fix ESI injection for URLs with `.` in their query parameters like http://localhost:1337/beta/ansible/insights/clusters?clusters.quick_date_range=last_62_days

### DIFF
--- a/packages/config-utils/standalone/services/default/chrome.js
+++ b/packages/config-utils/standalone/services/default/chrome.js
@@ -39,7 +39,7 @@ module.exports.registerChrome = ({ app, chromePath, keycloakUri, https, proxyVer
     // Unfortunately we have to alter the contents of ALL text/html responses like akamai
     // This means modifying the response object directly.
     app.use((req, res, next) => {
-        const ext = path.extname(req.url);
+        const ext = path.extname(req.url.replace(/\?.*/, ''));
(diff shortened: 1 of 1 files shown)


packages/config-utils/standalone/services/default/chrome.js
```diff
@@ -39,7 +39,7 @@ module.exports.registerChrome = ({ app, chromePath, keycloakUri, https, proxyVer
     // Unfortunately we have to alter the contents of ALL text/html responses like akamai
     // This means modifying the response object directly.
     app.use((req, res, next) => {
-        const ext = path.extname(req.url);
+        const ext = path.extname(req.url.replace(/\?.*/, ''));
```
rvsia, line 42: Maybe I would extract this to a separate function and call it something like "removeUrlQuery" so everybody know what it does without solving the regex. But this is just a minor nitpick. 🌵

rvsia (approved):
🏆

packages/config-utils/standalone/services/default/chrome.js
rvsia: Or you can just extract the regex and name it.

redallen: If I end up needing it again I'll extract it!

PR #1228, opened 2021-08-05
Change path in esm scss exports

rvsia:
This changes the wrong export `'./index.scss'` to correct `export { default as style } from "../../RuleTable/./index.css";`

But I am not sure if this is what we want as it still throws a warning:

```
WARNING in ../rule-components/esm/RuleTable/RuleTable.js 40:0-63
export 'default' (reexported as 'style') was not found in '../../RuleTable/./index.css' (module has no exports)
```


rvsia:
Closing in favor of https://github.com/RedHatInsights/frontend-components/pull/1246

PR #1230, opened 2021-08-05
fix(fedmod): properly share scalprum

redallen:
Webpack requires that you pass a `requiredVersion` when using this syntax.

Fix for @jschuler.


jschuler:
Thanks @redallen ! Tested this change in my workspace and it worked

jschuler (approved):

Hyperkid123 (approved):

PR #1239, opened 2021-08-12
Wrap async comp

karelhala:
### Missing styles on fed modules

When importing fed module the app wouldn't apply module's styles because of scoping. This PR fixes such issue, by adding container as a wrapper around async component and applying classname same as is the name of the application.

There are potentional bugs with visuals where the content might not span over the entire content. Should this error arise we can investigate if the app is not doing anything strange with the styling or we can apply the class to first rendered element.

### DIFF
--- a/packages/components/src/AsyncComponent/index.js
+++ b/packages/components/src/AsyncComponent/index.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { ScalprumComponent } from '@scalprum/react-core';
 import { Bullseye, Spinner } from '@patternfly/react-core';
+import classNames from 'classnames';
 
 const BaseAsyncComponent = ({
     appName,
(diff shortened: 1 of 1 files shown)


packages/components/src/AsyncComponent/index.js
```diff
@@ -9,18 +9,22 @@ const BaseAsyncComponent = ({
...
-        <ScalprumComponent
-            appName={appName}
-            module={module}
-            scope={scope || appName}
-            ErrorComponent={fallback}
-            ref={innerRef}
-            fallback={fallback}
-            {...props}
-        />
+        <section className={`${className || ''} ${appName}`}>
```
Hyperkid123, line 16: What if we add a `component` prop (like `a`, `div`, `span`, etc.)  to allow the consumer to customize the element wrapper? And default to section. It would be useful if you are sharing just a component, not a whole app (like a card).

Also probably use `classnames` for the class. It's installed anyway.

Hyperkid123 (approved):

PR #1241, opened 2021-08-17
feat(RemediationsButton): Allow passing props for button

bastilian:
This will enable apps to provide additional props for the button like an OUIA id[0]

[0] https://github.com/RedHatInsights/compliance-frontend/pull/1330

### DIFF
--- a/packages/remediations/src/RemediationButton/RemediationButton.js
+++ b/packages/remediations/src/RemediationButton/RemediationButton.js
@@ -42,14 +42,16 @@ class RemediationButton extends React.Component {
     }
 
     render() {
+        const { children, buttonProps } = this.props;
+
         if (this.state.remediations && !this.state.hasPermission) {
             return (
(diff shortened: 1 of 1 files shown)


packages/remediations/src/RemediationButton/RemediationButton.js
```diff
@@ -42,14 +42,16 @@ class RemediationButton extends React.Component {
     }
 
     render() {
+        const { children, ...buttonProps } = this.props;
```
rvsia, line 45: This can selects props that do not belong to the button such as 

```jsx
    dataProvider: propTypes.func.isRequired,
    onRemediationCreated: propTypes.func,
```

Maybe add a new prop 'buttonProps: PropTypes.object' and use it.

rvsia (approved):

bastilian:
@rvsia Thank you for reviewing and merging!

PR #1242, opened 2021-08-17
Add 'mobile' or 'desktop' to conditional filter ID

fhlavac:
Added '-mobile' or '-desktop' to conditional filter ID for QA

@john-dupuy

### DIFF
--- a/packages/components/src/ConditionalFilter/ConditionalFilter.js
+++ b/packages/components/src/ConditionalFilter/ConditionalFilter.js
@@ -64,11 +64,11 @@ class ConditionalFilter extends Component {
                                         {
                                             ...activeItem.type !== conditionalFilterType.custom &&
                         {
-                            placeholder: placeholder || activeItem.placeholder || `Filter by ${activeItem.label}`,
-                            id: (activeItem.filterValues && activeItem.filterValues.id) || currentValue
+                            placeholder: placeholder || activeItem.placeholder || `Filter by ${activeItem.label}`
                         }
                                         }
                                         { ...activeItem.filterValues }
+                                        id={`${(activeItem.filterValues && activeItem.filterValues.id) || currentValue}-mobile`}
                                     />
 
                                 </ToolbarItem>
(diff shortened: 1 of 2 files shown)


packages/components/src/ConditionalFilter/ConditionalFilter.js
```diff
@@ -64,11 +64,11 @@ class ConditionalFilter extends Component {
                                         {
                                             ...activeItem.type !== conditionalFilterType.custom &&
                         {
-                            placeholder: placeholder || activeItem.placeholder || `Filter by ${activeItem.label}`,
-                            id: (activeItem.filterValues && activeItem.filterValues.id) || currentValue
+                            placeholder: placeholder || activeItem.placeholder || `Filter by ${activeItem.label}`
                         }
                                         }
                                         { ...activeItem.filterValues }
+                                        id={`${(activeItem.filterValues && activeItem.filterValues.id) || currentValue}-mobile`}
```
john-dupuy, line 71: Is it possible to add an `ouiaId` to these too?

rvsia (commented):
I am not sure if we want to change ID when the component is mobile or not. I guess the ID should always stay the same. Isn't ouiaId enough?

fhlavac:
@john-dupuy is the ouiaId enough?

rvsia:
@fhlavac Oh, there are two inputs at that same time? 👀  Then it probably makes sense to have different IDs. 

It would be nice to show only one of the inputs in DOM, I have a hook for getting PF screen variant, so maybe we could integrate it here: https://github.com/RedHatInsights/sources-ui/blob/master/src/hooks/useScreen.js

john-dupuy:
> It would be nice to show only one of the inputs in DOM

I definitely agree here :+1:

Hyperkid123:
@fhlavac OK We can make it show only one. I wanted to avoid having additional listeners but I guess it is optimal to have just one. @fhlavac I will close this PR and create a new one with only one input in the DOM.

PR #1256, opened 2021-08-31
Add useScreenSize hook

rvsia:
- [useScreenSize](#usescreensize)
  - [Usage](#usage)
  - [Breakpoints](#breakpoints)
  - [Variants](#variants)
  - [isSmallScreen](#issmallscreen)

# useScreenSize

`useScreenSize` is a React hook that listens to windows resizing and returns a PatternFly variant of the current size. This hook is helping when different React structures has to be shown accordingly to the window size. For example, when you want to show a dropdown instead of a button set.

## Usage

```jsx
import useScreenSize from '@redhat-cloud-services/frontend-components/useScreenSize';

const DummyComponent = () => {
    const currentVariant = useScreenSize();

    return <div>Current PF4 variant is: {currentVariant}</div>;
};
```

## Breakpoints

`breakpoints` object is a map of `{ [variant]: size (as number) }`.

## Variants

Current variants can be found as keys of `breakpoints` object.

```jsx
['xs', 'sm', 'md', 'lg', '2xl']
```

## isSmallScreen

`isSmallScreen` is a simple helper that returns `true` if the current variant is a small screen.

```jsx
import { useScreenSize, isSmallScreen } from '@redhat-cloud-services/frontend-components/useScreenSize';

const DummyComponent = () => {
    const currentVariant = useScreenSize();

    if(isSmallScreen(currentVariant)) {
        return <div>Mobile content</div>;
    }

    return <div>Desktop content</div>;
};
```

### DIFF
--- a/packages/components/src/useScreenSize/breakpoints.js
+++ b/packages/components/src/useScreenSize/breakpoints.js
@@ -1,0 +1,17 @@
+import xsBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_xs';
+import smBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_sm';
+import mdBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_md';
+import lgBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_lg';
+import xlBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_xl';
+import xxlBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_2xl';
+
+const breakPoints = {
+    xs: parseInt(xsBreakpoint.value.replace('px', '')),
+    sm: parseInt(smBreakpoint.value.replace('px', '')),
+    md: parseInt(mdBreakpoint.value.replace('px', '')),
+    lg: parseInt(lgBreakpoint.value.replace('px', '')),
+    xl: parseInt(xlBreakpoint.value.replace('px', '')),
+    '2xl': parseInt(xxlBreakpoint.value.replace('px', ''))
+};
+
+export default breakPoints;
--- a/packages/components/src/useScreenSize/breakpoints.js
+++ b/packages/components/src/useScreenSize/breakpoints.js
@@ -1,9 +1,9 @@
-import xsBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_xs';
-import smBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_sm';
-import mdBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_md';
-import lgBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_lg';
-import xlBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_xl';
-import xxlBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_2xl';
+import xsBreakpoint from '@patternfly/react-tokens/dist/js/global_breakpoint_xs';
+import smBreakpoint from '@patternfly/react-tokens/dist/js/global_breakpoint_sm';
+import mdBreakpoint from '@patternfly/react-tokens/dist/js/global_breakpoint_md';
+import lgBreakpoint from '@patternfly/react-tokens/dist/js/global_breakpoint_lg';
+import xlBreakpoint from '@patternfly/react-tokens/dist/js/global_breakpoint_xl';
+import xxlBreakpoint from '@patternfly/react-tokens/dist/js/global_breakpoint_2xl';
 
 const breakPoints = {
     xs: parseInt(xsBreakpoint.value.replace('px', '')),
--- a/packages/components/src/useScreenSize/getVariant.js
+++ b/packages/components/src/useScreenSize/getVariant.js
@@ -5,13 +5,13 @@ const getVariant = () => {
 
     let result;
 
-    Object.entries(breakpoints).some(([ variant, size ]) => {
-        if (width > size) {
-            result = variant;
-        } else {
-            return true;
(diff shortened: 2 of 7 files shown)


packages/components/src/useScreenSize/breakpoints.js
```diff
@@ -0,0 +1,17 @@
+import xsBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_xs';
+import smBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_sm';
+import mdBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_md';
+import lgBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_lg';
+import xlBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_xl';
+import xxlBreakpoint from '@patternfly/react-tokens/dist/esm/global_breakpoint_2xl';
```
Hyperkid123, line 6: You want to import from the js version of the build. Otherwise, you might break nodejs env (tests) which will require transpiling the modules.

packages/components/src/useScreenSize/getVariant.js
```diff
@@ -0,0 +1,19 @@
...
+
+    Object.entries(breakpoints).some(([ variant, size ]) => {
+        if (width > size) {
+            result = variant;
+        } else {
+            return true;
+        }
+    });
+
+    return result;
```
karelhala, line 16: Why not use `find` instead of assign and somewhat complicated `some`?
```JS
return Object.entries(breakpoints).find(([ variant, size ]) => width > size)?.[0];
```

packages/components/src/useScreenSize/isSmallScreen.js
```diff
@@ -0,0 +1,7 @@
+import breakpoints from './breakpoints';
+
+const variants = Object.keys(breakpoints);
+
+const isSmallScreen = screenSize => variants.indexOf(screenSize) <= variants.indexOf('sm');
```
karelhala, line 5: Why two `indexOf`? It can be done just by simple object access
```JS
const isSmallScreen = screenSize => (breakpoints?.[screenSize] <= breakpoints.sm);
```

packages/components/src/useScreenSize/useScreenSize.js
```diff
@@ -0,0 +1,26 @@
...
+
+const useScreen = () => {
+    const [ variant, setVariant ] = useState(() => getVariant());
+    const prev = useRef(variant);
+
+    useEffect(() => {
+        function handleResize() {
+            const newVariant = getVariant();
+            if (newVariant !== prev.current) {
+                prev.current = newVariant;
```
karelhala, line 13: You have to store function reference here, instead of the actual value, otherwise next cycle will not calculate with new screen size.
```suggestion
            const newVariant = getVariant;
            if (newVariant() !== prev.current()) {
                prev.current = newVariant;
```

packages/components/src/useScreenSize/getVariant.js
rvsia: That would find just the first variant that is bigger than the current size, we need to find the last one. We could reverse the  array first but I think that would make it somewhat complicated. This `some` will do the minimal required numbers of cycles without a need to reverse anything.

packages/components/src/useScreenSize/useScreenSize.js
rvsia: Why to store the function reference?

It's just a string and it compares two strings.

If you store a function in the ref, you would get always true because both of the functions are the same.

karelhala: Then the state initialization is incorrect, as you are storing the callback to it.

rvsia: Nope, passing a function to useState hooks is a init function - a function, that runs only once on the initial render and returns the initial state.

If you want to store a function in state, you have to do `useState(() => () => blabla())`

packages/components/src/useScreenSize/getVariant.js
rvsia: Using `some` this way is basically a while cycle you can end when you need it.

packages/components/src/useScreenSize/useScreenSize.js
Hyperkid123: @karelhala it's not. You can use a function to initialize `useState`. It will use the return value.
https://reactjs.org/docs/hooks-reference.html#lazy-initial-state

packages/components/src/useScreenSize/getVariant.js
karelhala: I had to run it locally to see what it actually does tbh. It's a clever way of how to use the `some` function, but I think it's a bit complicated to read. Because what this function does is finding the last breakpoint which is smaller than the current screen size. 
```JS
const breakpoints = {
  'sm': 100,
  'md': 200,
  'xl': 300,
  'xxl': 400
}

const getVariant = (width) => {
  let result;
​
    Object.entries(breakpoints).some(([ variant, size ]) => {
        if (width > size) {
            result = variant;
        } else {
            return true;
        }
    });
​
    return result;
}

getVariant(150); // returns 'sm'
getVariant(350); // returns 'xl'
```

I know we shouldn't use `break`, but this might be a bit more straight forward
```JS
const getVariant2 = (width) => {
  let result;
  for (const [variant, size] of Object.entries(breakpoints)) {
    if (size > width) {
      break;
    }
  result = variant;
}

getVariant2(150); // returns 'sm'
getVariant2(350); // returns 'xl'
```

packages/components/src/useScreenSize/useScreenSize.js
karelhala: Good to know!

packages/components/src/useScreenSize/getVariant.js
rvsia: small correction: `if (size >= width) {`

Yeah, I wanted to use `break` initially but using some was much cooler. 😎

Hyperkid123 (approved):

PR #1258, opened 2021-09-01
Use PDF generator v2 and refactor charts to use SVGs

karelhala:
### React-pdf v2
Since react-pdf is in version 2 we can start using it. This will help us by unblocking us on strict version of react-dom, since we can now use `usePDF` hook to refresh PDF when something renderes. When rendering charts each chart will create promise, once it is rendered we fullfill this promise.

### Additional charts support
Since we are now passing the paths, lines and texts directly to SVG elements, we can now support additional charts. This PR enables such feature, we should identify which charts are important to us and start using them.


Hyperkid123:
@karelhala did you try it on live data? In advisor for example?

packages/demo/src/index.html
```diff
@@ -7,6 +7,12 @@
     <meta name="theme-color" content="#000000">
     <link rel="stylesheet" type="text/css" href="https://unpkg.com/@patternfly/patternfly@latest/patternfly-base.css"/>
     <link rel="stylesheet" type="text/css" href="https://unpkg.com/@patternfly/patternfly@latest/patternfly-addons.css"/>
+    <style>
```
Hyperkid123, line 10: Can you remove the changes from the demo?

bastilian:
I tried testing this via npm link on this PR[0], but it doesn't seem to properly resolve the dependencies there[1].
Is there something else that needs to be done?

[0] https://github.com/RedHatInsights/compliance-frontend/pull/1351/files#diff-867474428a143645202e8d66d572075e10890e386ceccff95ad9a77d451c3afaR11
[1] https://gist.github.com/bastilian/77aad8ba3744c9eeac3ad26d9e28f4ac

bastilian:
@karelhala Should this already work actually? I tried it again with the same PR mentioned above only importing the actually `DownloadButton` instead. With that the modal that is supposed to include the button isn't opening anymore and crashing my browser. :/

karelhala:
@bastilian it works in demo mode, but there is a problem with async calls and renders. We are pushing the UI containerization, once that is done I'll come back to this and fix the async hell.

bastilian:
@karelhala Thanks for taking care of this! Just curious, is it an issue with async within the `DownloadButton`? I was messing around with `react-pdf/renderer` directly and trying to implement something to render a PDF to develop/test, but even directly using it in any variant, doesn't load it properly and the browser stalls.

karelhala:
@bastilian  This was previously a lot harder because we had to use canvas, but in pdf-renderer@v2 they introduced SVG elements, but still they are not regular SVG elements so we can't render directly PF charts in there. We have to render the chart outside the PDF, get the paths, remove the original PF chart and pass these paths to pdf-renderer svg elements. And the root cause of the async hell problem is that newer version of react-dom is asynchronous (which is awesome, but sucks for us). So we have to wait for the chart to render and once that is done refresh the PDF renderer and download/show the PDF. Preview is really simple as the preview can refresh multiple times when each chart renders. Download button is the one that is deep in async hell, as we have to wait for all charts, we don't know how many charts are actually there and download the file once all charts render.

akofink:
@karelhala do you have a jira issue ID for tracking this effort? Thanks!

karelhala:
@akofink yeah! There is https://issues.redhat.com/browse/RHCLOUD-16221 to track work on local development.

Hyperkid123:
@karelhala I will close this PR as we are moving forwards with the backed service.

PR #1261, opened 2021-09-03
Implement expand on hover for Truncate

bastilian:
We are removing `react-truncate` from compliance in favour of `Truncate`[0] 
This PR would allow us to also replace cases where we toggle the text on hover and previously (partly) used `react-truncate`, but had the hover implemented ourselves.

Here an example of what it looks like:

https://user-images.githubusercontent.com/7757/132039282-f6da281d-31d4-4e87-8f83-3758a1b34231.mp4

It also refactors the Truncate component to a function component.

[0] https://github.com/RedHatInsights/compliance-frontend/pull/1352

### DIFF
--- a/packages/components/src/Truncate/Truncate.js
+++ b/packages/components/src/Truncate/Truncate.js
@@ -10,7 +10,15 @@ const dangerousHtml = (html) =>
     ({ __html: sanitizeHtml(html) });
 
 const Truncate = ({
-    text = '', length = 150, expandText = 'Read more', collapseText = 'Collapse',  className, inline, spaceBetween
+    text = '',
+    length = 150,
+    expandText = 'Read more',
+    hideExpandText = false,
+    expandOnMouseOver = false,
+    collapseText = 'Collapse',
+    className,
+    inline,
+    spaceBetween
 }) => {
     const truncateClasses = classNames(
         'ins-c-truncate',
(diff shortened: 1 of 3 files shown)


packages/components/src/Truncate/Truncate.js
```diff
@@ -1,111 +1,91 @@
...
+        <Button
+            className='ins-c-collapse-button'
+            variant='link'
+            onClick={ toggleText }>
+            { collapseText }
+        </Button>;
+    const textWithOverflow = showText === false ? `${trimmedText}${textOverflow ? '...' : '' }` : text;
+    const html = dangerousHtml(textWithOverflow);
+    const mouseOverHandler = expandOnMouseOver && {
+        onMouseEnter: () => setShowText(true),
```
bastilian, line 54: Using `toggleText` for these worked in the browser, but it did not yield proper test results. 🤷 not sure why.

packages/components/src/Truncate/Truncate.test.js
```diff
@@ -61,6 +61,20 @@ describe('Truncate component', () => {
                     const wrapper = shallow(<Truncate text={text} inline={isInline} length={1000} />);
                     expect(toJson(wrapper)).toMatchSnapshot();
                 });
+
+                it('hovering over toggles to collapse', () => {
```
bastilian, line 65: This test could be split up in two tests to separate the testing of mouseenter and mouseleave.

rvsia (approved):

PR #1267, opened 2021-09-20
Router as singleton package

Hyperkid123:
cc @karelhala 

We should try this in CI once chrome is deployed. I will send coms once the singleton changes hit the pre-prod.


karelhala (approved):

Hyperkid123:
Replaced by v6

PR #1271, opened 2021-09-22
chore: Bump webpack-dev-server

bastilian:
As of recently `npm start` started failing for me. After investigating it turned out that `fsevents` which is used by `chokidar` to watch files was resolved to an older version of the package and cause `UnhandledPromiseRejectionWarning: TypeError: fsevents.watch is not a function`. (`fsevents` is a low level wrapper around macOS' `FSEvents` API, therefore the issue should only be in that platform.) 

Bumping webpack-dev-server resolves the issue.

### DIFF
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -43,7 +43,7 @@
         "util": "^0.12.4",
         "webpack": "^5.38.1",
         "webpack-cli": "^4.7.2",
-        "webpack-dev-server": "^3.11.2",
+        "webpack-dev-server": "^4.2.1",
         "write-file-webpack-plugin": "^4.5.1"
     }
 }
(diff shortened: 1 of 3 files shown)


packages/config/package.json
```diff
@@ -43,7 +43,7 @@
         "util": "^0.12.4",
         "webpack": "^5.38.1",
         "webpack-cli": "^4.7.2",
-        "webpack-dev-server": "^3.11.2",
+        "webpack-dev-server": "^4.2.1",
```
rvsia, line 46: There is lot of changes, so it's not that simple: https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md 😭 

cc @redallen

redallen:
It's good to upgrade, but you'll have to migrate all the webpack config in the `config` and `config-utils` packages.

bastilian:
Closing as @Hyperkid123 is going to take proper care of it.

PR #1273, opened 2021-09-24
added user to Keycloak for cost management

ddonahue007:
added user to Keycloak for cost management


redallen (approved):

Hyperkid123:
@ddonahue007 can you fix up the conflict, please?

rvsia (approved):

ddonahue007:
@rvsia I see a diff but no conflicts/merging needed to be done in my local repo. I do not have write access here so I can not use github to resolve the issue.

rvsia:
@ddonahue007 I think you should just remove `package.json` and  `package-lock.json` changes from this PR. I guess your NPM install just reordered the dependency list and updated some of the packages and none of these changes are required here.

PR #1275, opened 2021-09-27
fix(proxy): don't use agent for routes

redallen:
Using `agent` and `onProxyReq` crashes with this:
```sh
_http_outgoing.js:561
    throw new ERR_HTTP_HEADERS_SENT('set');
    ^
Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ClientRequest.setHeader (_http_outgoing.js:561:11)
    at ProxyServer.cookieTransform (/home/akofink/dev/repos/compliance-frontend/node_modules/@redhat-cloud-services/frontend-components-config-utilities/cookieTransform.js:48:18)
    at ProxyServer.emit (/home/akofink/dev/repos/compliance-frontend/node_modules/eventemitter3/index.js:184:35)
    at ClientRequest.<anonymous> (/home/akofink/dev/repos/compliance-frontend/node_modules/http-proxy/lib/http-proxy/passes/web-incoming.js:133:16)
    at ClientRequest.emit (events.js:412:35)
    at tickOnSocket (_http_client.js:784:7)
    at onSocketNT (_http_client.js:823:5)
    at processTicksAndRejections (internal/process/task_queues.js:83:21) {
  code: 'ERR_HTTP_HEADERS_SENT'
}
```

### DIFF
--- a/packages/config-utils/proxy.js
+++ b/packages/config-utils/proxy.js
@@ -84,7 +84,6 @@ module.exports = ({
                     onProxyReq: cookieTransform,
                     ...(currTarget === 'PORTAL_BACKEND_MARKER' && { router }),
                     ...typeof redirect === 'object' ? redirect : {},
-                    ...(agent && { agent })
                 };
             })
         );
--- a/packages/config-utils/proxy.js
+++ b/packages/config-utils/proxy.js
@@ -83,7 +83,7 @@ module.exports = ({
                     ws: true,
                     onProxyReq: cookieTransform,
                     ...(currTarget === 'PORTAL_BACKEND_MARKER' && { router }),
-                    ...typeof redirect === 'object' ? redirect : {},
+                    ...typeof redirect === 'object' ? redirect : {}
                 };
             })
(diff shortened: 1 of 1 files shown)


akofink (approved):
Thanks! Works well. Nit: we could expose an env variable perhaps to make it configurable? Not sure if anyone needs the proxy to a local route (probably not).

redallen:
If they were to try to proxy through the corporate proxy to a local route it would crash with the above error. It's a known upstream problem that would take picking apart networking internals on our side: https://github.com/http-party/node-http-proxy/issues/1287

I think we can add the option in the future if there's demand, but currently demand is low and cost is high.

akofink:
Still an ACK from me :) Thanks again, @redallen!

Hyperkid123:
@redallen are we good to merge this?

rvsia (approved):

redallen:
@Hyperkid123 Yep, good to merge!

PR #1276, opened 2021-09-29
Update webpack dev server

Hyperkid123:
### Changes
- updates to `webpack-dev-server@4`
- handly guide: https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md

cc @redallen do you have some standalone apps handy? I was testing app with webpack proxy and it seems to be OK. There might be more work required for the standalone proxy as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4539,7 +4539,6 @@
         },
         "node_modules/@nodelib/fs.scandir": {
             "version": "2.1.4",
-            "dev": true,
             "license": "MIT",
             "dependencies": {
                 "@nodelib/fs.stat": "2.0.4",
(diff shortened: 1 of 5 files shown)


redallen:
@Hyperkid123 I use https://github.com/redhatinsights/access-requests-frontend as a standalone app test. You could also use https://github.com/RedHatInsights/frontend-starter-app (although I haven't tested it in a few months).

My guess is that webpack 5 wouldn't break anything, but it's worth trying.

karelhala (approved):

Hyperkid123:
Tested with the standalone version it is working.

PR #1301, opened 2021-11-02
Fix group filter when selected is empty

rvsia:
When `selected` is undefined, the group filter cannot handle it, this add a conditional check.

Also some test cleanups.


karelhala (approved):

bastilian (approved):
Tested with compliance and makes the OS filter function as expected! Thank you!

Hyperkid123 (approved):

PR #1303, opened 2021-11-02
Remove PF styles from webpack build.

Hyperkid123:
We need a small change in chrome to make sure all styles are provided to the platform.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31503,9 +31503,9 @@
             }
         },
         "node_modules/webpack": {
-            "version": "5.59.1",
-            "resolved": "https://registry.npmjs.org/webpack/-/webpack-5.59.1.tgz",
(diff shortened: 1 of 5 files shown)


bastilian:
Does this mean that patternfly style from applications, included in the packages they depend on, will not be used, or?

Hyperkid123:
> Does this mean that patternfly style from applications, included in the packages they depend on, will not be used, or?

 @bastilian  I've just sent a memo about it. You can find more details there. You were probably not using your specific version anyway. :wink:

Hyperkid123:
Chrome is now ready for this change in all envs.

rvsia (approved):
Tested in Sources-UI. PF styles were included only 1x.

karelhala (approved):

PR #1305, opened 2021-11-08
Allow local chrome development.

Hyperkid123:
### Changes
- add a config to remove chrome routing, this will allow running chrome locally without any other services


rvsia (approved):

Hyperkid123:
We will need a bit more tweaking to make the `historyAPIFallback` work

Hyperkid123:
@rvsia ok I've added bypass for HTML requests. They will be all routed to the dev server index.html (which is what we want).

packages/config-utils/proxy.js
```diff
@@ -179,6 +180,17 @@ module.exports = ({
                 return false;
             },
             target,
+            bypass: (req) => {
```
rvsia, line 183: I think we could do `...isChrome && { bypass: ... }`

packages/config-utils/proxy.js
```diff
@@ -31,7 +31,8 @@ module.exports = ({
     useCloud = false,
     target = '',
     keycloakUri = '',
-    registry = []
+    registry = [],
+    isChrome = false
```
rvsia, line 35: The last thing - could we add this to readme.documenation

And we should also pass the same option here: https://github.com/RedHatInsights/frontend-components/blob/266c00d4949a1664ef58732428021780a3059a8b/packages/config/src/config.js#L187

Hyperkid123: I don't want to add this to docs yet. It's something that should only be used in chrome and nowhere else. It has the potential to break the env if used incorrectly. It would like to keep it :secret:.

rvsia: Okay, I get it.

So just please add it to the main config arguments. (Or does chrome use the proxy package directly?)

Hyperkid123: Yeah it only uses the proxy. It has its own webpack config.

rvsia: Then we are ready to go 🚀 🎸

PR #1308, opened 2021-11-11
Add HTTPSProxyAgent to prod envs.

Hyperkid123:
Some non-GET requests are having auth issues due to different origin header in production as well when using the proxy. We have to re-write it.

### DIFF
--- a/packages/config-utils/proxy.js
+++ b/packages/config-utils/proxy.js
@@ -51,12 +51,16 @@ module.exports = ({
     }
 
     let agent;
+
+    if (env.startsWith('prod') || env.startsWith('stage')) {
+        // PROD and stage are deployed with Akamai which requires a corporate proxy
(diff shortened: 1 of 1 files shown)


packages/config-utils/proxy.js
```diff
@@ -55,7 +55,10 @@ module.exports = ({
         // stage-stable / stage-beta branches don't exist in build repos
         // Currently stage pulls from QA
         env = env.replace('stage', 'qa');
-        // QA and stage are deployed with Akamai which requires a corporate proxy
+    }
+
+    if (env.startsWith('prod') || env.startsWith('qa')) {
```
karelhala, line 60: Shouldn't here be also `stage` condition?

Hyperkid123: You know what, lets make it a bit cleaner :wink:

rvsia (approved):

PR #1313, opened 2021-11-15
Make meta.notifications consume a function

fhlavac:
Added a possibility to pass a function as a value to `meta.notifications`. 
The function receives dispatched action's payload and it's expected to return a notification object.

@Hyperkid123


Hyperkid123 (approved):
:tada:

PR #1314, opened 2021-11-19
Fix webpack dev server container loading

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-17016

### Changes
- fix issues with remote container initialization in local development
- stop injecting JS into the HTML template

### Remote container init issue

The optimization configuration entry was causing issues in local development environment. It was unable to initialize the self-referencing remote container. The specific configuration that was causing the issue is the [`splitchunks all`](https://webpack.js.org/plugins/split-chunks-plugin/#splitchunkschunks) setting. We can remove the split chunks config as module federations add additional splitting behavior.

### DIFF
--- a/packages/config/src/config.js
+++ b/packages/config/src/config.js
@@ -45,35 +45,6 @@ module.exports = ({
     return {
         mode: mode || (isProd ? 'production' : 'development'),
         devtool: false,
-        optimization: {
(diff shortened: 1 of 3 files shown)


karelhala (approved):
kill it with :fire:

PR #1316, opened 2021-11-22
Add extensions plugin and adjust fed mod plugin

karelhala:
### Extensions plugin

Since we want to enable async extensions we require a webpack plugin that actually makes all the required pieces. This PR is inspired by openshift/console plugin system (notice the `remoteEntryCallback`) so we can easilly plug into hac effort. However should the need for different manifest and entry arise we can easilly change it by passing different options. We are using fed-mods config under the hood so we can easilly deploy sharable modules.

### Changes to fed modules config

As mentioned we will use the shared fed-modules config, this required slight adjustments to it (omstly around lib config).

### DIFF
--- a/packages/config-utils/README.md
+++ b/packages/config-utils/README.md
@@ -123,3 +123,71 @@ This packages exposes these federated shared dependencies
 * `@redhat-cloud-services/frontend-components` - version taken from your `package.json`
 * `@redhat-cloud-services/frontend-components-utilities` - version taken from your `package.json`
 * `@redhat-cloud-services/frontend-components-notifications` - version taken from your `package.json`
+
+## Extensions plugin
+
+In order to share some code into extension points or to add new extrnsion point we can use `ExtensionsPlugin`
+
+Simply import it in your webpack config and add it to your plugins
+
+```JS
+const { resolve } = require('path');
+const config = require('@redhat-cloud-services/frontend-components-config');
+const ExtensionsPlugin = require('@redhat-cloud-services/frontend-components-config/extensions-plugin');
+
+const { config: webpackConfig, plugins } = config({
+  rootFolder: resolve(__dirname, '../'),
+  ...(process.env.BETA && { deployment: 'beta/apps' }),
+});
+
+plugins.push(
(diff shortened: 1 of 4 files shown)


packages/config-utils/README.md
```diff
@@ -123,3 +123,71 @@ This packages exposes these federated shared dependencies
 * `@redhat-cloud-services/frontend-components` - version taken from your `package.json`
 * `@redhat-cloud-services/frontend-components-utilities` - version taken from your `package.json`
 * `@redhat-cloud-services/frontend-components-notifications` - version taken from your `package.json`
+
+## Extensions plugin
+
+In order to share some code into extension points or to add new extrnsion point we can use `ExtensionsPlugin`
```
fhlavac, line 129:
```suggestion
In order to share some code into extension points or to add new extension point we can use `ExtensionsPlugin`
```

packages/config-utils/extensions-plugin.js
```diff
@@ -0,0 +1,54 @@
...
+        if (!plugin) {
+            throw new Error('Missing plugin config!');
+        }
+
+        this.plugin = plugin;
+        this.fedMod = {
+            moduleName: 'plugin-entry',
+            libName: 'window.loadPluginEntry',
+            useFileHash: false,
+            libType: 'jsonp',
```
Hyperkid123, line 16: I am wondering if this will cause issues. Is it correct that the extensions are built with webpack 4? If so, I am afraid it can cause conflicts in the webpack namespace. Or are the CRC extension have a separate build using webpack 5?

Hyperkid123 (commented):
Looks ok in general.

packages/config-utils/extensions-plugin.js
Hyperkid123: Will you be testing the plugin before merging this PR? Or do you want to just do a "silent" release?

Hyperkid123 (approved):

packages/config-utils/extensions-plugin.js
karelhala: I've managed to run it, let's merge it!

PR #1322, opened 2021-12-15
Add full TS support.

Hyperkid123:
This is an initial PR to add proper Typescript support for all packages.

Changes are tested and demonstrated on the notifications package. The current config allows for gradual migration. Not every file in a package must be converted to TS in order to use TSC for the build.

### Changes
- add root tsconfig
- add lint config for TS files
- add JS config for TS files
- create a copy of styles for ESM output (we might want to use something like the babel plugin we have for babel builds in the future)
- add TS dependencies to root
- convert notifications package to TS


rvsia (approved):

packages/notifications/src/Notification/Notification.tsx
```diff
@@ -0,0 +1,129 @@
+import React, { useEffect, useRef } from 'react';
+import { Alert, TextContent, Text, TextVariants, AlertActionCloseButton, AlertVariant } from '@patternfly/react-core';
+import { CloseIcon } from '@patternfly/react-icons';
+import './notification.scss';
+
+export interface NotificationProps {
```
karelhala, line 6: This will be a lengthy discussion I can say it right now :smile: 

Do we want to use `interface` for this one? Instead of `type`? I went trough couple of articles and the decision is on us, I just want to understand if we know what is the reasoning.

Basically I would use interface for props that will be extended by apps, meaning someone were to import `NotificationProps` add their own props and then use that interface. Which is mostly not the case. Morever this can be achieved with types as well. If we were to write classes I would be completely fine with writing interface for that case.

packages/notifications/src/Notification/Notification.tsx
```diff
@@ -0,0 +1,129 @@
...
+          */
+         onDismiss: (id: number|string) => void,
+         /**
+          * Unique ID
+          */
+         id: number | string,
+         /**
+          * Alert variant. <a href="https://www.patternfly.org/v4/components/alert#types" target="_blank">More info.</a>
+          */
+         variant: AlertVariant,
```
karelhala, line 26: I don't think the variant is required prop. PF picks automatically `info` if no variant is passed.

packages/notifications/src/Notification/Notification.tsx
```diff
@@ -0,0 +1,129 @@
...
+        /**
+     * Flag to automatically call `onDismiss` after `dismissDelay` runs out.
+     */
+         autoDismiss?: boolean,
+         /**
+          * Flag to show/hide notification close button.
+          */
+         dismissable?: boolean,
+         /**
+          * Funcation called after dismiss action is triggered. (id) => void
```
karelhala, line 16:
```suggestion
          * Function called after dismiss action is triggered. (id) => void
```

packages/notifications/src/Notification/Notification.tsx
```diff
@@ -0,0 +1,129 @@
...
+    useEffect(() => {
+        setDismissTimeout();
+        return () => {
+            clearDismissTimeout();
+        };
+    }, []);
+    return (
+        <Alert
+            className="notification-item"
+            id={`${id}`}
```
karelhala, line 98: No need for extra string.
```suggestion
            id={id}
```

packages/notifications/src/Notification/Notification.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+    const handleDismiss = () => {
+        onDismiss(id);
+    };
+
+    const setDismissTimeout = () => {
+        if (autoDismiss) {
+            dismissTimeout.current = setTimeout(() => handleDismiss(), dismissDelay);
+        }
+    };
```
karelhala, line 81: I don't think this needs to be a function. We can just write this in the `useEffect` and be done with it.

packages/notifications/src/Notification/Notification.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+    const clearDismissTimeout = () => {
+        if (dismissTimeout.current) {
+            clearTimeout(dismissTimeout.current);
+        }
+    };
```
karelhala, line 87: Same as above, I don't think we need custom function for this one.

packages/notifications/src/NotificationPagination/NotificationPagination.tsx
```diff
@@ -11,7 +10,14 @@ import {
     PaginationVariant
 } from '@patternfly/react-core';
 
-export const NotificationPagination = ({ page, onSetPage, onClearAll, count }) => (
+interface NotificationPaginationProps {
```
karelhala, line 13: Same as above, we are not even exporting this interface, so I guess let's port it to `type`

packages/notifications/src/NotificationPortal/NotificationPortal.tsx
```diff
@@ -0,0 +1,31 @@
...
+    [key: string]: any,
+    notifications?: PortalNotificationConfig[]
+}
+
+export interface NotificationPortalProps {
+    removeNotification: ActionCreator<PortalNotificationConfig>,
+    clearNotifications: ActionCreator<Record<string, unknown>>,
+}
+
+const mapStateToProps = ({ notifications }: NotificationsState, initialProps: NotificationsState) => ({
```
karelhala, line 16: What do you think about moving this to a `useSelector`, `useDispatch` and wrapping the `Portal` into function component?

packages/notifications/src/Portal/Portal.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export interface PortalProps {
+    notifications?: PortalNotificationConfig[],
+    removeNotification: (id: number | string) => void,
+    onClearAll?: () => void,
+    rootId?: string
+}
+
+const Portal: React.ComponentType<PortalProps> = ({ notifications = [], removeNotification, rootId, onClearAll }) => {
+    const [ state, setState ] = useState({ page: 1 });
```
karelhala, line 26: I would write a type for the state as well.

packages/notifications/src/notificationsMiddleware/notificationsMiddleware.ts
```diff
@@ -34,10 +43,28 @@ const shouldDispatchDefaultError = ({
     hasCustomNotification,
     noErrorOverride,
     dispatchDefaultFailure
+}: {
+    isRejected?: boolean,
+    hasCustomNotification?: boolean,
+    noErrorOverride?: boolean,
+    dispatchDefaultFailure?: boolean
```
karelhala, line 50: I would write type and export it instead of direct typing.

packages/notifications/src/notificationsMiddleware/notificationsMiddleware.ts
```diff
@@ -34,10 +43,28 @@ const shouldDispatchDefaultError = ({
...
 }) => isRejected && !hasCustomNotification && !noErrorOverride && dispatchDefaultFailure;
 
-export const createNotificationsMiddleware = (options = {}) => {
-    const defaultOptions = {
+interface CreateNotificationsMiddlewareOptions {
```
karelhala, line 53: I think that this is a nice example of `type` instead of `interface`.

packages/notifications/src/notificationsMiddleware/notificationsMiddleware.ts
```diff
@@ -50,16 +77,16 @@ export const createNotificationsMiddleware = (options = {}) => {
     };
     const middlewareOptions = { ...defaultOptions, ...options };
 
-    const matchPending = type => type.match(new RegExp(`^.*${middlewareOptions.pendingSuffix}$`));
-    const matchFulfilled = type => type.match(new RegExp(`^.*${middlewareOptions.fulfilledSuffix}$`));
-    const matchRejected = type => type.match(new RegExp(`^.*${middlewareOptions.rejectedSuffix}$`));
+    const matchPending = (type: string) => Boolean(type.match(new RegExp(`^.*${middlewareOptions.pendingSuffix}$`)));
+    const matchFulfilled = (type: string) => Boolean(type.match(new RegExp(`^.*${middlewareOptions.fulfilledSuffix}$`)));
+    const matchRejected = (type: string) => Boolean(type.match(new RegExp(`^.*${middlewareOptions.rejectedSuffix}$`)));
```
karelhala, line 82: Might be cool to use [user-defined-type-guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards), it would require a bit of rewrite, so I'll leave it up to you (you can open a new issue to address this later)

packages/notifications/src/redux/reducers/notifications.ts
```diff
@@ -0,0 +1,54 @@
...
+
+function generateID(type: string): string {
+    return 'cloud-services' + type + '-' + new Date().getTime() + Math.random().toString(36).slice(2);
+}
+
+export interface AddNotificationPayload extends Omit<PortalNotificationConfig, 'id'> {
+    id?: string | number;
+}
+
+const addNotification = (notifications: PortalNotificationConfig[], { payload }: { payload: AddNotificationPayload }): PortalNotificationConfig[] =>
```
karelhala, line 16: I would write a new type for the function, because it's a bit complicated to read.

packages/notifications/src/redux/reducers/notifications.ts
```diff
@@ -0,0 +1,54 @@
...
+    [ ...notifications, { id: generateID('notification'), ...payload }];
+
+const removeNotification = (notifications: PortalNotificationConfig[], { payload }: { payload: string | number }) => {
```
karelhala, line 19: Same as above
```TS
type RemoveNotification = (
  notifications: PortalNotificationConfig[],
  { payload }: { payload: string | number }
) => PortalNotificationConfig[];
```

And use it as
```suggestion
const removeNotification: RemoveNotification = (notifications, { payload }) => {
```

packages/notifications/src/redux/reducers/notifications.ts
```diff
@@ -0,0 +1,54 @@
...
+    switch (action.type) {
+        case ADD_NOTIFICATION:
+            return addNotification(state, action);
+        case REMOVE_NOTIFICATION:
+            return removeNotification(state, action as unknown as RemoveNotificationAction);
+        case CLEAR_NOTIFICATIONS:
+            return clearNotifications();
+        default:
+            return state;
+    }
```
karelhala, line 50: I don't like switches at all, can we use `applyReducerHash`?

karelhala (commented):
A few question regarding mostly around usage of `interface` vs `type` I'd like to use `type` for all props and function declarations unless we are certain we'll expect consumers to enhance them.

package.json
```diff
@@ -61,7 +61,16 @@
             "PFReactTable": "@patternfly/react-table"
         },
         "testURL": "http://localhost:5000/",
-        "globalSetup": "<rootDir>/config/globalSetup.js"
+        "globalSetup": "<rootDir>/config/globalSetup.js",
```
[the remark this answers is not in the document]
Hyperkid123: I agree. I was already tempted to do it here, but I think we should do it separately.

packages/notifications/package.json
```diff
@@ -36,6 +37,9 @@
         "redux-promise-middleware": "6.1.2",
         "@redhat-cloud-services/frontend-components-utilities": "*"
     },
+    "devDependencies": {
+        "typescript": "^4.5.4"
```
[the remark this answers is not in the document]
Hyperkid123: This has to be there until we have it in all packages. Node has issues with finding the TSC binary otherwise.

packages/notifications/src/Notification/Notification.tsx
Hyperkid123: I used an interface here because I use this type but changed the id attribute for the redux functions as they do not require it.

My POV at the interface/type discussions is that components prop is a component interface, therefor I used interface. But honestly, I don't really care. It does the same thing in the end.

packages/notifications/src/Notification/Notification.tsx
Hyperkid123: It's not required by PF but I think we should require it.

packages/notifications/src/Notification/Notification.tsx
```diff
@@ -0,0 +1,129 @@
+import React, { useEffect, useRef } from 'react';
+import { Alert, TextContent, Text, TextVariants, AlertActionCloseButton, AlertVariant } from '@patternfly/react-core';
+import { CloseIcon } from '@patternfly/react-icons';
+import './notification.scss';
+
+export interface NotificationProps {
+        /**
+     * Flag to automatically call `onDismiss` after `dismissDelay` runs out.
+     */
+         autoDismiss?: boolean,
```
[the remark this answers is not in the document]
Hyperkid123: We should finally enable prettier for this repo as well.

packages/notifications/src/Notification/Notification.tsx
Hyperkid123: Well. Sadly it is. We require a number or string as an ID. PF only allows string. So you have to do some type of conversion.

packages/notifications/src/Notification/Notification.tsx
Hyperkid123: This function is also used in the `onMouseLeave` event in the Alert. It resets the timer if you hover over the notification.

packages/notifications/src/Notification/Notification.tsx
Hyperkid123: Same as above

packages/notifications/src/NotificationPortal/NotificationPortal.tsx
Hyperkid123: We could, but there is a lot of redux TS fiddling that has to be done. I don't want to deal with that right now.

packages/notifications/src/notificationsMiddleware/notificationsMiddleware.ts
Hyperkid123: I would do this later on. Same as with the class. I don't want to bother with that right now. I just need something published to verify the build and package works.

packages/notifications/src/redux/reducers/notifications.ts
Hyperkid123: Again we can do that later. We would first need `applyReducerHash` typings to make it work properly. And honestly, this is such a simple reducer I don't really see any benefit of using the object. Apart from getting rid of the switch startement.

packages/notifications/src/Notification/Notification.tsx
karelhala: I went trough a couple of articles and the most recent ones mentioned that `interfaces` should be used if you expect the consumer to export it. I'd like to come with same approach for all typings honestly, so let's either go with all interfaces or prefer `type` (which is better when typing funcitons) and only rarerly use `interface` (I would prefer type as it looks like it's the easier and more straight forward case).

packages/notifications/src/NotificationPortal/NotificationPortal.tsx
karelhala: Let's create an issue and implement it later on!

packages/notifications/src/notificationsMiddleware/notificationsMiddleware.ts
karelhala: Good point! Let's create an issue and implement it later on.

packages/notifications/src/redux/reducers/notifications.ts
karelhala: Yup, new issue and implement it later.

packages/notifications/src/Notification/Notification.tsx
Hyperkid123: As I said. I don't really mind either way. This specific has to be an interface for the reasons above. Once we verify the build works and the package is usable, We can create some sort of manifesto or w/e it called and come up will all the TS conventions we want to use in the future.

PR #1324, opened 2022-01-06
Enable TS docs generation.

Hyperkid123:
With the TS additions, the docs generator was failing the build. This PR prepares the groundwork for generating docs from TS/TSX source files.

Additional work is required to properly generate the content
- in TSX files, the prop types are not properly generated some values always result in `undefined`
- in TS files I had to use tsdoc instead of JSdoc so the content is not generated at all.


packages/docs/docgen.js
```diff
@@ -8,16 +8,25 @@ const createJsdocContent = require('./functions-generator');
...
+const DEFAULT_TS_BABEL_OPTIONS  = {
+    configFile: './.docgen.babelrc',
+    root: __dirname
+};
+
 const componentsSrc = path.resolve(__dirname, '../*/src');
-const files = glob.sync(`${componentsSrc}/**/*.js`).filter(file => !file.match(/((test|spec|index).js|(\/__mock__\/|\/__mocks__\/|\/test\/))/gmi));
+const files =
+    glob
+    .sync(`${componentsSrc}/**/*.{js,ts,tsx}`)
```
karelhala, line 19: How about excluding `d.ts`?
```suggestion
    .sync(`${componentsSrc}/**/*.[!d.]{js,ts,tsx}`)
```

Not sure about the dot, if it needs to be escaped or not.

packages/docs/functions-generator.js
```diff
@@ -1,6 +1,35 @@
 const jsdocParser = require('jsdoc3-parser');
+const path = require('path');
+const fse = require('fs-extra');
+const { exec } = require('child_process');
 
-async function parseFile(file) {
+async function parseTSFile(file) {
+    return new Promise((resolve, reject) => {
```
karelhala, line 7: I know it's the same pattern as previously used. But I am not a fan of callbacks, could this be solved by using [util.html#utilpromisifyoriginal](https://nodejs.org/api/util.html#utilpromisifyoriginal)? That way we'd fully use `async`/`await` and no need to mess with callbacks and nesting.

karelhala (approved):
Looking good! Just two small questions.

packages/docs/docgen.js
Hyperkid123: @karelhala the `.d.ts` files are filtered on the line below. The glob pattern does not play very nicely with eclusions. I find it simpler to read to split the exclusions.

packages/docs/functions-generator.js
Hyperkid123: I mean it could, but then this one function would be entirely different to every other user to generate the docs. 

I can imagine this to be a nice task for one of our upcoming interns to get familiar with the environment.

karelhala: Good point! I've created issue to addess this https://github.com/RedHatInsights/frontend-components/issues/1326

packages/docs/docgen.js
karelhala: Mhh it feels a bit too heavy to call `filter` just to remove `.d.ts` files if we could just add `?!(d.)` to it (previously the ignore was not correct).

That being said if you'd still prefer the filter to do the job I'd rather join them together. This way we are perfomring one extra loop just to ignore them, whic honestly feels a bit too complex.

Hyperkid123: @karelhala I don't think glob supports ?!(d.). That is why I've gone with the additional filter.

karelhala (approved):

PR #1329, opened 2022-01-13
Add option to load chrome HTML template via proxy.

Hyperkid123:
The HTML webpack will be copied to webpack output folder instead of using the generated template. 

Why copy instead of using a proxy? The same reason why we have difficulties loading the ESI tags. The headers for the template are already sent to the client which makes it impossible to mutate the actual response.

### DIFF
--- a/packages/config-utils/proxy.js
+++ b/packages/config-utils/proxy.js
@@ -32,7 +32,8 @@ module.exports = ({
     target = '',
     keycloakUri = '',
     registry = [],
-    isChrome = false
+    isChrome = false,
+    onBeforeSetupMiddleware = () => {}
 }) => {
     const proxy = [];
     const majorEnv = env.split('-')[0];
@@ -251,6 +252,8 @@ module.exports = ({
                         overwrite: true
                     });
                 }
+
+                onBeforeSetupMiddleware({ chromePath });
 
                 if (chromePath) {
                     registerChrome({
--- a/packages/config/README.md
+++ b/packages/config/README.md
@@ -7,6 +7,7 @@
     - [Removed features with webpack 5](#removed-features-with-webpack-5)
   - [useProxy](#useproxy)
     - [Attributes](#attributes)
+      - [useChromeTemplate](#useChromeTemplate)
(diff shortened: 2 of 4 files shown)


packages/config-utils/proxy.js
```diff
@@ -252,6 +254,8 @@ module.exports = ({
                     });
                 }
 
+                onBeforeSetupMiddleware({ chromePath });
```
rvsia, line 257: can be chromePath null? there is a condition to check it just below it

packages/config-utils/proxy.js
```diff
@@ -2,6 +2,7 @@
 // Webpack proxy and express config for `useProxy: true` or `standalone: true`
 const { execSync } = require('child_process');
 const path = require('path');
+const fs = require('fs');
```
rvsia, line 5: `fs` is not used here

packages/config/README.md
```diff
@@ -64,6 +66,23 @@ const { config: webpackConfig, plugins } = config({
 |[useCloud](#use-cloud)|`boolean`|Toggle to use old fallback to cloud.redhat.com paths instead of console.redhat.com.|
 |[target](#target)|`string`|Override `env` and `useCloud` to use a custom URI.|
 
+
+#### experimentalTemplate
```
rvsia, line 70: Maybe we could change name of this option, I think it's not very clear.

examples:
`loadChromeTemplate`
`useChromeTemplate`
`loadChromeHtml`
etc.

Hyperkid123: This option will eventually go away and will become part of the proxy.

packages/config-utils/proxy.js
Hyperkid123: I think I will leave it in here because I think we should have a consistent API, but add a check in the consumer.

packages/config/README.md
Hyperkid123: but I like `useChromeTemplate` better.

rvsia (approved):

PR #1331, opened 2022-01-14
Add TS guidelines.

Hyperkid123:


README.md
```diff
@@ -186,3 +186,161 @@ If none package suits scope of new changes, we need to create new package by cre
...
 
 You can also draft a release by adding label `release` or `release minor` and once this PR is merged new release will be triggered.
+
+## Typescript guidelines
+
+Typescript build has been finally introduced to the repository. All modules will be eventually refactored into typescript. When refactoring, or creating new modules, please follow these instructions when creating types:
+
+### React components
+
+1. All component props must be an interface and must be exported. This is required for consuming apps to use features like `omit` or `extends` when developing with typescript. A nice example is generating JSX from array structure, but for some reseason extension/exclustion of the interface is required.
```
karelhala, line 196: Do we want to use `interface`s? Consumers can extend types as well (either with [intersection-types](https://www.typescriptlang.org/docs/handbook/2/objects.html#intersection-types) or since TS 2.2 they can use interface and directly use `extends` the type). I think that using `type` gives us a bit more flexibility (type can't extend interface, but interface can extend type) and looks like many upstream projects prefer types as well (only for some specific usecases they export `interface`).

README.md
```diff
@@ -186,3 +186,161 @@ If none package suits scope of new changes, we need to create new package by cre
...
+        title,
+        description: message
+    }));
+
+    return alertsStructure.map(({ id, ...props }) => <AlertWrapper key={id} {...props} />);
+};
+
+```
+
+2. When creating a wrapper around PF components, always extend original components props. This will ensure that all original props are accepted by your component wrapper. If you have to exclude some original prop, use `Omit` generic. Exclusion might be required because the wrapper is setting a prop by default.
```
karelhala, line 231: God damn, I wanted to go with the types as they are giving us more flexibility. But upon looking at PF they straight up exort `interface` and that's it. Looks like we don't have any other chance (unless we want to have a mix of types and interfaces)

README.md
```diff
@@ -186,3 +186,161 @@ If none package suits scope of new changes, we need to create new package by cre
...
+```ts
+for (var i = 0, n = str.length; i < 10; i++) { }
+if (x < 10) { }
+function f(x: number, y: string): void { }
+```
+5. Use a single declaration per variable statement
+```ts
+(i.e. use var x = 1; var y = 2; over var x = 1, y = 2;).
+```
+6. else goes on a separate line from the closing curly brace.
```
karelhala, line 346: I would remove this rule, can we also add to use style rule about single quotes?

karelhala (commented):
I really wanted to go with the newer approach of types over interfaces, but PF exports just interfaces. Which is super dumb.

README.md
Hyperkid123: Yup. That is why I would go with interfaces. As we build on top PF most of the time, it is probably for the best to keep it same.

README.md
Hyperkid123: I will remove this one. For quotes, let's handle that through eslint.

karelhala (approved):

PR #1333, opened 2022-01-17
Use common FEC eslint config for the whole repository

Hyperkid123:
### Changes
- use eslint config from `packages/eslint-config`
- enable prettier for the whole repository
- apply automated and manual fixes


rvsia (approved):
👍

PR #1383, opened 2022-01-27
Document arguments of `getUserPermissions()`

leSamo:


Hyperkid123 (approved):
Thanks for the content. The docs should be actually deployed soon :tm:

PR #1387, opened 2022-02-02
Migrate async component to TS

### LINKED ISSUE #1388: AsyncComponent.js
file: https://github.com/RedHatInsights/frontend-components/tree/master/packages/components/src/AsyncComponent

karelhala:
Jira: RHCLOUD-17864

Closes #1388


packages/components/src/AsyncComponent/index.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import { ScalprumComponent } from '@scalprum/react-core';
+import { Bullseye, Spinner } from '@patternfly/react-core';
+import classNames from 'classnames';
+
+export interface AsyncComponentProps {
```
Hyperkid123, line 6: These props should extend a structure I call `AnyObject`. If we don't use that we forbid any other props that may be required by the child component.

Any object usually looks like this
```TSX
interface AnyObject {
  [key: string]: any
}
```

Hyperkid123 (approved):

rvsia (approved):

Issue #1389, opened 2022-02-02
CullingInformation.js


karelhala:
file: https://github.com/RedHatInsights/frontend-components/tree/master/packages/components/src/CullingInfo/CullingInformation.js

PR #1390, opened 2022-02-02
Transfer culling info to TS

karelhala:
Closes: #1389


packages/components/src/CullingInfo/CullingInformation.tsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+import { ExclamationCircleIcon, ExclamationTriangleIcon } from '@patternfly/react-icons';
+import { Tooltip } from '@patternfly/react-core';
+import classnames from 'classnames';
+import './CullingInformation.scss';
+import { CullingDate, calculateTooltip, CullingInfo } from './utils';
+export type Render = (config: { msg: string }) => React.ReactElement<any, any> | null;
+export interface CullingInformation {
```
Hyperkid123, line 8: As we are passing the `{...props}` to the `Tooltip` component, the interface should extend the `TooltipProps` PF component.

Hyperkid123 (changes requested):

packages/components/src/CullingInfo/CullingInformation.tsx
karelhala: Good point!

Hyperkid123 (approved):

rvsia (approved):

PR #1400, opened 2022-02-08
Add prop to disable bulk expand button in primary toolbar

leSamo:


packages/components/src/PrimaryToolbar/PrimaryToolbar.js
```diff
@@ -149,6 +151,7 @@ PrimaryToolbar.propTypes = {
     PropTypes.shape({
       onClick: PropTypes.func,
       isAllExpanded: PropTypes.bool,
+      isDisabled: PropTypes.bool
```
karelhala, line 154: Can we add `buttonProps` as well?
```suggestion
      isDisabled: PropTypes.bool,
      buttonProps: PropTypes.shape({
        [PropTypes.string]: PropTypes.object
       })
```

karelhala (commented):
Thank you for the PR, this will help further improvements to the button. I like the props option which allows custom props to be passed and it can rewrite any prop.

karelhala:
@leSamo can you run `npm run lint -- --fix` to fix the linting errors?

karelhala (approved):

Issue #1368, opened 2022-01-25
FilterDropdown.js


Hyperkid123:
file: https://github.com/RedHatInsights/frontend-components/tree/master/packages/components/src/Filters/FilterDropdown.js

@aneelac22 please also convert the class components to function components.

PR #1401, opened 2022-02-08
Migrate Filters to TypeScript

### LINKED ISSUE #1367: FilterInput.js
file: https://github.com/RedHatInsights/frontend-components/tree/master/packages/components/src/Filters/FilterInput.js

@aneelac22 please also convert the class components to function components.

aneelac22:
closes #1367 #1368 

[Jira Ref RHCLOUD-17754]


packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { Dropdown, DropdownToggle, Level, DropdownProps } from '@patternfly/react-core';
+import FilterInput from './FilterInput.js';
+import './filter-dropdown.scss';
+
+export interface FilterDropdownProps extends Omit<DropdownProps, 'onSelect'> {
+  addFilter: (filterName: string | number, selectedValue: string | number, type: string) => void;
+  removeFilter: (filterName: string | number, selectedValue: string | number) => void;
+  hideCategories: [];
+  filters: object;
+  filterCategories: [
```
Hyperkid123, line 11: Can you create and export separate type for the filterCategory?

packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,103 @@
+import React, { useState } from 'react';
+import { Dropdown, DropdownToggle, Level, DropdownProps } from '@patternfly/react-core';
+import FilterInput from './FilterInput.js';
+import './filter-dropdown.scss';
+
+export interface FilterDropdownProps extends Omit<DropdownProps, 'onSelect'> {
+  addFilter: (filterName: string | number, selectedValue: string | number, type: string) => void;
+  removeFilter: (filterName: string | number, selectedValue: string | number) => void;
+  hideCategories: [];
+  filters: object;
```
Hyperkid123, line 10: This should use the `FilterObject` props type.

packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,103 @@
+import React, { useState } from 'react';
+import { Dropdown, DropdownToggle, Level, DropdownProps } from '@patternfly/react-core';
+import FilterInput from './FilterInput.js';
+import './filter-dropdown.scss';
+
+export interface FilterDropdownProps extends Omit<DropdownProps, 'onSelect'> {
+  addFilter: (filterName: string | number, selectedValue: string | number, type: string) => void;
```
Hyperkid123, line 7: Most of these props are optional. You can check the original propTypes.

packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,103 @@
...
+  removeFilter: (filterName: string | number, selectedValue: string | number) => void;
+  hideCategories: [];
```
Hyperkid123, line 9: I think this should be an array of strings.

packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,103 @@
...
+        isChecked ? addFilter(filterName, selectedValue, type) : removeFilter(filterName, selectedValue);
+        break;
+      case 'radio':
+        addFilter(filterName, selectedValue, type);
+        break;
+    }
+  };
+
+  const onToggle = (isOpen: boolean) => {
+    setIsOpen(!isOpen);
```
Hyperkid123, line 57: I think the isOpen value should not be negated.

aneelac22: Is this even needed, can I just use onToggle and isOpen as a prop then?

packages/components/src/Filters/FilterDropdown.tsx
aneelac22: I actually got this wrong, I think filters should be array, but what is `FilterObject`?

packages/components/src/Filters/FilterDropdown.tsx
Hyperkid123: You should work around that with conditional access etc. We can't change the API. It will break existing builds.

packages/components/src/Filters/FilterDropdown.tsx
Hyperkid123: Sorry I mean `FilterInputProps` that is where the prop is used.

packages/components/src/Filters/FilterDropdown.tsx
Hyperkid123: Yes you can.

packages/components/src/Filters/FilterDropdown.tsx
Hyperkid123: I had something like this in mind instead
```typescript
export type FilterCategory = {
  title: string;
  type: string;
  urlParam: string;
  values: {
    label: React.ReactNode;
    value: string;
  }[]
}

export interface FilterDropdownProps extends Omit<DropdownProps, 'onSelect'> {
  ...
  filterCategories: FilterCategory[]
  ...
}

```

packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,108 @@
...
+  label,
+  ouiaId,
+  ouiaSafe,
+  addFilter,
+  removeFilter,
+  isOpen = false,
+  onToggle,
+  ...props
+}) => {
+  const ouiaStateId = 'RHI/FilterDropdown';
```
Hyperkid123, line 50: If this is never re-assigned we might as well bump it out of the function. There is no point in creating this variable on each render.

packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,108 @@
...
+      ];
+    }
+  ];
+}
+
+export interface FilterDropdownProps extends Omit<DropdownProps, 'onSelect'> {
+  addFilter?: (filterName?: string, selectedValue?: string | number, type?: string) => void;
+  removeFilter?: (filterName?: string, selectedValue?: string | number) => void;
+  hideCategories?: [string];
+  isOpen: boolean;
```
Hyperkid123, line 26: This attribute was originally handled through state, not props. Can you change it back, please? I think we have misunderstood ourselves in my previous review. I thought that you were asking about the internal `onToggle` function which was unnecessary and the state set function could have been called directly in the JSX part of the file.

packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,108 @@
...
+  onToggle: (isOpen: boolean) => void;
+  filters?: Record<string, string>;
+  filterCategories: filterCategoriesProps['filterCategories'];
+  onSelect: ((event?: React.SyntheticEvent<HTMLDivElement, Event> | undefined) => void) | undefined;
```
Hyperkid123, line 30: Where is this prop coming from? I can't find it in the original.

packages/components/src/Filters/FilterInput.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { useState, useEffect } from 'react';
+import { Checkbox, Radio } from '@patternfly/react-core';
+
+export interface FilterInputProps {
+  addRemoveFilters?: (value?: string, param?: string, type?: string, checked?: boolean) => void;
+  className?: string;
+  currentPage?: string;
+  id: string;
+  label: React.ReactNode;
+  param: string;
```
Hyperkid123, line 10: Based on the original implementation, the `param` prop should be optional.

packages/components/src/Filters/FilterInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+  value?: string;
+  filters?: Record<string, string>;
```
Hyperkid123, line 12: I think this should be `Record<string, unknown>`. We will describe the accepted value later, but we can't limit ourselves just to string.

packages/components/src/Filters/FilterInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+  type?: string;
+  onChange?: (checked: boolean, event: React.FormEvent<HTMLInputElement>) => void;
```
Hyperkid123, line 14: `onChange` is not present in the original prop types definition.

packages/components/src/Filters/FilterInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+const FilterInput: React.FC<FilterInputProps> = ({ addRemoveFilters, param, filters, type = 'checkbox', value, id, label }) => {
```
Hyperkid123, line 17: `filters` should have `{}` default value

packages/components/src/Filters/FilterDropdown.tsx
aneelac22: I don't get it, it just seems redundant to have setState when it is going to be the exact same value

packages/components/src/Filters/FilterInput.tsx
aneelac22: In Radio name is required

packages/components/src/Filters/FilterDropdown.tsx
Hyperkid123: Well the `isOpen` variable cannot be migrated just like to be suddenly an outside prop. That is the main issue.

It was not set to be the exact same value but it was changed to the value sent by the toggle. We were only passing it along to the state The original code is as follows
```js
  onToggle = (isOpen /** the value is set by the PF component */) => {
    this.setState({
      isOpen,
    });
  };
```

packages/components/src/Filters/FilterInput.tsx
Hyperkid123: Then we need an optional runtime type check. It is not true that the param is required for every case.

packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,103 @@
...
+  addFilter,
+  removeFilter,
+  isOpen = false,
+  onToggle,
+  ...props
+}) => {
+  const ouiaStateId = 'RHI/FilterDropdown';
+
+  const addRemoveFilters = (selectedValue?: string | number, filterName?: string, type?: string, isChecked?: boolean) => {
+    switch (type) {
```
karelhala, line 47: I know it was previously a `switch` but I prefer avoiding it like a plague. I think it would be better to refactor this a bit and change this switch to `if`/`else if` statement. And let's have both `checkbox` and `radio` as constants.

packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,103 @@
...
+      case 'checkbox':
+        isChecked ? addFilter && addFilter(filterName, selectedValue, type) : removeFilter && removeFilter(filterName, selectedValue);
```
karelhala, line 49: It's better to use the safe accessor
```suggestion
        isChecked ? addFilter?.(filterName, selectedValue, type) : removeFilter?.(filterName, selectedValue);
```

packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,103 @@
...
+        break;
+      case 'radio':
+        addFilter && addFilter(filterName, selectedValue, type);
```
karelhala, line 52: Same as above, safe accessor is better than shortcutting.
```suggestion
        addFilter?.(filterName, selectedValue, type);
```

packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,103 @@
...
+          {label}
+        </DropdownToggle>
+      }
+      isOpen={isOpen}
+      ouiaId={ouiaFinalId}
+      ouiaSafe={ouiaSafe}
+    >
+      <div className="pf-c-dropdown__menu-item">
+        {filterCategories &&
+          filterCategories.map(
```
karelhala, line 75: You can simplify this by safe accessor
```suggestion
        {filterCategories?.map(
```

packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,103 @@
...
+            (data: FilterCategory, index) =>
+              hideCategories &&
+              !hideCategories.includes(data.urlParam) && (
```
karelhala, line 78: You can simplify this by safe accessor
```suggestion
              !hideCategories?.includes(data.urlParam) && (
```

packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,103 @@
...
+                <Level key={`${data.urlParam}${index}`}>
+                  {data.title}
+                  {data.values.map((item: { label: React.ReactNode; value: string }, key: string | number) => (
```
karelhala, line 81: Let's not access the properties directly, if they are not passed down properly the component will blow up.
```suggestion
                  {data.title || ''}
                  {data.values?.map((item: { label: React.ReactNode; value: string } = {}, key: string | number) => (
```

packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,103 @@
...
+                      id={`${data.urlParam}${key}`}
+                      label={item.label}
+                      addRemoveFilters={addRemoveFilters}
+                      param={data.urlParam}
+                      type={data.type}
+                      value={item.value}
+                      filters={filters}
+                    />
+                  ))}
+                  {index !== filterCategories.length - 1 && <br />}
```
karelhala, line 94: I know this has been ported from the previous implementation, but I am curious if we could style this with css instead of `br` element.

packages/components/src/Filters/FilterInput.tsx
karelhala: It's a bit tricky but we can extend some general HTMLElement prop so we don't have to include the `onChange` prop. How about `HTMLElement`?

packages/components/src/Filters/FilterInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const [checked, setChecked] = useState(false);
+  useEffect(() => {
+    if (type !== 'radio') {
+      if (filters && value && param in filters && filters[param] && filters[param].includes(value)) {
+        setChecked(true);
+      } else {
+        setChecked(false);
+      }
+    }
+  });
```
karelhala, line 27: This effect is performed on each render, since we are firing state change in it we'd end in infinite loop. Previously it used to run only on render, which is not ideal as well, but some apps might already operate with it. Let's go for now with just that and later add `filters`, `value` and `param` as dependency of this hook. Also might be better to use safe accessor.
```suggestion
  useEffect(() => {
    if (type !== 'radio') {
      if (value && filters?.[param]?.includes?.(value)) {
        setChecked(true);
      } else {
        setChecked(false);
      }
    }
  }, []);
```

packages/components/src/Filters/FilterInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const handleChange = () => {
+    switch (type) {
+      case 'checkbox':
+        setChecked(!checked);
+        break;
+    }
```
karelhala, line 33: This is just simple `if` let's avoid `switch` as much as possible.

packages/components/src/Filters/FilterInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+    addRemoveFilters && addRemoveFilters(value, param, type, checked);
```
karelhala, line 34: Let's use safe accessor here as well.
```suggestion
    addRemoveFilters?.(value, param, type, checked);
```

packages/components/src/Filters/FilterInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+        aria-label={param}
+        id={id}
+        label={label}
+        name={param}
+        onChange={handleChange}
+        value={value}
+        ouiaId={id}
+      />
+    );
+  }
```
karelhala, line 51: No need for `else` if we are returning the value.
```suggestion
  }
  return (
    <Radio
      isChecked={filters && filters[param] === value}
      aria-label={param}
      id={id}
      label={label}
      name={param}
      onChange={handleChange}
      value={value}
      ouiaId={id}
    />
  );
```

packages/components/src/Filters/FilterInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const handleChange = () => {
+    switch (type) {
+      case 'checkbox':
+        setChecked(!checked);
+        break;
+    }
+    addRemoveFilters && addRemoveFilters(value, param, type, checked);
+  };
+  if (type !== 'radio') {
+    return <Checkbox aria-label={param} id={id} isChecked={checked} label={label} onChange={handleChange} value={value} ouiaId={id} />;
```
karelhala, line 37: Preivously the `aria-label` was `label` any reason to switch it to `param`? Also the `ouiaId` was previously `label` that would break many, many tests.
```suggestion
    return <Checkbox aria-label={label} id={id} isChecked={checked} label={label} onChange={handleChange} value={value} ouiaId={label} />;
```

packages/components/src/Filters/FilterInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+  } else {
+    return (
+      <Radio
+        isChecked={filters && filters[param] === value}
+        aria-label={param}
```
karelhala, line 42: Same as above, let's use `label` instead of `param`
```suggestion
        aria-label={label}
```

packages/components/src/Filters/FilterInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+        break;
+    }
+    addRemoveFilters && addRemoveFilters(value, param, type, checked);
+  };
+  if (type !== 'radio') {
+    return <Checkbox aria-label={param} id={id} isChecked={checked} label={label} onChange={handleChange} value={value} ouiaId={id} />;
+  } else {
+    return (
+      <Radio
+        isChecked={filters && filters[param] === value}
```
karelhala, line 41: Let's use safe accessor
```suggestion
        isChecked={filters?.[param] === value}
```

packages/components/src/Filters/FilterInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+        aria-label={param}
+        id={id}
+        label={label}
+        name={param}
+        onChange={handleChange}
+        value={value}
+        ouiaId={id}
```
karelhala, line 48: This should be `label`
```suggestion
        ouiaId={label}
```

karelhala (changes requested):
Looking good! There are just a few changes, we need to adjust. Mainly the toggle of dropdown to be changed from component managed to consumer managed.

packages/components/src/Filters/FilterInput.tsx
aneelac22: label PropType is node and aria-label is supposed to be a string

aneelac22: The other one I changed it back

packages/components/src/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  const addRemoveFilters = (selectedValue?: string | number, filterName?: string, type?: string, isChecked?: boolean) => {
+    if (type === 'checkbox') {
+      isChecked ? addFilter?.(filterName, selectedValue, type) : removeFilter?.(filterName, selectedValue);
+    } else {
+      addFilter?.(filterName, selectedValue, type);
+    }
+  };
+
+  const ouiaFinalId = ouiaId !== undefined ? ouiaId : ouiaStateId;
```
karelhala, line 51: Good thing moving `ouiaStateId` from the component! We can further improve this by assigning default value on line 35
```JS
ouiaId = ouiaStateId,
```

It's the same, but it'll make the component slightly cleaner.

packages/components/src/Filters/FilterInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+        setChecked(false);
+      }
+    }
+  }, []);
+  const handleChange = () => {
+    switch (type) {
+      case 'checkbox':
+        setChecked(!checked);
+        break;
+    }
```
karelhala, line 32: Previously it was a `switch` which got under our radar. I think it would be awesome to change it to `if` since it just checks for one `type`.

packages/components/src/Filters/FilterInput.tsx
karelhala: @aneelac22 any opinion on this one? I am fine with keeping the shortcutting, it just feels a bit more natural to use safe accessor.

packages/components/src/Filters/FilterInput.tsx
karelhala: Good point about the `label` being a node vs `string`. I'd just bluntly cast the value to `string`. We can investigate further the update to use `param` instead of `label, but I'd like to avoid API changes as much as possible.

packages/components/src/Filters/FilterInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <Radio
+      isChecked={filters && !!value && !!param && filters[param] === value}
+      aria-label={String(label)}
+      id={id}
+      label={label}
+      name={param}
+      onChange={handleChange}
+      value={value}
+      ouiaId={String(label)}
```
karelhala, line 47: Yeah! I'd change the Checkbox to this as well.

karelhala (commented):
Looking much, much better! Thank you for taking in my feedback, there are just a few cosmetic changes I'd like to apply as well. But overall looks good.

karelhala (approved):
Looking good!

PR #1416, opened 2022-02-10
port over platex docs

ryelo:


Hyperkid123 (approved):
Looks good. I will make sure the nav is generated and we have a link on the main page.

PR #1419, opened 2022-02-15
Make unleash proxy client as singleton package.

Hyperkid123:
Part of: https://issues.redhat.com/browse/RHCLOUD-15343

### TODO
- [x] config
- [x] docs
- [x] test in app: https://github.com/RedHatInsights/tower-analytics-frontend/pull/749

### DIFF
--- a/packages/config-utils/federated-modules.js
+++ b/packages/config-utils/federated-modules.js
@@ -42,6 +42,13 @@ module.exports = ({ root, exposes, shared = [], debug, moduleName, useFileHash =
     sharedDeps.push({ '@scalprum/react-core': { requiredVersion: '*', singleton: true } });
   }
 
+  /**
+   * Make sure the unleash proxy client is a singleton
+   */
+  if (dependencies['@unleash/proxy-client-react']) {
+    sharedDeps.push({ '@unleash/proxy-client-react': { singleton: true, requiredVersion: dependencies['@unleash/proxy-client-react'] } });
+  }
+
   if (debug) {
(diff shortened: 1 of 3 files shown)


packages/config-utils/federated-modules.js
```diff
@@ -42,6 +42,13 @@ module.exports = ({ root, exposes, shared = [], debug, moduleName, useFileHash =
     sharedDeps.push({ '@scalprum/react-core': { requiredVersion: '*', singleton: true } });
   }
 
+  /**
+   * Make sure the unleash proxy client is a singleton
+   */
+  if (dependencies['@unleash/proxy-client-react']) {
+    sharedDeps.push({ '@unleash/proxy-client-react': { singleton: true, requiredVersion: dependencies['@unleash/proxy-client-react'] } });
```
karelhala, line 49: The version is driven by chrome isn't it? I think it would be better to just use `singleton` in here.
```suggestion
    sharedDeps.push({ '@unleash/proxy-client-react': { singleton: true });
```

Hyperkid123: @karelhala  Yeah it is but we want to get the console warning if the app is using a different version. The API of this is not exactly stable ATM.

karelhala: Good point about the warnings, I am thinking if it wouldn't be beneficial to move it to `strictVersion` at some point in the future.

karelhala (approved):

PR #1455, opened 2022-03-03
Enable TS build in utils package.

Hyperkid123:
### Changes
- enable TSC build for utils package
- replace manually created .d.ts files in the utils package
- migrate interceptors to TS
- include chrome API types on the global window object


packages/utils/src/RouterParams/RouterParams.js
```diff
@@ -4,14 +4,30 @@ import { withRouter, matchPath } from 'react-router-dom';
...
 
-export const routerParams = (Component) => {
-  class RouterParams extends React.Component {
-    componentDidMount() {
-      const {
-        match: { params, path, url },
-        onPathChange,
-        location,
-      } = this.props;
+class RouterParams extends React.Component {
```
karelhala, line 7: Can we create a JIRA ticket to make this component as function component?

karelhala:
@Hyperkid123 yeah, that could be a good way of doing things. `@RedhatInisghts/component-types` should be a good package name. Do we want to move all types under one package, or rather multiple ones?

Hyperkid123:
@karelhala I think we should move there only types that are present in the packages. There is no point in moving everything. The first candidate will be Chrome API as it is used in multiple packages but does not originate in any of them. We will be probably finding a lot more as we move further with the migrations.

packages/components/package.json
```diff
@@ -54,5 +54,8 @@
         "@patternfly/patternfly": ">=4.102.2",
         "node-sass-package-importer": "^5.3.2",
         "@types/react": "^16.9.34"
+    },
+    "optionalDependencies": {
```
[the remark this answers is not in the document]
Hyperkid123: This can be now removed because the type used was in the ChromeAPI but now that is part of the common types package.

packages/utils/src/RouterParams/RouterParams.js
Hyperkid123: Yeah we should.

karelhala (approved):

PR #1461, opened 2022-03-04
fix(TableToolbar): Results and selected shouldn't have a default

bastilian:
With a default of 0 the component will show "0 Results" etc. unnecessarily


Hyperkid123 (commented):
I think we might require UX input on this one. @karelhala d you remember who did the original mockups?

Hyperkid123 (approved):
NVM i see the defaults were added after the TS migration.

PR #1478, opened 2022-03-15
feat: Adding ouia high order component and hook

josejulio:
Adds a high order component to wrap resources into an Ouia container or a hook if the ouia properties are meant to be set in an already existing DOM elements.

### DIFF
--- a/packages/components/src/Ouia/WithOuia.tsx
+++ b/packages/components/src/Ouia/WithOuia.tsx
@@ -1,0 +1,43 @@
+import React from 'react';
+import { OuiaProps, OuiaDataAttributes, makeOuiaAttributes } from './Ouia';
+
+/**
+ * Parameters to configure the attributes of the Ouia component.
+ */
+export type WithOuiaParams =
+  | string
+  | {
+      type: string;
+      module?: string;
+      InnerComponent?: React.ComponentType<OuiaDataAttributes> | string;
+    };
+
+/**
+ * High order component to wrap a Ouia component. Allows to set the component-type, safe and component-id properties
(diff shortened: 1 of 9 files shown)


packages/components/src/Ouia/WithOuia.tsx
```diff
@@ -0,0 +1,43 @@
...
+    };
+
+/**
+ * High order component to wrap a Ouia component. Allows to set the component-type, safe and component-id properties
+ * @param Component component that we are wrapping
+ * @param params allows to provide some configuration for the component, it can be a single string to set it as a type
+ * or an object to allow to configure the type, module and InnerComponent.
+ * @returns a Component that wraps it's content in a Ouia component.
+ */
+const WithOuia = <P,>(Component: React.ComponentType<P>, params: WithOuiaParams) => {
```
Hyperkid123, line 22: Extra `,` after the generic

josejulio: That's intended. This is a `tsx` file, if I don't add the `,` the compilar interprets the generic as `jsx` and complains.
One way around is to add the `,`. 

The other is to add an explicit `extends any`, but that triggers the following rule `ESLint: Constraining the generic type `P` to 'any' does nothing and is unnecessary.(@typescript-eslint/no-unnecessary-type-constraint)`.

I just learn now that the autofix of that rule adds the `,` :)

Hyperkid123: Oh really? I dod not know that :D This is what we get when mixing XML and TS :+1:

Hyperkid123 (approved):

PR #1489, opened 2022-03-24
[Chore] - Migrate ReducerRegistry to TS

### LINKED ISSUE #1473: ReducerRegistry.js
file: https://github.com/RedHatInsights/frontend-components/tree/master/packages/utils/src/ReducerRegistry/ReducerRegistry.js

adithyaakrishna:
Signed-off-by: Adithya Krishna <adikrish@redhat.com>

**Description:**

This PR Fixes #1473


packages/utils/src/ReducerRegistry/ReducerRegistry.ts
```diff
@@ -1,7 +1,8 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import { applyMiddleware, combineReducers, createStore, compose } from 'redux';
 
-export function applyReducerHash(reducerHash, initialState = {}) {
-  return function (state = initialState, action) {
+export function applyReducerHash(reducerHash: any, initialState: Record<string, unknown>) {
```
Hyperkid123, line 4: You have to initialize the `initialState` argument. Or will be treated as a required attribute.

packages/utils/src/ReducerRegistry/ReducerRegistry.ts
```diff
@@ -41,12 +52,12 @@ export class ReducerRegistry {
    *
    * @param newReducers the object of new reducers.
    */
-  register(newReducers) {
+  register(newReducers: Record<string, never>) {
```
Hyperkid123, line 55: Can you use `any` instead of `never`?

```TS
newReducers: Record<string, any>
```

After this change you might be able to remove the type casting bellow.

Hyperkid123 (changes requested):

packages/utils/src/ReducerRegistry/ReducerRegistry.ts
adithyaakrishna: Done, Cap'n 💯

packages/utils/src/ReducerRegistry/ReducerRegistry.ts
adithyaakrishna: I am a bit confused regarding this, would adding something like `initialState?: Record<string, unknown>` be fine in this case? 😶

rvsia: In the original code there is a default value: `initialState = {}`, you miss it the ts version.

Hyperkid123 (approved):

adithyaakrishna:
@Hyperkid123 @rvsia Apologies for not finishing this up, I had missed @rvsia 's comment in the PR and I'll make sure this won't happen again :)

PR #1493, opened 2022-04-01
Create advisor-components package and add RuleDetails component

gkarat:
This adds a new package "advisor-components" that will contain the components used in OCP, RHEL Advisors (/openshift/insights/advisor, /insights/advisor), and integrations with other services (including those outside console.redhat.com). Also adds first component RuleDetails used in both OCP and RHEL Advisors (see screenshots).

Frame 1(3)
Frame 2(1)


packages/advisor-components/src/ContextWrapper.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return routerProps ? (
+    <Router {...routerProps}>
+      <IntlProvider locale={locale} messages={translations[locale]}>
+        <ProductContext.Provider value={env}>{children}</ProductContext.Provider>
+      </IntlProvider>
+    </Router>
+  ) : (
+    <IntlProvider locale={locale} messages={translations[locale]}>
+      <ProductContext.Provider value={env}>{children}</ProductContext.Provider>
```
karelhala, line 36: Wouldn't it be better to pass the `env` values directly to the components? I see the usage just in `RuleDetailsBase` and we can pass the `env` directly to it.

packages/advisor-components/src/RebootRequired/RebootRequired.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  return (
+    <span className="system-reboot-message">
+      <PowerOffIcon className={reboot_required ? 'reboot-required-icon' : 'no-reboot-required-icon'} />
+      <TextContent className="system-reboot-message__content">
+        <Text component={TextVariants.p}>
+          {intl.formatMessage(messages.systemReboot, {
+            strong: (str) => <strong>{str}</strong>,
+            status: reboot_required ? intl.formatMessage(messages.is) : intl.formatMessage(messages.isNot),
+          })}
```
karelhala, line 25: This should be sent in props.

packages/advisor-components/src/RebootRequired/RebootRequired.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { PowerOffIcon } from '@patternfly/react-icons'; // TODO: change PF imports
+import { useIntl } from 'react-intl';
+
+import messages from '../messages';
+
+interface RebootRequiredProps {
+  reboot_required: boolean;
+}
+
+const RebootRequired: React.FC<RebootRequiredProps> = (reboot_required) => {
```
karelhala, line 14: Probably a typo, this should be destructed
```suggestion
const RebootRequired: React.FC<RebootRequiredProps> = ({ reboot_required }) => {
```

packages/advisor-components/src/RebootRequired/RebootRequired.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import React from 'react';
+import { Text, TextContent, TextVariants } from '@patternfly/react-core';
+import { PowerOffIcon } from '@patternfly/react-icons'; // TODO: change PF imports
+import { useIntl } from 'react-intl';
+
+import messages from '../messages';
+
+interface RebootRequiredProps {
+  reboot_required: boolean;
```
karelhala, line 11: Can you use cammel case instead of snake one?
```suggestion
  rebootRequired: boolean;
```

packages/advisor-components/src/RuleDetails/RuleDetails.scss
```diff
@@ -0,0 +1,91 @@
...
+.ins-c-rule-details__split {
+  display: inline-flex;
+  justify-content: space-around;
+  .pf-l-split__item {
+    position: relative;
+  }
+}
+
+.ins-c-rule-details__stack {
+  width: 600px;
```
karelhala, line 17: Couldn't this be achieved by `grid` system? I am a bit worried about small/large screens.

packages/advisor-components/src/RuleDetails/RuleDetails.scss
```diff
@@ -0,0 +1,91 @@
...
+    display: flex;
+    justify-content: space-between;
+    width: 100%;
+
+    .pf-c-content {
+      font-size: var(--pf-global--FontSize--sm);
+    }
+
+    .description-stack-override {
+      width: 452px;
```
karelhala, line 29: Same as above

packages/advisor-components/src/RuleDetails/RuleDetails.scss
```diff
@@ -0,0 +1,91 @@
...
+  .ins-c-rule-details__stackitem {
+    display: flex;
+    justify-content: space-between;
+    width: 100%;
+
+    .pf-c-content {
+      font-size: var(--pf-global--FontSize--sm);
+    }
+
+    .description-stack-override {
```
karelhala, line 28: Can we use `ins-c-*` prefix here?
```suggestion
    .ins-c-description-stack-override {
```

packages/advisor-components/src/RuleDetails/RuleDetails.scss
```diff
@@ -0,0 +1,91 @@
...
+      width: 452px;
+
+      .severity-line {
```
karelhala, line 31: Same as above with the prefix.

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
```diff
@@ -0,0 +1,247 @@
...
+import ContextWrapper, { ContextWrapperProps, ProductContext } from '../ContextWrapper';
+import messages from '../messages';
+import { barDividedList, topicLinks } from '../common';
+import { RiskOfChange } from '../types';
+import { Likelihood } from '../types';
+import { TopicRhel } from '../types';
+import { Rating } from '../types';
+import { RuleContentOcp } from '../types';
+import { RuleContentRhel } from '../types';
+import { AdvisorProduct } from '../types';
```
karelhala, line 28: These should be merged together
```suggestion
import { RiskOfChange, AdvisorProduct, Likelihood, TopicRhel, Rating, RuleContentOcp, RuleContentRhel } from '../types';
```

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
```diff
@@ -0,0 +1,247 @@
...
+      );
+
+    switch (product) {
+      case AdvisorProduct.ocp:
+        return ruleDescription(rule.generic, true);
+        break;
+      case AdvisorProduct.rhel:
+        return isDetailsPage ? ruleDescription(rule.generic, true) : ruleDescription(rule.summary);
+    }
+  };
```
karelhala, line 70: This is a simple `if` case. I am not a big fan of `switch` statements, can we use that instead?

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
```diff
@@ -0,0 +1,247 @@
...
+            <StackItem>
+              <Link key={`${rule.rule_id}-link`} onClick={() => onViewAffectedClick && onViewAffectedClick(rule.rule_id)} to="#">
+                {intl.formatMessage(messages.viewAffectedSystems, {
+                  systems: (rule as RuleContentRhel).impacted_systems_count,
+                })}
+              </Link>
+            </StackItem>
+          );
+        }
+    }
```
karelhala, line 99: Same as above, not a huge fan of switches. Any reason to use `switch`?

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
```diff
@@ -0,0 +1,247 @@
...
+  const renderViewAffected = () => {
+    switch (product) {
+      case AdvisorProduct.ocp:
+        if ((rule as RuleContentOcp).impacted_clusters_count > 0) {
+          return (
+            <StackItem>
+              <Link key={`${rule.rule_id}-link`} onClick={() => onViewAffectedClick && onViewAffectedClick(rule.rule_id)} to="#">
+                {intl.formatMessage(messages.viewAffectedClusters, {
+                  clusters: (rule as RuleContentOcp).impacted_clusters_count,
+                })}
```
karelhala, line 81: This should be passed from the parent component.

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+  return (
+    <Split className="ins-c-rule-details__split" hasGutter>
+      <SplitItem>
+        <Stack hasGutter>
+          {header && <StackItem>{header}</StackItem>}
+          <StackItem>{renderDescription()}</StackItem>
+          {(rule as RuleContentRhel).node_id && (
+            <StackItem>
+              <a rel="noopener noreferrer" target="_blank" href={`https://access.redhat.com/node/${(rule as RuleContentRhel).node_id}`}>
```
karelhala, line 142: I'd rather see the full `href` being passed from parent component. When the managers call for change of URL we'd have to change it here, release new version, use this new version and release the app itself.

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
```diff
@@ -0,0 +1,247 @@
...
+            </StackItem>
+          )}
+          {topics && rule.tags && topicLinks(rule as RuleContentRhel, topics).length > 0 && (
+            <StackItem>
+              <strong>{intl.formatMessage(messages.topicRelatedToRule)}</strong>
+              <br />
+              {barDividedList(topicLinks(rule as RuleContentRhel, topics))}
+            </StackItem>
+          )}
+          {isDetailsPage && onVoteClick && <RuleRating {...{ ruleId: rule.rule_id, ruleRating: rule.rating, onVoteClick }} />}
```
karelhala, line 155: No need to spread
```suggestion
          {isDetailsPage && onVoteClick && <RuleRating ruleId={ rule.rule_id} ruleRating={ rule.rating } onVoteClick={ onVoteClick } />}
```

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
```diff
@@ -0,0 +1,247 @@
...
+                        />
+                      </StackItem>
+                      <StackItem>{renderImpactSeverityLine()}</StackItem>
+                    </Stack>
+                  </Stack>
+                </span>
+              </StackItem>
+              {resolutionRisk && resolutionRiskDesc && (
+                <React.Fragment>
+                  <hr></hr>
```
karelhala, line 203: Hmmm, can this be achieved with styling?

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
```diff
@@ -0,0 +1,247 @@
...
+                    <strong>{intl.formatMessage(messages.riskOfChange)}</strong>
+                  </StackItem>
+                  <StackItem className={`pf-u-display-inline-flex alignCenterOverride pf-u-pb-sm pf-u-pt-sm`}>
+                    <span className="ins-c-rule-details__stackitem">
+                      <span>
+                        <InsightsLabel
+                          text={intl.formatMessage(messages[RISK_OF_CHANGE_LABEL_KEY[resolutionRisk as RiskOfChange]])}
+                          value={resolutionRisk}
+                          hideIcon
+                          rest={{}}
```
karelhala, line 214: Does passing empty object do anything?

packages/advisor-components/src/RuleRating/RuleRating.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const [submitted, setSubmitted] = useState(false);
+  const [thankYou, setThankYou] = useState(intl.formatMessage(messages.feedbackThankYou));
+
+  const updateRuleRating = async (newRating: Rating) => {
+    const calculatedRating = rating === newRating ? 0 : newRating;
+    try {
+      await onVoteClick(ruleId, calculatedRating);
+      setRating(calculatedRating);
+      setSubmitted(true);
+      setTimeout(() => setThankYou(''), 3000);
```
karelhala, line 32: Wouldn't it be better to use `debounce` here? So we won't send multiple requests?

packages/advisor-components/src/RuleRating/RuleRating.tsx
```diff
@@ -0,0 +1,52 @@
...
+    } catch (error) {
+      console.error(error); // eslint-disable-line no-console
+    }
+  };
+
+  return (
+    <span className="ratingSpanOverride">
+      {intl.formatMessage(messages.ruleHelpful)}
+      <Button variant="plain" aria-label="thumbs-up" onClick={() => updateRuleRating(1)} ouiaId="thumbsUp">
+        {rating === 1 ? <ThumbsUpIcon className="like" size="sm" /> : <OutlinedThumbsUpIcon size="sm" />}
```
karelhala, line 42: I am not a big fan of plain `like` and `dislike` classnames. Can we use something like `ins-c-like` and `ins-c-dislike`? Or something along these lines.

karelhala (commented):
Looking good! I'd like us to investigate usage of `intl` and `router`, looks like many of it can be taken out to parent component. Btw [packages/advisor-components/src/RuleRating/RuleRating.scss](https://github.com/RedHatInsights/frontend-components/pull/1493/files#diff-884045563f21beff6b1a77da6036fe3bd18eee780ba64577653b1056b8a34bce) is empty file, any reason behind this?

packages/advisor-components/src/ContextWrapper.tsx
gkarat: OK, I think it's simpler 👍🏼

packages/advisor-components/src/RebootRequired/RebootRequired.tsx
gkarat: Fixed 👍🏼

packages/advisor-components/src/RuleDetails/RuleDetails.scss
gkarat: Yeah, sure it can... the thing is, I was copy-pasting these CSS stylesheets from the previous old implementation and didn't dig into it yet... I'll revisit this part soon and push a commit

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
gkarat: bad habit :D thank you

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
gkarat: Well, the thing is, once a new product supporting Advisor will come with its special rule structure and so on, the switch will provide a better syntax to treat multiple values of AdvisorProduct. However, currently, this is not an issue, I think you point is legit.. reworked it in the previous commits

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
```diff
@@ -0,0 +1,247 @@
...
+    }
+  };
+
+  const renderViewAffected = () => {
+    switch (product) {
+      case AdvisorProduct.ocp:
+        if ((rule as RuleContentOcp).impacted_clusters_count > 0) {
+          return (
+            <StackItem>
+              <Link key={`${rule.rule_id}-link`} onClick={() => onViewAffectedClick && onViewAffectedClick(rule.rule_id)} to="#">
```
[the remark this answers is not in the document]
gkarat: Done in the last commit: replaced import from react-router-dom with requiring the Link component from parent 👍🏼

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
```diff
@@ -0,0 +1,229 @@
...
+  isDetailsPage: boolean;
+  topics: TopicRhel[];
+  /**
+   * onVoteClick - a callback used to update the rating of a particular rule
+   * @param {string} ruleId - ID (usually in plugin|error_key format) of the rule that needs to be updated
+   * @param {number} newRating rating (-1, 0, 1)
+   */
+  onVoteClick?: (ruleId: string, calculatedRating: Rating) => unknown;
+  showViewAffected?: boolean;
+  linkComponent?: any;
```
gkarat, line 38: Not sure which type to use here. When I tried `React.Component`, the compiler errored with
`error TS2604: JSX element type 'Link' does not have any construct or call signatures.`

The same error message was obtained after using `React.JSXElementConstructor<LinkProps>` as a type...

packages/advisor-components/src/RuleRating/RuleRating.tsx
gkarat: @karelhala, isn't only one network request made by calling `onVoteClick`? Why do you say there are multiple?

packages/advisor-components/src/RuleDetails/RuleDetails.scss
gkarat: Rewrote it, however, not by using a grid, but flex instead (this is for me a bit familiar, and the same result is achievable). 
There are still some bits to be tweaked (Actions button position in small devices), however, I would like to tackle it in a separate PR if you don't mind
 
image
image

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
gkarat: What do you mean exactly? This is already achieved by defining styles for the line in a SCSS file

packages/advisor-components/src/ContextWrapper.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export const ProductContext = React.createContext(AdvisorProduct.rhel);
+
+const ContextWrapper: React.FC<ContextWrapperProps> = ({ routerProps, env, children }) => {
+  const locale = navigator.language.slice(0, 2);
+
+  return routerProps ? (
+    <Router {...routerProps}>
+      <IntlProvider locale={locale} messages={translations[locale]}>
```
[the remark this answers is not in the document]
karelhala: @gkarat you should be able to pass strings from parent component, which is preferable. Your application knows which strings should be used where, if we were to pass messages from this repository to translators they wouldn't have full picture. Plus whenever you'd introduce new language we'd have to pass the request to translators as well. Hence would be better to pass the strings from your app, instead of having them in `messages` object. You can still parametrize those messages on the app side. All of the information is there, no need to bring the `IntlProvider` over.

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
karelhala: Well `hr` element just draws a line, it would be better if we could write the line in CSS instead of by element. The thing is HTML should be used to display the content and not to draw some lines, breaks and such, for this CSS should be used. I would understand it if there would have been a big wall of text afterall the w3c definition for `hr` is `thematic changes in the content`. Is this thematic change in the content?

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
karelhala: Mhh, might be a good to pass default value for the `rest` so we don't have to pass empty objects.

packages/advisor-components/src/RuleRating/RuleRating.tsx
karelhala: What if the user starts to click multiple times on the button? I would have guessed we'd show the thank you message every 3 seconds. With debounce we'd show the `thank you` after the final button press.

gkarat: That is a nice point, thank you. Reworked in https://github.com/RedHatInsights/frontend-components/pull/1493/commits/669206b9399b2bab1651533197aa94fba6ffbf13

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
gkarat: So you propose to replace it with smth like `span` and apply the same styles that were already made here https://github.com/RedHatInsights/frontend-components/pull/1493/files/f20483172ca3844413ddf2c58e6ac926127931a4#diff-89292165f3efa1dbd55471691761cd20dab9a9bfb5fab6eeb832e63e5d5a71caR3 ?

To your last question: I am rewriting the legacy code and design implementation. The line divides a group with two severity lines for issue likelihood and impact number and a group with the only severity line showing the resolution risk. Can't say how much these values are thematically different, but this was reflected in the old RHEL Advisor mock-ups. @doruskova probably can provide some input here :)

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
gkarat: done in https://github.com/RedHatInsights/frontend-components/pull/1493/commits/e9eddcf5f5ce4ad8d15b8129808cb563f8103ef2

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
gkarat: Checked the source code, the `rest` doesn't have a default value. The component just spreads it. If empty object,nothing is spread:
```
...
    <Label {...rest} {...props} className={classNames(className, `ins-c-label-${value}`)} icon={!hideIcon && VALUE_TO_STATE[value].icon}>
      {text || VALUE_TO_STATE[value].text}
    </Label>
```

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
doruskova: I don't have anything to add here.

karelhala: @gkarat yeah, span with table borders would be awesome! I understand it's a bit legacy code, but while we're in it let's polish it a bit.

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
karelhala: Yeah, that's the issue. I'll open a PR to fix that (let's keep the empty onject here for backwards compatibility)

packages/advisor-components/src/RuleDetails/RuleDetails.tsx
gkarat: Thank you, revisitied it in https://github.com/RedHatInsights/frontend-components/pull/1493/commits/3679ee927d574ef32c0f188682f8c7ab2819dfec

packages/advisor-components/src/common.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export const topicLinks = (rule: RuleContentRhel, topics: TopicRhel[]) =>
+  topics
+    ? compact(
+        topics.map(
+          (topic) =>
+            intersection(topic.tag.split(' '), rule.tags.split(' ')).length && (
+              <React.Fragment key={topic.slug}>
+                <Link to={`/topics/${topic.slug}`}>{`${topic.name}`}</Link>
+              </React.Fragment>
```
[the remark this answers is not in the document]
gkarat: Nope; fixed :)

gkarat:
@karelhala, I have revisited intl and stuff in the last commit https://github.com/RedHatInsights/frontend-components/pull/1493/commits/690295bd0cb3d3bc33eb3708dbef51d45760456c. PTAL :-)

I know this https://github.com/RedHatInsights/frontend-components/commit/690295bd0cb3d3bc33eb3708dbef51d45760456c#diff-515c50375d5c4d4620f4b523db96890465236664f7e3f2b9192e29c64f586cf0R19 looks a bit extra thing. However, it could help parent apps learn which message IDs are expected (e.g., it imports the array of IDs and maps them to formatted messages, and then finally pass it to RuleDetails via 'messages'). WDYT? Don't you have better ideas?

packages/advisor-components/src/RuleRating/RuleRating.tsx
```diff
@@ -0,0 +1,52 @@
+import './RuleRating.scss';
+
+import React, { useState } from 'react';
+import { useIntl } from 'react-intl';
+
+import { Button } from '@patternfly/react-core/dist/js/components/Button/Button';
+import OutlinedThumbsDownIcon from '@patternfly/react-icons/dist/js/icons/outlined-thumbs-down-icon';
+import OutlinedThumbsUpIcon from '@patternfly/react-icons/dist/js/icons/outlined-thumbs-up-icon';
+import ThumbsDownIcon from '@patternfly/react-icons/dist/js/icons/thumbs-down-icon';
+import ThumbsUpIcon from '@patternfly/react-icons/dist/js/icons/thumbs-up-icon';
```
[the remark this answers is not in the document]
gkarat: You are true, https://github.com/RedHatInsights/frontend-components/blob/master/babel.config.js#L45 rewrites the paths to dist imports already => no reason to use dist imports in the code 👍🏼  reworked

packages/advisor-components/src/common.tsx
```diff
@@ -0,0 +1,26 @@
...
+import intersection from 'lodash/intersection';
+
+import { RuleContentRhel, TopicRhel } from './types';
+
+// eslint-disable-next-line no-undef
+export const barDividedList = (list: JSX.Element[]) =>
+  list.map((element, index) => (
+    <React.Fragment key={index}>
+      {element}
+      {index + 1 !== list.length && <strong className="verticalDivider">&nbsp;&#124;&nbsp;</strong>}
```
karelhala, line 12: I'd rather wrap the `{element}` in `span` and apply some styling to show table border on right.

karelhala (commented):
Looking good! Just a few minor enhancements and we are golden.

packages/advisor-components/src/common.tsx
gkarat: @karelhala, isn't it complicating things a bit? :) I mean, the result will be the same, however, the current &#124 vertical pipe already serves the punctuation function (comma). Writing CSS to repeat the Unicode character seems to me an extra thing :)

karelhala: If we are treating the pipe as comma I am fine with it.

packages/advisor-components/src/RuleDescription/RuleDescription.scss
```diff
@@ -0,0 +1,6 @@
+.genericOverride {
```
karelhala, line 1: I would rename this
```suggestion
.ins-c-generic__override {
```

packages/advisor-components/src/RuleDescription/RuleDescription.tsx
```diff
@@ -0,0 +1,35 @@
...
+  product: AdvisorProduct;
+  rule: RuleContentOcp | RuleContentRhel;
+  isDetailsPage: boolean;
+}
+
+const RuleDescription: React.FC<RuleDescriptionProps> = ({ product, rule, isDetailsPage }) => {
+  const ruleDescription = (data: string, isGeneric = false) => (
+    <React.Fragment>
+      {data && (
+        <span className={isGeneric ? 'genericOverride' : ''}>
```
karelhala, line 20:
```suggestion
        <span className={isGeneric ? 'ins-c-generic__override' : ''}>
```

karelhala (commented):
Just one final adjustment of clas and we are done!

packages/advisor-components/src/RuleDescription/RuleDescription.tsx
```diff
@@ -0,0 +1,35 @@
...
+    </React.Fragment>
+  );
+
+  if (product === AdvisorProduct.ocp) {
+    return ruleDescription(rule.generic, true);
+  } else {
+    // RHEL rule
+    return isDetailsPage ? ruleDescription(rule.generic, true) : ruleDescription(rule.summary);
+  }
+};
```
karelhala, line 33: We can further improve this by returning directly the component, you can even remove the `return` and return this directly.

```suggestion
const RuleDescription: React.FC<RuleDescriptionProps> = ({ product, rule, isDetailsPage }) => (
    <React.Fragment>
      {(rule.generic || rule.summary) && (
        <span className={(product === AdvisorProduct.ocp || isDetailsPage) ? 'ins-c-generic__override' : ''}>
          <Markdown rehypePlugins={[rehypeRaw, rehypeSanitize]}>
            {(product === AdvisorProduct.ocp || isDetailsPage) ? rule.generic : rule.summary }
          </Markdown>
        </span>
      )}
    </React.Fragment>
  );
```

karelhala (commented):
Looking good! Just one final code cleanup (nothing problematic, just cosmetic change)

karelhala (approved):

Hyperkid123:
@karelhala will you deploy the initial version of the package?

karelhala:
@Hyperkid123 yup, can do. Let's merge and release

karelhala:
@gkarat package has been released. Can you please test it in your app? https://www.npmjs.com/package/@redhat-cloud-services/frontend-components-advisor-components

gkarat:
> @gkarat package has been released. Can you please test it in your app? https://www.npmjs.com/package/@redhat-cloud-services/frontend-components-advisor-components

Yeah, this is great, thank you for helping with it! And yes, I am about to import this component in OCP Advisor, so if there are any issues, I'll let you know 👍🏼

PR #1494, opened 2022-04-01
Bounce PROD requests when not using Red Hat VPN

rvsia:
- [x] todo: `npm install node-fetch@2`

All non-GET PROD requests are bounced on server to target, so Akamai won't be messed up because of different headers

@Hyperkid123 releasing as a minor version because it adds a new behavior

### DIFF
--- a/packages/config-utils/proxy.js
+++ b/packages/config-utils/proxy.js
@@ -53,8 +53,8 @@ module.exports = ({
 
   let agent;
 
-  if (env.startsWith('prod') || env.startsWith('stage')) {
-    // PROD and stage are deployed with Akamai which requires a corporate proxy
+  if (env.startsWith('stage')) {
+    // stage is deployed with Akamai which requires a corporate proxy
     agent = new HttpsProxyAgent(proxyURL);
   }
 
--- a/packages/config-utils/proxy.js
+++ b/packages/config-utils/proxy.js
@@ -36,8 +36,8 @@ module.exports = ({
   registry = [],
   isChrome = false,
   onBeforeSetupMiddleware = () => {},
-  bounceProd = true,
-  useAgent,
+  bounceProd = false,
+  useAgent = true,
 }) => {
   const proxy = [];
   const majorEnv = env.split('-')[0];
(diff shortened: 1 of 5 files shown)


karelhala (approved):

mkholjuraev:
@rvsia Thanks for the PR. I have run it locally and was able to do all types of requests in prod environment without being behind the VPN.

packages/config-utils/proxy.js
```diff
@@ -53,8 +57,10 @@ module.exports = ({
 
   let agent;
 
-  if (env.startsWith('prod') || env.startsWith('stage')) {
-    // PROD and stage are deployed with Akamai which requires a corporate proxy
+  const isProd = env.startsWith('prod');
+
+  if (env.startsWith('stage') || useAgent) {
+    // stage is deployed with Akamai which requires a corporate proxy
```
Hyperkid123, line 63: We need the proxy agent in prod as well. The `useAgent` should be `true` by default.

rvsia: I do not think so because now PROD works without the agent thanks to the bouncing. So by default it will work for anyone un/connected to VPN. But this option is made for people who are connected to the VPN and for whatever reason they do not want to use the bouncing. (There could and probably there would be some hidden bugs.)

rvsia: Yeah, Patch team wanted to develop on prod without the vpn. So I tried to fix that... and I came up with the bounce mechanism included in this PR. Check the bypass function.

Hyperkid123: Can we first hide it entirely behind a flag? So we can test a bit before updating every config? I would like to avoid breaking any other local development setup.

I have a few services I can test it that I know do not work in prod env. (One of which is chrome)

rvsia: Sure, I can switch the default variables to true<>false, I just thought the default behavior without VPN makes more sense.

rvsia:
@Hyperkid123 Updated so the new behavior is optional

Hyperkid123 (approved):

PR #1498, opened 2022-04-04
fix(skeletonTable): make columns optional

rvsia:
- since typescript conversion, there is a default value and it broke several applications

cc @kruai


Hyperkid123:
@rvsia can you update the sanpshots?

rvsia:
@Hyperkid123 done and green 🍏

karelhala (approved):

PR #1500, opened 2022-04-05
Add define plugin to replace CRC_APP_NAME with insights.appname

karelhala:


packages/config/src/plugins.js
```diff
@@ -53,6 +54,11 @@ module.exports = ({
           },
           ...(replacePlugin || []),
         ]),
+        new DefinePlugin({
+          // we have to wrap the appname string in another string because of how define plugin explodes strings
+          CRC_APP_NAME: JSON.stringify(insights?.appname),
```
Hyperkid123, line 59: We might also want to add a new global variable for the `CRC_APP_NAME` variable to the eslint config.

packages/config/src/plugins.test.js
```diff
@@ -52,3 +53,20 @@ it('replacePlugin should update', () => {
   });
   enabledPlugins[REPLACE].replace({ html: '@@another string @@env' }, (_, { html }) => expect(html).toBe('test-string string '));
 });
+
+it('definePlugin should have default replace of CRC_APP_NAME', () => {
+  const enabledPlugins = plugins({
+    insights: { appname: 'test_app' },
+  });
+  console.log(enabledPlugins[DEFINE_PLUGIN]);
```
Hyperkid123, line 61: console.log

Hyperkid123 (approved):

Issue #1514, opened 2022-04-19
Insights Label Critical Icon update


Fewwy:
The Insights Label component uses Critical Icon.
frontend-components/packages/components/src/InsightsLabel/CriticalIcon.tsx

The PatternFly library updated the icon and it should be updated in the frontend component library.

Label in a context: https://marvelapp.com/prototype/g70a1he/screen/84434999
PatternFly PR: https://github.com/patternfly/patternfly/pull/4758

Expected: the icon is updated to the new one.

PR #1516, opened 2022-04-19
Add new CrtiticalRiskIcon component to the InsightsLabel

Fewwy:
Removed the code of the previous icon and tests for it
Icon after my changes
image


Hyperkid123 (dismissed):

package.json
```diff
@@ -59,7 +59,7 @@
         "@octokit/rest": "^16.34.0",
         "@patternfly/quickstarts": "^1.4.1-rc.1",
         "@patternfly/react-core": "^4.147.0",
-        "@patternfly/react-icons": "^4.10.2",
+        "@patternfly/react-icons": "^4.53.16",
```
Hyperkid123, line 62: Can you update the peer dependency of the components package? Consumers with older PF icons version will not have the icon available and it can cause issues.

Fewwy: @Hyperkid123 Done, please take a look

Hyperkid123: Please run `npm install` in the repository root. The lockfile is still not updated.

Fewwy: @Hyperkid123 ran npm install and pushed the update

packages/components/package.json
```diff
@@ -34,7 +34,7 @@
     "peerDependencies": {
         "@patternfly/react-table": ">=4.5.7",
         "@patternfly/react-core": ">=4.18.5",
-        "@patternfly/react-icons": ">=4.3.5",
+        "@patternfly/react-icons": "4.53.16",
```
Hyperkid123, line 37: This has to use the `>=` operator. The current solution would mean that you **have to use 4.53.16**. Which is no good. We want **at least 4.53.16* or higer*

Hyperkid123 (changes requested):

Fewwy:
@Hyperkid123 Updated the peerdependancy based on your requirements

Hyperkid123:
closes #1514

PR #1518, opened 2022-04-19
Advisor components: fix the issue with rendering of view affected button

gkarat:
Replaces nullish coalescing operator with the AND operator (&&) and fixes the issue when rule content had impacted_clusters_count or impacted_systems_count defined, however, the button "View affected {clusters|systems}" was rendered.

Improves #1493


gkarat:
@karelhala, @Hyperkid123, PTAL. Should I increase the npm version manually? Or there is a bot that will do it?

Hyperkid123:
> @karelhala, @Hyperkid123, PTAL. Should I increase the npm version manually? Or there is a bot that will do it?

Only the first release has to be done manually

Hyperkid123 (approved):

PR #1528, opened 2022-04-26
Fixing insights label styles

Fewwy:
The labels had gray borders that were incorrect (the PatternFly styling guide says that they need to be colored)
https://www.patternfly.org/v4/components/label/html
I removed unnecessary code that was adding custom classes and used the native PatternFly class to fix the CSS issue.
I left the CSS code for 1 class that covers the "gold" color because it is not provided by PatternFly labels color selection.
image


Fewwy:
@gkarat @Hyperkid123 PTAL

packages/components/src/InsightsLabel/InsightsLabel.tsx
```diff
@@ -34,7 +33,14 @@ export interface InsightsLabelProps extends LabelProps {
 
 const InsightsLabel: React.FunctionComponent<InsightsLabelProps> = ({ value = 1, text, hideIcon, className, rest, ...props }) => {
   return (
-    <Label {...rest} {...props} className={classNames(className, `ins-c-label-${value}`)} icon={!hideIcon && VALUE_TO_STATE[value].icon}>
+    <Label
+      {...rest}
+      {...props}
+      className={
+        value === 4 ? 'pf-c-label pf-m-red' : value === 3 ? 'pf-c-label pf-m-orange' : value === 1 ? 'pf-c-label pf-m-blue' : 'ins-c-label-2'
```
gkarat, line 40: `pf-c-label` can be omitted here I believe.

Also, consider using an object to map total risk value to the appropriate class name like:
```
{ 1: 'pf-m-blue', 2: 'ins-c-label-2', ... }
```

packages/components/src/InsightsLabel/labels.scss
```diff
@@ -42,13 +13,6 @@
...
-    background-color: var(--ins-c-label--blue-bg);
-    .pf-c-label__content {
-        color: var(--inc-c-label--content-blue);
-        .pf-c-label__icon {
-            color: var(--inc-c-label--icon-blue);
-        }
-    }
-}
+.ins-c-label-2.pf-c-label {.pf-c-label__content::before {
+    border: 1px solid var(--inc-c-label--icon-gold);
```
gkarat, line 17: I have a feeling that the colors from https://marvelapp.com/prototype/g70a1he/screen/84434999 are different. Could you please double check it?

packages/components/src/InsightsLabel/InsightsLabel.tsx
rvsia, line 40: Could we add that to `VALUE_TO_STATE` to remove these ternary operators?

packages/components/src/InsightsLabel/labels.scss
Fewwy: Good catch. I asked Marie to give me the exact # number of the color.

packages/components/src/InsightsLabel/InsightsLabel.tsx
Fewwy: @rvsia Good suggestion. I will do that

packages/components/src/InsightsLabel/labels.scss
Fewwy: Updated the colors according to the Marie's mockup

rvsia (approved):

packages/components/src/InsightsLabel/InsightsLabel.tsx
```diff
@@ -6,19 +6,19 @@ import { CriticalRiskIcon } from '@patternfly/react-icons';
...
     text: string;
+    class: string;
   };
 };
 const VALUE_TO_STATE: InsightsLabelValueMapping = {
-  1: { icon: <AngleDoubleDownIcon />, text: 'Low' },
-  2: { icon: <EqualsIcon />, text: 'Moderate' },
-  3: { icon: <AngleDoubleUpIcon />, text: 'Important' },
-  4: { icon: <CriticalRiskIcon />, text: 'Critical' },
+  1: { icon: <AngleDoubleDownIcon />, text: 'Low', class: 'pf-c-label pf-m-blue' },
```
gkarat, line 18: No reason to add `pf-c-label` here. The `Label` component already adds it while rendered.

packages/components/src/InsightsLabel/labels.scss
```diff
@@ -1,36 +1,8 @@
...
-
-.ins-c-label-3.pf-c-label {
-    background-color: var(--ins-c-label--orange-bg);
-    .pf-c-label__content {
-        color: var(--inc-c-label--content-orange);
-        .pf-c-label__icon {
-            color: var(--inc-c-label--icon-orange);
-        }
-    }
+    --inc-c-label--border-gold:         #F9E0A2;
```
gkarat, line 5: Nit-pick: you can refer to `--pf-global--palette--gold-100` which has the same value: https://www.patternfly.org/v4/guidelines/colors/#color-palette

gkarat (approved):
LGTM. A couple of things can be even more improved, but it's not crucial

packages/components/src/InsightsLabel/InsightsLabel.tsx
```diff
@@ -6,19 +6,19 @@ import { CriticalRiskIcon } from '@patternfly/react-icons';
...
+  2: { icon: <EqualsIcon />, text: 'Moderate', class: 'ins-c-label-2' },
+  3: { icon: <AngleDoubleUpIcon />, text: 'Important', class: 'pf-c-label pf-m-orange' },
+  4: { icon: <CriticalRiskIcon />, text: 'Critical', class: 'pf-c-label pf-m-red' },
```
Hyperkid123, line 21: Do now use class. Use the LabelComponent props API instead. For the colors it should be the `color` prop. https://www.patternfly.org/v4/components/label-group/#label.

packages/components/src/InsightsLabel/InsightsLabel.tsx
```diff
@@ -34,7 +34,7 @@ export interface InsightsLabelProps extends LabelProps {
 
 const InsightsLabel: React.FunctionComponent<InsightsLabelProps> = ({ value = 1, text, hideIcon, className, rest, ...props }) => {
   return (
-    <Label {...rest} {...props} className={classNames(className, `ins-c-label-${value}`)} icon={!hideIcon && VALUE_TO_STATE[value].icon}>
+    <Label {...rest} {...props} className={VALUE_TO_STATE[value].class} icon={!hideIcon && VALUE_TO_STATE[value].icon}>
```
Hyperkid123, line 37: ^^

packages/components/src/InsightsLabel/InsightsLabel.tsx
Fewwy: Ok, sure

packages/components/src/InsightsLabel/InsightsLabel.tsx
Fewwy: Yep, you are correct

packages/components/src/InsightsLabel/labels.scss
Fewwy: Yep, changed to the var

Hyperkid123 (approved):

PR #1540, opened 2022-05-03
advisor-components: Cypress component testing support

gkarat:
This adds cypress component testing[^1] support to the `advisor-components` package.

- The framework is already utilized in the ocp-advisor application repo[^2]. Since there are some presentational components that we would like to move to this library, we need to make sure we have the same tooling available here in order to move the existing tests or write the new ones.
- The framework makes it easy to visualize particular components and run tests on them (a mix of storybook and testing tool). On average, the tests run are quite fast and can serve well as a way more effective replacement for some integration tests written using the iqe framework (used for integration tests across RH products).
- Usage of this tool here encourages QE folks to also contribute here for the reason from the previous point.
- This complements existing tooling with jest. Right now, I tried to implement this only for the advisor-components repository as a proof of concept.

image

- [x] Set up initial cypress configs for the package
- [x] Add first basic tests
- [x] Update docs

[^1]: https://docs.cypress.io/guides/component-testing/introduction
[^2]: https://github.com/RedHatInsights/ocp-advisor-frontend


package.json
```diff
@@ -116,7 +116,8 @@
         "start": "lerna run --parallel start --ignore=docs",
         "build": "lerna run build --stream --no-private --ignore=@redhat-cloud-services/frontend-components-config*",
         "generate-translations": "npm run build && node packages/utils/src/mergeMessages.js",
-        "test": "jest",
+        "test:ct": "npx -w ./packages/advisor-components cypress run-ct",
```
gkarat, line 119: This will be changed in the future once other packages will start utilizing cypress-ct. For now, `-w` lets to run cypress tests in the advisor package.

Hyperkid123: We don't want to drop jest. We will have to have two scripts. One for jest and one for cypress

gkarat: @Hyperkid123, I didn't propose to drop it, we will also use it in advisor-components. PTAL https://github.com/RedHatInsights/frontend-components/pull/1540/files/ce8082bc404e8e9027a2c8c52c783ee2de3c537a#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R120: I am just adding one more command on top of `jest` so that we can include cypress tests in `npm run test` command

gkarat: I will add more info soon with a brief overview of pros/cons/reasons, it is still work in progress

Hyperkid123: Sorry I just looked at your comment and it looks like the te command is gone :face_exhaling: I was too lazy to look at the whole file :+1:

gkarat:
@Hyperkid123, @karelhala, I have a feeling that the build is failing because there is a platform code-freeze or something: the tests are all passing, but the `1` code I think is a result of uploading cache/build https://app.travis-ci.com/github/RedHatInsights/frontend-components/builds/250224709#L2046

packages/advisor-components/src/RuleDetails/RuleDetails.spec.ct.js
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
```
gkarat, line 1: This is a single spec I am adding within this patch. Serves more like an illustration of typical cypress specs and tests RuleDetails component.

Fewwy (approved):
LGTM!

Hyperkid123:
@gkarat can you look at the failing CI? The lint job is failing for your cypress tests. We could ignore the CT tests from the linter but I would prefer if we first try to fix it.

gkarat:
> @gkarat can you look at the failing CI? The lint job is failing for your cypress tests. We could ignore the CT tests from the linter but I would prefer if we first try to fix it.

oh, thanks, didn't notice that it was a lint issue

packages/advisor-components/package.json
```diff
@@ -29,8 +29,8 @@
         "@patternfly/react-core": ">=4.162.16",
         "@patternfly/react-icons": ">=4.3.5",
         "prop-types": ">=15.6.2",
-        "react": ">=16.14.0 || >=17.0.0",
-        "react-dom": ">=16.14.0 || >=17.0.0",
+        "react": "^16.14.0",
```
Hyperkid123, line 32: Is there any reason you do not include the v17 here? If it's unable to parse the version use 17 instead. Even that is no longer lts.

gkarat: @Hyperkid123, TBH, I don't remember the exact problem: I couldn't install all the required dependencies. However, now I tried to revert this change, then run `npm run clean && npm i`, and it looks like it builds fine. So I changed the versions back at https://github.com/RedHatInsights/frontend-components/pull/1540/commits/85aa3b3b64ddde6a6b17bb8da16fdc2464eef308

PR #1541, opened 2022-05-04
add download button documentation

InsaneZein:
adds documentation about the download button


Hyperkid123 (changes requested):
We now add docs to the documentation package. These docs are deprecated. You can follow this guide:

setup: http://front-end-docs-insights.apps.ocp4.prod.psi.redhat.com/contributing

generated docs: http://front-end-docs-insights.apps.ocp4.prod.psi.redhat.com/contributing/frontend-components

You should able to re-use most of the text

InsaneZein:
> We now add docs to the documentation package. These docs are deprecated. You can follow this guide:


I've added docs to the right place now.

Hyperkid123 (commented):
@InsaneZein ok the docs and example are working.

I have two things.

1. can you add sone description to each of the prop?
2. Can you define what is the `extraItems?: any[];` prop type should be? The `any[]` is not helpful in the docs.

InsaneZein:
> @InsaneZein ok the docs and example are working.
> 
> I have two things.
> 
> 1. Can you add some description to each of the prop?
> 2. Can you define what is the `extraItems?: any[];` prop type should be? The `any[]` is not helpful in the docs.

I've added descriptions to each prop except extraItems. Still trying to figure out what that one is for

Hyperkid123 (approved):

PR #1543, opened 2022-05-09
Add disableDotRule flag set to true in the webpack config

gkarat:
Fixes https://github.com/RedHatInsights/ocp-advisor-frontend/issues/29

This allows to use dots in URL path names and fixes the "Cannot GET" error when building any of the applications (that consume this config) locally.

image


Hyperkid123 (approved):

Hyperkid123:
@gkarat looks good. Can you just add the missing attribute to the test file? Thanks

PR #1547, opened 2022-05-17
Remove Dropdown props from DownloadButton that are never used

josejulio:
These are overriden by DownloadButton and are never used - no point in accepting them.


Hyperkid123 (commented):
Rather than removing it, can we just pass it to the component? In case anyone wants to change the variant?

josejulio:
`DownloadButton` component manually specifies a `toggle`, `dropdownItems` and `isPlain`. 

Which ones would you like to pass trough if they are specified? All or only `isPlain` ?

Hyperkid123:
@josejulio only the `isPlain`. the rest depends on internal variables.

josejulio:
@Hyperkid123  done

packages/components/src/DownloadButton/DownloadButton.tsx
```diff
@@ -48,7 +48,7 @@ const DownloadButton: React.FunctionComponent<DownloadButtonProps> = ({
       {conditionallyTooltip(
         <Dropdown
           {...props}
-          isPlain
+          isPlain={props.isPlain ?? true}
```
rvsia, line 51: We can move `isPlain` before `{...props}` so consumers can freely overwrite it.

rvsia:
```jsx
<Dropdown 
  isPlain
  {...props}
  ...
  />
```

PR #1558, opened 2022-06-13
Configure proxy to download development version of chrome.

Hyperkid123:
part of: https://issues.redhat.com/browse/RHCLOUD-19837

Downloads chrome from `dev-beta` or `dev-stable` branches to enable React development environment.

### TODO
- [x] wait for dev-stable to be published

### DIFF
--- a/packages/config-utils/proxy.js
+++ b/packages/config-utils/proxy.js
@@ -38,6 +38,7 @@ module.exports = ({
   onBeforeSetupMiddleware = () => {},
   bounceProd = false,
   useAgent = true,
+  useDevBuild = true,
 }) => {
   const proxy = [];
   const majorEnv = env.split('-')[0];
@@ -284,8 +285,9 @@ module.exports = ({
             defaultServices.chrome = defaultServices.chrome({});
           }
 
+          const chromeEnv = useDevBuild ? (env.includes('-beta') ? 'dev-beta' : 'dev-stable') : env;
           chromePath = checkoutRepo({
-            repo: `${defaultServices.chrome.path}#${env}`,
+            repo: `${defaultServices.chrome.path}#${chromeEnv}`,
             reposDir,
             overwrite: true,
           });
--- a/packages/docs/pages/proxies/webpack-proxy.mdx
+++ b/packages/docs/pages/proxies/webpack-proxy.mdx
@@ -10,6 +10,7 @@
   - [useProxy](#useproxy)
     - [Attributes](#attributes)
       - [useChromeTemplate](#useChromeTemplate)
+      - [useDevBuild](#useDevBuild)
       - [localChrome](#localchrome)
       - [registry](#registry)
       - [Custom routes](#custom-routes)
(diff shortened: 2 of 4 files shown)


packages/docs/pages/proxies/webpack-proxy.mdx
```diff
@@ -86,11 +88,33 @@ const { config: webpackConfig, plugins } = config({
 });
 ```
 
+#### useDevBuild
+
+*boolean*
+
+Defaults to **true**. Flag to disable loading of development version of chrome. Has effect only in production environment.
```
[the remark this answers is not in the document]
Hyperkid123: Only in a local environment. So if you want to run webpack proxy on prod.foo... local and want production assets you set this to false.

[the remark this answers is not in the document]
himdel: Ooh, thanks, I was wondering if it was something in the wording or in my understanding, glad to know it was the latter :)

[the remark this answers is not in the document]
Hyperkid123: Yeah, this config is only for the webpack-proxy. The build is not affected at all.

packages/config-utils/proxy.js
```diff
@@ -38,6 +38,7 @@ module.exports = ({
   onBeforeSetupMiddleware = () => {},
   bounceProd = false,
   useAgent = true,
+  useDevBuild = true,
```
karelhala, line 41: Shouldn't this be default as `false`? So when running on stage.foo you'd get dev React by default (no way to opt-out of this) and by default prod.foo will give you the prod React unless you specify `useDevBuild: true`?

Hyperkid123: I would argue that for development you want a development bundle regardless of the environment. Plus folks that use prod env for development (there are a few of them) never actually test pre-prod chrome and are always surprised when something changes.

karelhala: The problem I have with this approach is that it could use outdated version of prod environent.

rvsia (approved):

packages/config-utils/proxy.js
bastilian: Would it maybe be possible to show an indicator in the UI (only visible in dev) which version, maybe even commit of chrome is currently used?

Hyperkid123: @bastilian  That is something we can do yeah. But that would have to be done in chrome. 

@karelhala it will always use the latest stable/beta version.

packages/docs/pages/proxies/webpack-proxy.mdx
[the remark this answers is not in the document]
Hyperkid123: The code snippet is a bit misleading. the section in which it is dedicated to the webpack proxy. it should be fine if you see the whole section.

karelhala (approved):

PR #1572, opened 2022-06-27
fix: Block hook from updating unmounted component

bastilian:
On occasion (mildly put) the timing is off and the RBAC hook tries to update a component that is unmounted. 
Screenshot 2022-06-29 at 10 41 18

This change will prevent that from happening.


mkholjuraev (approved):
LGTM"

PR #1576, opened 2022-06-30
feat(RBAC): Extend RBAC hook with context

bastilian:
This adds a separate hook for RBAC using a context, to use with a provided `RBACProvider`.
This change is a follow up to https://github.com/RedHatInsights/frontend-components/pull/1572.

Using the hook with the context in inventory[0] has improved the loading behaviour of components using RBAC and appears to have eliminated the "update on unmounted component" issue. 

[0] https://github.com/RedHatInsights/insights-inventory-frontend/pull/1603


packages/components/src/RBACProvider/RBACProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+  };
+};
+
+export interface RBACProviderProps {
+  appName: string;
+}
+
+export const RBACProvider: React.FunctionComponent<RBACProviderProps> = ({ appName, children }) => {
+  const [permissionState, setPermissionState] = useState(initialPermissions);
+  let loading = false;
```
Hyperkid123, line 25: You probably want to store this value in a `useRef`. The value can be potentially re-declared during a render and you can be dealing with a race condition.

packages/utils/src/RBAC/RBAC.ts
```diff
@@ -60,4 +61,20 @@ export function hasAllPermissions(userPermissions: (Access | string)[], permissi
   });
 }
 
+export const initialPermissions = {
+  isLoading: true,
+  isOrgAdmin: false,
+  permissions: [],
+  hasAccess: (_?: any) => {
```
Hyperkid123, line 68: the `hasAccess` argument should have the actual types defined `(Access | string)[]`. Based on the values used at runtime.

packages/components/src/RBACProvider/RBACProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+import {
+  RBACContext,
+  doesHavePermissions,
+  getRBAC,
+  hasAllPermissions,
+  initialPermissions,
+} from '@redhat-cloud-services/frontend-components-utilities/RBAC';
+
+const hasAccessWithUserPermissions = (userPermissions: Access[]) => {
```
Hyperkid123, line 13: The type should be `(Access | string)[]`

packages/components/src/RBACProvider/RBACProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return (requiredPermissions: any, checkAll = true) => {
```
Hyperkid123, line 14: The required permissions type should be `string[]`

packages/utils/src/RBACHook/RBACHook.ts
```diff
@@ -35,4 +30,13 @@ export function usePermissions(appName: string, permissionsList: string[], disab
   return permissions;
 }
 
+export const usePermissionsWithContext = (requiredPermissions: string[]) => {
```
Hyperkid123, line 33: Type does not match

Hyperkid123 (commented):
What is the purpose of this empty file?
```
packages/utils/src/RBACHook/constants.ts
```

packages/components/src/RBACProvider/RBACProvider.tsx
bastilian: Changed to using ref.

packages/utils/src/RBAC/RBAC.ts
bastilian: Update.

packages/components/src/RBACProvider/RBACProvider.tsx
bastilian: Updated.

packages/components/src/RBACProvider/RBACProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return (requiredPermissions: string[], checkAll = true): boolean => {
+    return checkAll ? hasAllPermissions(userPermissions, requiredPermissions) : doesHavePermissions(userPermissions, requiredPermissions);
+  };
+};
+
+export interface RBACProviderProps {
+  appName: string;
+}
+
+export const RBACProvider: React.FunctionComponent<RBACProviderProps> = ({ appName, children }) => {
```
karelhala, line 23: Can we also have optional loader component? Which would defaults to
```
<Bullseye>
  <Spinner size="xl" />
</Bullseye>
```

packages/components/src/RBACProvider/RBACProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const loading = useRef(false);
```
karelhala, line 24: Any reason for using ref instead of state? The ref change will not trigger a re-render, meaning when app initiates a re-fetch (when for instance `appName` changes) the UI will not show spinner.

packages/components/src/RBACProvider/RBACProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+      ...currentPerms,
+      isLoading: false,
+      isOrgAdmin,
+      permissions: userPermissions,
+    }));
+    loading.current = false;
+  };
+
+  useEffect(() => {
+    if (!loading.current && permissionState.permissions.length == 0) {
```
karelhala, line 41: I don't think this is a correct condition, what if app requests another permissions for different `appName`, but the permissions will already be fetched? I'd change it to check if `appName` is defined (it will be undefined on initial render)

packages/utils/src/RBAC/RBAC.ts
```diff
@@ -60,4 +61,20 @@ export function hasAllPermissions(userPermissions: (Access | string)[], permissi
   });
 }
 
+export const initialPermissions = {
+  isLoading: true,
+  isOrgAdmin: false,
+  permissions: [],
+  hasAccess: (_?: (Access | string)[]) => {
+    return;
+  },
```
karelhala, line 70: Extra function body
```suggestion
  hasAccess: (_?: (Access | string)[]) => undefined,
```

karelhala (commented):
Looking good, just a couple of minor changes

packages/components/src/RBACProvider/RBACProvider.tsx
bastilian: Will add that, yes!

packages/components/src/RBACProvider/RBACProvider.tsx
bastilian: I think we actually don't need this anymore. This was to prevent the `state update on unmounted component error`. I'll try to actually remove it and see how it behaves without it.

packages/components/src/RBACProvider/RBACProvider.tsx
```diff
@@ -10,7 +9,7 @@ import {
   initialPermissions,
 } from '@redhat-cloud-services/frontend-components-utilities/RBAC';
 
-const hasAccessWithUserPermissions = (userPermissions: (Access | string)[]) => {
+const hasAccessWithUserPermissions = (userPermissions: string[]) => {
```
Hyperkid123, line 12: Bot of the `hasAllPermissions` and `doesHavePermissions` accept the `Access[]` type as well as `string[]` for the first argument. Any reason why we don't want it here?

```TS
export function hasAllPermissions(userPermissions: (Access | string)[], permissionList: string[]): boolean {

export function doesHavePermissions(userPermissions: (Access | string)[], permissionList: string[]): boolean {

```

bastilian: Thanks for the note! I'll update that as well. I'm just not as confident with  my TS to see these things right away, yet. :)

Hyperkid123: If you are re-using anying that is already written in TS, the best bet is just to re-use the same TS definition for arguments.

Hyperkid123 (approved):

PR #1593, opened 2022-08-03
fix(docs): Generate docs more memory efficient

bastilian:
When running `npm run generate:components` the tasks could require up to 3GB+ (and probably more if it would not get stopped by heap limits). This makes it a bit more efficient. Memory usage still increases (and gets GCed), but stays manageable.


Hyperkid123:
I am taking a look now :eyes:

Hyperkid123 (approved):

Hyperkid123:
Thanks for the upgrades!

PR #1594, opened 2022-08-05
addes a shared component for no registered systems in apps

mkholjuraev:
This introduces a new UI design for cases when there is not any system registered in an app. This new design is almost similar to the existing NotConnected component.


Hyperkid123:
@mkholjuraev can you please add the missing snapshot to fix the ci?

Hyperkid123 (approved):

PR #1610, opened 2022-09-12
Change CF label Node to ReactNode

fhlavac:
Changed the property type to `ReactNode`


packages/components/src/ConditionalFilter/ConditionalFilter.tsx
```diff
@@ -37,7 +37,7 @@ export interface TextInputProps {
  */
 export interface ConditionalFilterItem {
   id?: string;
-  label?: Node;
+  label?: Element;
```
Hyperkid123, line 40: I am curious why Element and not Node. Further more, the item label is used in a `String(capitalize(label))` method. This points toward it being a string type if anything.

fhlavac: @Hyperkid123 replaced with `ReactNode`, thank you :slightly_smiling_face:

Hyperkid123 (approved):

PR #1619, opened 2022-10-05
Add Router v6 migration steps guide docs

Hyperkid123:
The email with info will be sent soon.

### DIFF
--- a/packages/docs/components/blog-nav/index.js
+++ b/packages/docs/components/blog-nav/index.js
@@ -1,0 +1,55 @@
+/* eslint-disable react/prop-types */
+import { createUseStyles } from 'react-jss';
+import Link from 'next/link';
+import { MenuContent, MenuItem, MenuList, Title } from '@patternfly/react-core';
+import BookOpenIcon from '@patternfly/react-icons/dist/esm/icons/book-open-icon';
+import classNames from 'classnames';
+
+const useStyles = createUseStyles({
+  blogContainer: {
+    padding: 8,
+  },
+  blogLink: {
+    color: 'var(--pf-c-menu__list-item--Color)',
+  },
+  '@media (min-width: 1200px)': {
+    blogContainer: {
+      position: 'absolute',
+      top: 80,
+      right: 16,
+    },
+  },
+});
+
+const BlogNav = () => {
+  const classes = useStyles();
+  return (
+    <div className={classes.blogContainer}>
+      <div className="pf-c-menu">
+        <MenuContent>
+          <MenuList iconSize="large">
+            <MenuItem>
+              <Title headingLevel="h2">Whats new?</Title>
+            </MenuItem>
+            <MenuItem
+              component={({ props, children }) => (
+                <Link href="/blog/router-v6">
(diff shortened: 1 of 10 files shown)


packages/docs/components/blog-nav/index.js
```diff
@@ -0,0 +1,55 @@
...
+        <MenuContent>
+          <MenuList iconSize="large">
+            <MenuItem>
+              <Title headingLevel="h2">Whats new?</Title>
+            </MenuItem>
+            <MenuItem
+              component={({ props, children }) => (
+                <Link href="/blog/router-v6">
+                  <button className="pf-c-menu__item">
+                    <a {...props} className={classNames(props?.className, classes.blogLink)} href="/blog/router-v6">
```
karelhala, line 38: Shouldn't there be a prevent default?

packages/docs/pages/blog/router-v6.mdx
```diff
@@ -0,0 +1,161 @@
+import { Alert } from '@patternfly/react-core'
+
+# Router v6 upgrade
+
+The router package [react-router-dom](https://reactrouter.com/en/main) has undergone a lot of changes in it lates v6 major release. Frontend applications in HCC are on a v5 which is no longer the LTS. Migration to V6 comes with some challenges. Mainly a router context can no longer be nested.
```
karelhala, line 5:
```suggestion
The router package [react-router-dom](https://reactrouter.com/en/main) has undergone a lot of changes in its latest v6 major release. Frontend applications in HCC are on v5 which is no longer the LTS. Migration to V6 comes with some challenges. Mainly a router context can no longer be nested.
```

packages/docs/components/blog-nav/index.js
Hyperkid123: No the <Link> component takes care of that. The `a` tag is actually required in order to properly render the tree.

packages/docs/pages/blog/router-v6.mdx
```diff
@@ -0,0 +1,161 @@
...
+
+## What are immediate actions?
+
+<Alert isInline variant="danger" title="Do not start migrating to router v6!">
+  The chrome UI has to be prepared first. Any application will have to adjust to chrome changes as well.
+</Alert>
+
+### My application has marked react-router-dom dependency as singleton
+
+If your application is on v5.x.x and has the router dependency marked as singleton, it will break once we upgrade chrome to v6. Please make sure this is not the case. This can only happen if you are using custom config, or have explicitly set the router dependency as shaped in your config.
```
karelhala, line 17:
```suggestion
If your application is on v5.x.x and has the router dependency marked as singleton, it will break once we upgrade chrome to v6. Please make sure this is not the case. This can only happen if you are using custom config, or have explicitly set the router dependency as shaped in your config.

Let us (**@crc-experience-team** [#forum-consoledot-ui](https://app.slack.com/client/T027F3GAJ/C023VGW21NU)) so we can keep track of your application and please upgrade your react-router-dom to v6.
```

karelhala (commented):
Looking good! Just a few typos but we are golden

PR #1625, opened 2022-10-10
fix(CypressUtils):adding missing package to the library

Fewwy:
We need the library to run the the cypress functions

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -33,7 +33,8 @@
         "react": "^16.14.0 || ^17.0.0",
         "react-dom": "^16.14.0 || ^17.0.0",
         "react-redux": ">=7.0.0",
-        "react-router-dom": "^5.0.0"
+        "react-router-dom": "^5.0.0",
+        "cypress": "^10.9.0"
     },
     "dependencies": {
         "@redhat-cloud-services/types": "^0.0.10",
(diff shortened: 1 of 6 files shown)


gkarat (commented):
@Hyperkid123, @Fewwy, can we put it to `peerDependencies` instead and define the version as `>=10.0.0 < 11.0.0`? I am not sure what are the benefits of putting it to `optionalDependencies`: the package will be still installed for everyone unless they explicitly add `--no-optional`.

gkarat:
And since this is just a collection of cypress functions (=library), I don't think we really need to put it to dependencies list here. It's just enough to mention here that these functions are compatible with 10.x version of cypress, and it will be up to the host apps (e.g. RHEL Advisor https://github.com/RedHatInsights/insights-advisor-frontend/pull/1028) to manually list the cypress dependency in its list.

packages/utils/package.json
```diff
@@ -33,11 +33,12 @@
         "react": "^16.14.0 || ^17.0.0",
         "react-dom": "^16.14.0 || ^17.0.0",
         "react-redux": ">=7.0.0",
-        "react-router-dom": "^5.0.0"
+        "react-router-dom": "^5.0.0",
+        "cypress": "^10.9.0"
```
gkarat, line 37:
```suggestion
        "cypress": ">=10.0.0 < 11.0.0"
```
please, so that we tolerate all 10.x versions

gkarat (changes requested):

packages/utils/package.json
Fewwy: Done

Hyperkid123 (approved):

gkarat (approved):

PR #1628, opened 2022-10-12
Use proper group label and correct group type

karelhala:
### Description

One last type adjustment, previously we incorrectly used `Group` as member of selected items, however it's `GroupItem` where we hold selected value `item` as reference for current item (of type `GroupFilterItem`) and `group` as reference to the group this item is member of.

Also dug a bit in the past and when we refactored this component to typescript we introduced a bug which prevented us from rendering groupped items.

### DIFF
--- a/packages/components/src/ConditionalFilter/GroupFilter.tsx
+++ b/packages/components/src/ConditionalFilter/GroupFilter.tsx
@@ -78,6 +78,15 @@ export interface TreeViewItem extends GroupFilterItem {
   name?: string;
 }
 
+export interface GroupItem {
+  /** Optional isSelected flag */
+  isSelected?: boolean;
+  /** Reference back to the group */
+  group: Group;
+  /** Current group filter item */
+  item: GroupFilterItem;
+}
+
 export interface Group {
   /** Optional groupSelectable flag. */
   groupSelectable?: boolean;
@@ -119,7 +128,7 @@ export interface GroupFilterProps {
   /** Optional text filter placeholder. */
   placeholder?: string;
   /** Selected filters object. */
-  selected?: Record<string, Record<string, Group | boolean>>;
+  selected?: Record<string, Record<string, GroupItem | boolean>>;
   /** Optional showMore button title element. */
   showMoreTitle?: React.ReactNode;
   /** Optional object containing properties for showMore element. */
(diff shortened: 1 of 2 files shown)


packages/components/src/ConditionalFilter/GroupFilter.tsx
```diff
@@ -317,7 +326,7 @@ const GroupFilter: React.FunctionComponent<GroupFilterProps> = ({
                 {menuItems.length > 0 && <MenuGroup>{renderItems(menuItems as GroupFilterItem[])}</MenuGroup>}
                 {groupMenuItems.map((group: GroupFilterItem | Group, groupKey: number) => (
                   <MenuGroup
-                    label={(group as Group).groupSelectable && typeof group.label === 'string' ? group.label : undefined}
+                    label={!(group as Group).groupSelectable && typeof group.label === 'string' ? group.label : undefined}
```
karelhala, line 329: https://github.com/RedHatInsights/frontend-components/blob/123a089577ca093cb9eb39decf23ec28f41578bf/packages/components/src/ConditionalFilter/GroupFilter.js#L190

Original code for reference. I was thinking of using something similar to what was used before, but the condition was even more complex. `(group as Group).groupSelectable ? undefined : (group.label === 'string' ? group.label : undefined)`. Although checking the docs, we can render anything, not only string.

karelhala: Oh, patternfly has incorrect types https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/Menu/MenuGroup.tsx#L11 I'll open an issue and PR on their side to allow React.Node elements as well.

Hyperkid123: Oh yeah, there are a few of these. Can you send the PF PR for reference? We might want to do some typecasting anyway so we are not blocked until next stable release.

karelhala: I've opened an issue so far https://github.com/patternfly/patternfly-react/issues/8209 while coding the PR I realised even cheekier approach. What if we'd want to send a component directly to the header? Currently working on writing tests etc.

Hyperkid123 (approved):

PR #1640, opened 2022-10-24
Enable webpack cache in fec dev script by default.

Hyperkid123:


fhlavac (approved):

packages/config/src/scripts/dev.webpack.config.js
```diff
@@ -47,6 +47,10 @@ const internalProxyRoutes = {
 };
 
 const { config: webpackConfig, plugins } = config({
+  // do not hash files in dev env
+  useFlieHash: false,
```
bastilian, line 51: Should probably be `useFileHash` :)

Hyperkid123: It should
 :+1:

bastilian (approved):
LGTM! Thank you @Hyperkid123!

PR #1641, opened 2022-10-24
Add _unstableHotReload to chunk mapper.

Hyperkid123:
Prerequisite for hot loading for module development.

Hot reload requires only a single runtime chunk which breaks our chunk mapper (no entry chunk is found).

Added new _**Unstable**_ way to look for entry chunks:

```JSON
{
  "starter": {
    "entry": [
      "/beta/apps/starter/starter.js",
      "/beta/apps/starter/js/runtime.js"
    ]
  }
}
```

Entry runtime chunks are now two, one for federated modules and one for global runtime. The federated module entry chunk is a part of the runtime chunk and requires it in order to initialize the webpack sharing scope.

These two should be now separate to not break existing builds but there is a possibility that we could use the same method for finding chunks in bot scenarios.

### DIFF
--- a/packages/config-utils/chunk-mapper.js
+++ b/packages/config-utils/chunk-mapper.js
@@ -1,35 +1,59 @@
+const hotChunkRegex = RegExp('\\.hot-update\\.js$');
 class ChunkMapper {
   constructor(options) {
+    const { _unstableHotReload, ...rest } = options;
     this.config = {};
-    this.options = options || {};
+    this.options = rest || {};
+    this._unstableHotReload = _unstableHotReload;
   }
 
   apply(compiler) {
     compiler.hooks.emit.tap('ChunkMapper', (compilation) => {
       const prefix = this.options.prefix || (RegExp('^/.*/$').test(compiler.options.output.publicPath) ? compiler.options.output.publicPath : '/');
-      compilation.chunks.forEach(({ name, files, runtime }) => {
+      if (this._unstableHotReload) {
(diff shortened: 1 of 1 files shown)


packages/config-utils/chunk-mapper.js
```diff
@@ -1,35 +1,59 @@
 class ChunkMapper {
   constructor(options) {
+    const { _unstableHotReload, ...rest } = options;
     this.config = {};
-    this.options = options || {};
+    this.options = rest || {};
+    this._unstableHotReload = _unstableHotReload;
+    this.hotChunkRegex = RegExp('\\.hot-update\\.js$');
```
karelhala, line 7: Can we add an option to customize this? By default we'd use the hot-update regexp, but we could add option to customize it.

karelhala (approved):

packages/config-utils/chunk-mapper.js
Hyperkid123: not sure if there is anything to customize :thinking: . Webpack will always call the hot-update chunks the same.

karelhala: If that's the thing I don't think we have to initialize the value on constructor level. But rather have it outside constructor.

Hyperkid123: Done

karelhala: I meant more like a class member, but this works as well :smiley:

Hyperkid123: Yeah sorry :D it does the same thing anyways

PR #1651, opened 2022-11-01
feat(CypressUtils): Docs for cypress and improved doctypes

Fewwy:
Docs for cypress and improved doctypes


packages/utils/doc/cypress.md
```diff
@@ -0,0 +1,114 @@
...
+The [cypress](https://www.cypress.io/) library is a great tool that allows to write scalable component tests.
+
+These utility functions are created to create scalable tests utilizing Patternfly UI library.
+
+## Usage
+1) Install Cypress v10+ on your project. You can find the documentation [here](https://docs.cypress.io/guides/component-testing/quickstart-react).
+2) Go through the installation and setup, then set up the first component test.
+3) Add more tests for components utilizing the functions we provide.
+
+### Table Functions
```
ikerreyes, line 12: Are all these generated automatically? Because otherwise, it will be quite an effort to have a function, document it and then have to changes the doc file. Maybe we should leave this just for the files. We can just point to the main folder containing all utilities. Anyone using them should be able to read the code docstrings.

bastilian: In theory, the documentation generator should parse the JS files and add them to the navigation. 

@Fewwy If you run the docker-compose setup, it should in theory start up the docs development server and you can doublecheck if any new documentation is popping up there for these functions.

Hyperkid123: Yup, we can parse the JS/TS docs in the source files from this package as well.

bastilian: @Fewwy You should run `docker-compose up`. This will boot up a container running the neccessary scripts to build the documentation served then via `localhost:3000`. You should see some changes to a navigation JSON file. 

You might need to move the markdown file into the docs package though for it to be picked up by the docs generator.

Hyperkid123: I can't remember from the top of my head but this specific package might be disabled. But just add the JS/TS doc and then remind me very day like 5 times and I can make it appear in the docs.

Fewwy: I added the navigation menu and moved .mdx file to the same folder as other utils components.

Fewwy: image

Hyperkid123:
I had to re-open to get travis unstuck

Hyperkid123 (approved):

PR #1667, opened 2022-11-14
feat: make fec compatible with monorepo setups

riccardo-forina:
Use `npm exec` to run `webpack` from wherever it's installed, letting the package manager figure the right path out instead of hardcoding it in the script.

This enables the use of `fec` in projects structured as monorepo.

### DIFF
--- a/packages/config/src/scripts/dev-script.js
+++ b/packages/config/src/scripts/dev-script.js
@@ -40,9 +40,8 @@ async function devScript(argv, cwd) {
     } else {
       configPath = resolve(__dirname, './dev.webpack.config.js');
     }
-    processArgs.push(`node_modules/.bin/webpack serve -c ${configPath}`);
     await setEnv(cwd);
-    spawn('node', processArgs, {
+    spawn(`npm exec -- webpack serve -c ${configPath}`, processArgs, {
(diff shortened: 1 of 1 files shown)


Hyperkid123 (approved):

packages/config/src/scripts/dev-script.js
```diff
@@ -40,9 +40,8 @@ async function devScript(argv, cwd) {
     } else {
       configPath = resolve(__dirname, './dev.webpack.config.js');
     }
-    processArgs.push(`node_modules/.bin/webpack serve -c ${configPath}`);
     await setEnv(cwd);
-    spawn('node', processArgs, {
+    await spawn(`npm exec -- webpack serve -c ${configPath}`, processArgs, {
```
Hyperkid123, line 44: We can't use the await, there are additional commands that can run in parallel bellow.

Hyperkid123 (changes requested):

packages/config/src/scripts/dev-script.js
riccardo-forina: 👍

Hyperkid123 (approved):

PR #1671, opened 2022-11-25
Wrap notifications portal with ErrorBoundary and use hooks instead of  connect

karelhala:
JIRA: RHCLOUD-23125

### Description

When using Notifications portal there can be a missmatch of react-redux versions. When this happens the UI breaks completely whic is not good for user experience. This PR wraps usage of redux context provider with error boundary and if such missmatch happens it will show an error but not break the entire UI.

There's also a new change to `ErrorBoundary` component - `silent` mode. When any error happens during render the error boundary will catch it and shows rather informative error. Sometimes we don't want to surface this error to customers, but just show it in console. This prop is just for that, not to render anything else other than `null`.

### DIFF
--- a/packages/notifications/src/NotificationPortal/NotificationPortal.tsx
+++ b/packages/notifications/src/NotificationPortal/NotificationPortal.tsx
@@ -1,28 +1,38 @@
-import { connect } from 'react-redux';
-import { ActionCreator, Dispatch } from 'redux';
+import React from 'react';
+import { useDispatch, useSelector } from 'react-redux';
 import Portal, { PortalNotificationConfig } from '../Portal';
 import { clearNotifications, removeNotification } from '../redux/actions/notifications';
+import { ErrorBoundary } from '@redhat-cloud-services/frontend-components/ErrorBoundary';
 
 export interface NotificationsState {
-  [key: string]: any;
   notifications?: PortalNotificationConfig[];
 }
 
 export interface NotificationPortalProps {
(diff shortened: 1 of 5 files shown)


Hyperkid123 (approved):

packages/notifications/src/NotificationPortal/NotificationPortal.tsx
```diff
@@ -1,28 +1,37 @@
...
+};
 
-const mapDispatchToProps = (dispatch: Dispatch, initialProps: NotificationPortalProps) => ({
-  removeNotification: initialProps.removeNotification ? initialProps.removeNotification : (id: number | string) => dispatch(removeNotification(id)),
-  onClearAll: initialProps.clearNotifications ? initialProps.clearNotifications : () => dispatch(clearNotifications()),
-});
-
-export const NotificationPortal = connect<NotificationsState>(mapStateToProps, mapDispatchToProps)(Portal);
+export const NotificationPortal: React.FC<NotificationPortalProps> = (props) => (
+  <ErrorBoundary headerTitle="Notifications portal" silent>
```
gkarat, line 31: Can we make it possible to override `silent` by the props from `NotificationPortal`? So that apps that import the portal can also have an ability to modify this behavior...

karelhala: Sure thing, it'll show generic error component above your UI.

PR #1677, opened 2022-12-02
Remove redux integration from Main component.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-23126


packages/components/src/Main/Main.tsx
```diff
@@ -1,110 +1,24 @@
...
-    return {
-      staticPart: [],
-      dynamic: undefined,
-    };
-  };
-
-  const { dynamic, staticPart } = calculateLocation();
+export const Main: React.FC<MainProps> = ({ children, className, ...props }) => {
+  useEffect(() => {
+    console.error(`Using deprecated "Main" component. Do not use it. Either remove it from your JSX or replace it by "section" HTML element.`);
```
bastilian, line 13: It would make sense to have a eslint check for this rather than log output.

Maybe there could be a general "check for deprecated FEC stuff" and have a list/config in the shared eslint for those function/component names.

Hyperkid123: I can do that as a follow. We can add it to the existing rule. But it's not a trivial fix. I will make a note to work on it.

bastilian: @Hyperkid123 That would be awesome! Thank you! I think we don't necessarily need a fix or even make it an error. It can just be a warning without a fix and only suggest to remove or replace it. The eslint errors are a bit more "in your face" since they'll pop up everywhere. :D

packages/components/src/Main/Main.tsx
```diff
@@ -1,110 +1,24 @@
...
-              page-type={staticPart.join('-')}
-              className={`${classNames(className, 'pf-l-page__main-section pf-c-page__main-section')}`}
-            >
-              {children}
-            </section>
-          ),
-        }[theme];
-      }}
-    </DarkContext.Consumer>
+    <section {...props} className={`${classNames(className, 'pf-l-page__main-section pf-c-page__main-section')}`}>
```
bastilian, line 16: Maybe we should not deprecate this component. Testing the PR it looked like it provides some essential styling that some apps appear to rely on.

jiridostal (approved):

karelhala (approved):

PR #1679, opened 2022-12-06
fix(TemplateProcessor): fix asterisk rendering in pydata

tompsota:
Fixes https://issues.redhat.com/browse/ADVISOR-2913.

### DIFF
--- a/packages/advisor-components/src/TemplateProcessor/TemplateProcessor.tsx
+++ b/packages/advisor-components/src/TemplateProcessor/TemplateProcessor.tsx
@@ -15,19 +15,26 @@ const DOT_SETTINGS = {
   strip: false,
 };
 
+// codespan and code tags escape child text by default, therefore we have to remove all escapes from there
+const walkTokens = (token: marked.Token) => {
+  if (token.type === 'codespan' || token.type === 'code') {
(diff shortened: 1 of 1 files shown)


gkarat (changes requested):
Thanks! I tested it locally with content-preview and ocp-advisor, and it fixes the issue. I'd ask for one extra thing: can you also escape underscores (_)? So that "\_test\_" doesn't result in _test_, but \_test\_.

P.S.: we have to perceive it rather as a workaround, because if you try \* test \*, then you will have an unordered list. That means in the future we have to properly disable any parsing of the content after the template is poly-filled with data definitions. Given that such cases are very rare/do not happen in rule content, then it can be low-prioritized.

tompsota:
> I'd ask for one extra thing: can you also escape underscores (_)? So that "_test_" doesn't result in test, but _test_.

This should be fixed now

? if you try * test *, then you will have an unordered list

I tried to reproduce the issue, but * test * in pydata results in `* test *` as expected, see the screenshot from Content Preview. If there isn't any special edge case where this is broken we can close this without any follow-up issues.

Screenshot 2022-12-08 at 12 14 17

PR #1694, opened 2023-01-10
[RHCLOUD-21728] Remove timestamps from generated files

BlakeHolifield:
* Removes timestamps from generated filenames
* Adds tests for timestamp removal and `useFileHash` options.

returns `js/[name].[fullhash].js`  or  `js/[name].js`

generates fed-mods with rbac as:

``` json
{
    "rbac": {
        "entry": [
            "/beta/apps/rbac/rbac.5301862932bb2afa7124.js"
        ],
        "modules": [
            "/beta/apps/rbac/js/832.5301862932bb2afa7124.js",
            "/beta/apps/rbac/css/832.4f0f4b061e35f2c8b0e7.css"
        ]
    }
}
```

### DIFF
--- a/packages/config-utils/federated-modules.js
+++ b/packages/config-utils/federated-modules.js
@@ -25,6 +25,7 @@ module.exports = ({ root, exposes, shared = [], debug, moduleName, useFileHash =
 
   const { dependencies, insights } = require(resolve(root, './package.json')) || {};
   const appName = moduleName || (insights && jsVarName(insights.appname));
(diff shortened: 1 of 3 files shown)


Hyperkid123 (commented):
@BlakeHolifield I think we just straight up remove the timestamp for everybody. We should not have any issues with Akamai as w have changed the net storage policy a bit lately.

Hyperkid123 (approved):

PR #1713, opened 2023-02-13
Migrate to dynamic plugin sdk

Hyperkid123:
### Changes
- make webpack config build output compatible with open shift dynamic plugin SDK

### TODO
- [x] figure out the missing publicPath in the new manifest
- [x] wait for chrome SDK@3 is in prod-stable

note: We should release a new major version after this because the build will no longer support HTML webpack plugin and cannot be used for stand-alone builds.


Hyperkid123:
I've tested the integration on source UI and it works flawlessly. Some chrome updates are required first. We will also require some manifest pre-processing on the chrome side for compatibility reasons.

Hyperkid123:
We have beta versions for both packages (5.0.0-beta.0 configs and 2.0.0-beta.0 utils) released and tested with chrome using the sdk@v3. All works fine. I think we might want to do a manual stable release once chrome has migrated to v3.

Hyperkid123:
We will have a follow-up that allows adding extensions and changing the version if the plugin config.

packages/config-utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@redhat-cloud-services/frontend-components-config-utilities",
-    "version": "1.5.31",
+    "version": "2.0.0-beta.1",
```
karelhala, line 3: Any reason for the beta tag name? It will be overwritten by the release script anyway.

karelhala (approved):
Looking good!

packages/config-utils/package.json
Hyperkid123: I think I will do a manual major release. I have used these to create the same beta releases for testing.

karelhala (approved):

karelhala (approved):

PR #1715, opened 2023-02-22
Fix GroupFilter padding

AsToNlele:
[ESSNTL-2639](https://issues.redhat.com/browse/ESSNTL-2639)

This also fixes css in [Global filter](https://github.com/RedHatInsights/insights-chrome/blob/9c5a398076163a25a58a155d2343d985a926ed68/src/components/GlobalFilter/global-filter-menu.scss#L17-L18)

Before:
image

After:
image


packages/components/src/ConditionalFilter/group-filter.scss
```diff
@@ -1,6 +1,7 @@
 @import '~@redhat-cloud-services/frontend-components-utilities/styles/_all.scss';
 
 button.pf-c-menu-toggle {
```
Hyperkid123, line 3: I think this breaks becayse of usage of global styling. Can we scope it with some className? This can break other elements on screen with the same selector.

I believe that was the reason of the broken padding.

Hyperkid123 (approved):

PR #1717, opened 2023-02-28
Scalprum version

mkholjuraev:
This PR bumps the scalprum core package version in components and remediation packages. I have found out that as the current versions do not match the chrome one, there are 3 instances of the same package in the browser. This PR removes one of them. One more instance could be removed if pdf-generator package had the same version, but upgrading it seems to break change and I am not able to fix the issue.

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3521,15 +3521,13 @@
         },
         "node_modules/@frsource/base64": {
             "version": "1.0.4",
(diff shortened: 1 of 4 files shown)


Hyperkid123 (approved):

Hyperkid123:
@mkholjuraev the CI is stuck on an invalid lock file. Can you re-install your dependencies locally?

PR #1726, opened 2023-03-22
Add eslint rule for chromeAPI call

victoria-dos:
Deprecated chromeAPI call from the window:
https://github.com/RedHatInsights/insights-chrome/pull/2310

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -6,7 +6,7 @@ const config = {
   coverageDirectory: './coverage/',
   collectCoverage: true,
   transformIgnorePatterns: [
-    'node_modules/(?!@patternfly|@data-driven-forms)',
+    'node_modules/(?!@patternfly|@data-driven-forms|lodash-es|@openshift)',
 
     // Uncomment the below line if you face any errors with jest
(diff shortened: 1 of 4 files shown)


Hyperkid123 (approved):

bastilian:
Thank you for this! 

This lint rule is already going to be helpful to find lingering instances where `insights.chrome` is still used, but would it be possible to also add a rule to ensure (just for consistency) that `useChrome` is only used via the default import? 

In the rush of moving some apps to the hook we may not have been consistent with that, and it might avoid headaches later if we try to use the same import right away everywhere.

Hyperkid123:
@bastilian That can be done, but probably not in this PR. Feel free to open a issue/ticket and assign it to me. We'll pick up when we can. Or if you'd like you can open a PR.

Hyperkid123:
@victoria-dos you have a couple of new eslint error caused because of this new rule in this file `packages/utils/src/interceptors/interceptors.ts`.

Hyperkid123 (approved):

PR #1734, opened 2023-04-04
feat(utils/interceptors): Provide axios interceptors via hook

bastilian:
This introduces a hook to provide the axios interceptors with chrome via `useChrome` to allow applications to migrate to interceptors that do not call chrome via the global scope. 

An example of an application migrating to this hook can be found here: https://github.com/RedHatInsights/sed-frontend/pull/191


adonispuente (approved):
Was able to test with the sed-frontend PR, works as intended. LGTM!

PR #1740, opened 2023-04-25
fix(advisor components): fix for broken hydra api links

Fewwy:
The Hydra API returns an incorrect link in the response. Unfortunately, backend devs don't seem interested in fixing this so I created this bandage fix. 
I also included a FIX that returns the correct link depending on the PROD or STAGE environment.
image

how to test - run rhel advisor with frontend components using npm link.
This is the system that you should check https://stage.foo.redhat.com:1337/insights/advisor/systems/7cc1aa10-84c3-423e-b6f0-d79af025430d#SIDs=&tags=
If the system doesn't exist - ping me so I can run a script that uploads it


packages/advisor-components/src/ReportDetails/ReportDetails.tsx
```diff
@@ -35,6 +35,11 @@ const ReportDetails: React.FC<ReportDetailsProps> = ({ report, kbaDetail, kbaLoa
     }
   };
 
+  const fixBrokenHydraLink = (url: string) => {
+    const linkToArray = url.split('/');
+    return `https://access.redhat.com/solutions/${linkToArray.at(-1)}`;
```
Hyperkid123, line 40: I don't really know what is this supposed to be used for but, what about stage env. The stage uses `access.stage...`. Is this supposed to always hit prod?

Fewwy: Links lead to the prod even on the stage env 
image

Hyperkid123 (approved):

gkarat (approved):
LGTM, tested locally and compared with the current links mapping at console.stage.redhat.com. All links seem to be leading to correct URLs now 👍🏼 .

You have one error in your tests:
```
Error: Image diff factor (0.072%) is bigger than maximum threshold option 0.025.
```
Not sure why this difference appear (your change doesn't introduce any visual change). I'd suggest you to increase the threshold here https://github.com/Fewwy/frontend-components/blob/reportdetails-fix/packages/advisor-components/src/ReportDetails/ReportDetails.spec.ct.js#L23. This should temporarily fix the problem

Hyperkid123:
The issue might be with a different OS. If the snapshot was made on MacOS, the CI runs on Linux are there are some browser differences (Text boldness, scroll bar, etc.)

PR #1742, opened 2023-04-27
Fix GroupFilter styling on inputs

AsToNlele:
[ESSNTL-2639](https://issues.redhat.com/browse/ESSNTL-2639)

This is a fix for my [previous fix](https://github.com/RedHatInsights/frontend-components/pull/1715). I didn't notice, it broke the 'Operation system' filter in Inventory, sorry about that.

I wanted to do:
`button.ins-c-group-menu-toggle:has(input) {
    padding: 0;
}`
but `:has()` selector is not supported in Firefox yet

Before:
image

After:
image

All other filters are displayed correctly
image

### DIFF
--- a/packages/components/src/ConditionalFilter/group-filter.scss
+++ b/packages/components/src/ConditionalFilter/group-filter.scss
@@ -1,7 +1,7 @@
 @import '~@redhat-cloud-services/frontend-components-utilities/styles/_all.scss';
 
-button.ins-c-group-menu-toggle {
-    padding: 0;
+button.ins-c-group-menu-toggle .pf-c-form-control {
+    margin: -1rem 0 -1rem 0;
 }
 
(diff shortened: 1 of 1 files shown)


packages/components/src/ConditionalFilter/group-filter.scss
```diff
@@ -1,7 +1,7 @@
 @import '~@redhat-cloud-services/frontend-components-utilities/styles/_all.scss';
 
-button.ins-c-group-menu-toggle {
-    padding: 0;
+button.ins-c-group-menu-toggle input {
+    margin: -7px 0px -7px 0px;
```
Hyperkid123, line 4: We should use available PF CSS variables.

Hyperkid123 (approved):

PR #1749, opened 2023-05-25
fix(advisor-3041): fix for reportdetails

Fewwy:
In my last PR, I totally forgot that we have both "articles" and "solutions" links
This should remove the hardcoded "solutions" and put "solutions" or "articles" depending on what is returned by the API

System ID to test 1daf9c25-ac62-4d27-a44d-4f59276ed770 in RHEL Advisor


mkholjuraev (approved):
Codewise LGTM!

PR #1752, opened 2023-05-30
DRAFT: enable source map generation for prod-beta env without source urls

mkholjuraev:
This is a part of effort to try uploading source maps using sentry-cli during release pipeline. Further PRs are to be created to demo the change. I have verified that `append:false` inside webpack config prevents adding sourmap urls into source files.

[Here](https://github.com/RedHatInsights/patchman-ui/pull/1068) is another PR in tenant app that gives a bit more insight how those apps will upload the sourcemaps.


mkholjuraev:
Found better way to upload source artifacts using sentry/webpack plugin.Closing this.

PR #1757, opened 2023-06-09
fix(advisor-3041): fixing external link crash

Fewwy:
Fixing more issues with this advisor component
https://issues.redhat.com/browse/ADVISOR-3041


AsToNlele (approved):
After linking this PR with Advisor it fixes the issue :+1:

Hyperkid123 (approved):

PR #1764, opened 2023-06-19
Reexport components already migrated to PF

fhlavac:
[RHCLOUD-26558](https://issues.redhat.com/browse/RHCLOUD-26558)

Reexported components already migrated to PF. In some cases added Red Hat specific stuff to not cause issues for the existing imports.


fhlavac:
Added eslint rule to warn about using deprecated imports
image

packages/eslint-config/lib/rules/disallow-pf-migrated-components.js
```diff
@@ -0,0 +1,50 @@
...
+    },
+    fixable: 'code',
+    schema: [],
+  },
+  create: function (context) {
+    return {
+      ImportDeclaration(node) {
+        const sourceValue = node.source.value;
+        const specifiers = node.specifiers;
+        const forbiddenImports = ['ErrorBoundary', 'NotAuthorized', 'ErrorState'];
```
karelhala, line 22: Can we move this out of the function so it's more visible and easier for us to add new deprecated values?

karelhala (commented):
Looking good! Just a minor question.

packages/components/src/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -1,75 +1,13 @@
...
-      );
-    }
-
-    return this.props.children;
-  }
-}
-
-export default ErrorBoundaryPage;
+import React from 'react';
+import { ErrorBoundary as ErrorBoundaryPF } from '@patternfly/react-component-groups';
```
Hyperkid123, line 2: We should be exporting the `dist/dynamic` absolute path to preserve the module federation compatibility.

Hyperkid123 (changes requested):

Hyperkid123 (approved):

Hyperkid123:
@fhlavac the groups repo will be ported to pf4@5 soon. Not sure if we want to wait a little bit.

karelhala (approved):

Hyperkid123 (approved):

PR #1767, opened 2023-06-26
Use better image linking so mdx provider can intercept it

karelhala:
#### Description

MDX provider has a bit of hard time finding linked img elements. That's because we used directly `` component. It's better to use direct imports `` instead so we can intercept them when building the docs site.


apinkert (approved):
neat 📸

PR #1773, opened 2023-06-30
Pdf export

mkholjuraev:
Adds useExportPDf hook. I will add documentation in a while.


adonispuente (approved):
Code-wise this seems good!

PR #1775, opened 2023-07-04
ESSNTL-920 Last seen color does not meet color contrast

jiridostal:


bastilian (approved):
LGTM! Thank you @jiridostal!

PR #1776, opened 2023-07-04
feat(proxy): RHIF-268 - Add local app route feature

bastilian:
This adds a feature to allow any app that uses the proxy configuration to pass a environment variable to configure routes for other applications locally with

```
$ LOCAL_APPS=advisor:8002,compliance:8003 npm run start:proxy
```

Currently this is done in some, but not all applications, and it sometimes slightly differs in approaches. This consolidates this in one place and allows us to have a consistent way of running multiple applications across all applications.

### DIFF
--- a/packages/config-utils/proxy.js
+++ b/packages/config-utils/proxy.js
@@ -16,6 +16,68 @@ const { registerChrome } = require('./standalone/services/default/chrome');
 
 const defaultReposDir = path.join(__dirname, 'repos');
 
+const checkLocalAppHost = (appName, hostUrl, port) => {
+  const check = execSync(`curl --max-time 5 --silent --head ${hostUrl} | awk '/^HTTP/{print $2}'`).toString().trim();
+
+  if (check !== '200') {
+    console.error('\n' + appName[0].toUpperCase() + appName.substring(1) + ' is not running or available via ' + hostUrl);
+    console.log(
+      '\nMake sure to run `npm run start -- --port=' +
+        port +
+        '` in the ' +
+        appName +
+        " application directory to start it's webpack dev server and the bundle is built.\n"
+    );
+
+    return false;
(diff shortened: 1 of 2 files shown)


packages/config-utils/proxy.js
```diff
@@ -39,9 +97,12 @@ module.exports = ({
   bounceProd = false,
   useAgent = true,
   useDevBuild = true,
+  localApps = process.env.LOCAL_APPS,
+  defaultLocalAppHost = process.env.LOCAL_APP_HOST || 'stage.foo.redhat.com',
```
Hyperkid123, line 101: The `defaultLocalAppHost` can also be `prod.foo.redhat.com`. It should be adjusted based on the env.

Hyperkid123 (commented):
:100:

packages/config-utils/proxy.js
bastilian: @Hyperkid123 The `defaultLocalAppHost` is now dependent on the `majorEnv` and should work for prod and stage now.

packages/config-utils/proxy.js
```diff
@@ -84,26 +149,13 @@ module.exports = ({
     routes = require(routesPath);
   }
 
-  if (routes) {
+  if (routes?.length > 0) {
```
Hyperkid123, line 152: This expects the routes to be an array but by default, it is an object. This will disable any existing proxy configurations using the object.

Hyperkid123 (changes requested):

packages/config-utils/proxy.js
bastilian: Reverted that.

Hyperkid123 (approved):

Hyperkid123 (approved):

PR #1787, opened 2023-08-01
PF4@5 upgrade

Hyperkid123:
**This one is not getting merged to master!**

The target branch is a temporary version 5 branch. More verification will be required before we can release the next major version of FEC packages.

### TODO
- [x] update dependencies
- [x] pass build (includes usage of deprecated components)
- [x] replace deprecated components that do not require major refactoring
- [x] align component API
- [x] fix failing tests
- [x] adjust cypress to use official component testing
- [x] create new branch for legacy components and short term support before all applications migrate to PF5 

### Breaking changes
- making many props mandatory
  - mainly change/click/etc callbacks
  - components do not bring any value without these
- Removed pagination component entirely. Duplicate of PF pagination
- Removed custom Input component. Duplicate of PF TextInput


packages/components/src/Filters/FilterInput.tsx
```diff
@@ -25,18 +25,32 @@ const FilterInput: React.FC<FilterInputProps> = ({ addRemoveFilters, param = '',
     }
   }, []);
   const handleChange = () => {
+    console.log({ type, value, param, checked });
```
apinkert, line 28: do we still need this log?

Hyperkid123: We don't

Hyperkid123:
I am merging this into the pre-release branch to unblock the rest of the work.

PR #1884, opened 2023-08-11
Config update

Hyperkid123:
### TODO:
- [x] use TS for the source files
- [ ] use TS for the binary files
- [ ] add new utilities functions to enable easier migration
- [ ] ditch ts-loader and babel-loader for swc-loader
- [ ] add a script to create base tsconfig.json for all apps (if they don't have it)


Hyperkid123:
Replaced by different PR

PR #1886, opened 2023-08-13
fix: Fix usage of null values in RBAC RDs

gkarat:
Make sure null values are not left out when RBAC RD value is an array.


gkarat:
Ooopsie, let me remove these console.logs first 😄

bastilian (approved):
LGTM! Thank you @gkarat!

PR #1892, opened 2023-08-17
Apply new tooling to the shared configuration

Hyperkid123:
### Changes
- migrate config sources files to TS
- add fec preflight check script to install required development dependencies
- apply explicit PF module sharing list to the fec build scrips


fhlavac (approved):

packages/config-utils/src/fec-logger.ts
```diff
@@ -0,0 +1,36 @@
...
+};
+
+const logFunctions = {
+  [LogType.error]: console.error,
+  [LogType.warn]: console.warn,
+  [LogType.info]: console.info,
+  [LogType.debug]: console.debug,
+};
+
+function getFecMessage(level: LogType, ...data: any[]) {
```
florkbr, line 30: @Hyperkid123 not sure what convention we use in this repo yet, but I typically try to avoid `any` where possible.

packages/config/.eslintrc
```diff
@@ -2,5 +2,8 @@
     "env": {
         "browser": false,
         "node": true
+    },
+    "rules": {
+        "@typescript-eslint/no-var-requires": "off"
```
florkbr, line 7: Are we sure we want to turn this off globally for the entire monorepo? Or should we disable it per-case basis?

packages/config/.gitignore
```diff
@@ -0,0 +1,6 @@
+bin
+lib
+!src/bin
+!src/bin/*
```
florkbr, line 4: Do we really need both entries the folder and the wildcard here?

packages/config/.npmignore
```diff
@@ -1,9 +1,12 @@
...
 .babelrc
 babel.config.js
 doc/
 config/
 .eslintrc
+tsconfig.json
+src
+
+!lib/
+!lib/*
```
florkbr, line 10: Same here

packages/config/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+      "isolatedModules": true,
+      "plugins": [],
+      "lib": ["ES2015"]
+  },
+  "include": [
+      "./src/**/*"
+  ],
+  "exclude": [
+    "./src/**/*.test.js",
+    "./src/**/*.spec.js",
```
florkbr, line 20: Doesn't seem to be an issue now, but we also may want to catch the `.jsx` extension in the future

packages/config/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+      "rootDir": "./src",
+      "outDir": "./",
+      "module": "commonjs",
+      "target": "ES5",
+      "declarationDir": "./",
+      "allowJs": true,
```
florkbr, line 9: should we consider enforcing [strict](https://www.typescriptlang.org/tsconfig#strict)? Perhaps that can be a followup...

packages/config/src/lib/index.ts
```diff
@@ -0,0 +1,106 @@
+import { LogType, fecLogger } from '@redhat-cloud-services/frontend-components-config-utilities';
```
florkbr, line 1: Github wasn't able to diff this with the old file location so I went ahead and did it locally:
```diff
➜  Desktop diff config-index-old.js config-index-new.ts -w
1,2c1,3
< const config = require('./src/config');
< const plugins = require('./src/plugins');
---
> import { LogType, fecLogger } from '@redhat-cloud-services/frontend-components-config-utilities';
> import createConfig, { CreateConfigOptions } from './createConfig';
> import createPlugins, { CreatePluginsOptions } from './createPlugins';
3a5,6
> export * from './createConfig';
> export * from './createPlugins';
4a8,13
> type DevServerConfiguration = import('webpack-dev-server').Configuration;
> type WebpackConfiguration = import('webpack').Configuration;
> export interface FecWebpackConfiguration extends WebpackConfiguration {
>   devServer?: DevServerConfiguration;
> }
> 
11c20
< const getAppEntry = (rootFolder, isProd) => {
---
> const getAppEntry = (rootFolder: string, isProd?: boolean) => {
16c25
<       console.warn('Found multiple entry-dev files. Using', entries[0]);
---
>       fecLogger(LogType.warn, `Found multiple entry-dev files. Using: ${entries[0]}`);
26c35
<     console.warn('Found multiple entry files. Using', entries[0]);
---
>     fecLogger(LogType.warn, `Found multiple entry files. Using: ${entries[0]}`);
32c41,53
< module.exports = (configurations) => {
---
> type FecConfigurationOptions = Omit<CreateConfigOptions, 'publicPath' | 'appEntry' | 'appName'> &
>   CreatePluginsOptions & {
>     deployment?: string;
>     debug?: boolean;
>     appEntry?: string;
>   };
> 
> const createFecConfig = (
>   configurations: FecConfigurationOptions
> ): {
>   config: ReturnType<typeof createConfig>;
>   plugins: ReturnType<typeof createPlugins>;
> } => {
40c61
<     console.log('[fec] no git branch detected, using main for webpack "main" config.');
---
>     fecLogger(LogType.info, 'no git branch detected, using main for webpack "main" config.');
50,59c71,80
<     /* eslint-disable no-console */
<     console.log('~~~Using variables~~~');
<     console.log(`Root folder: ${configurations.rootFolder}`);
<     console.log(`Current branch: ${gitBranch}`);
<     !generateSourceMaps && console.log(`Source map generation for "${gitBranch}" deployment has been disabled.`);
<     console.log(`Beta branches: ${betaBranches}`);
<     console.log(`Using deployments: ${appDeployment}`);
<     console.log(`Public path: ${publicPath}`);
<     console.log(`App entry: ${appEntry}`);
<     console.log(`Use proxy: ${configurations.useProxy ? 'true' : 'false'}`);
---
>     console.group();
>     fecLogger(LogType.debug, '~~~Using variables~~~');
>     fecLogger(LogType.debug, `Root folder: ${configurations.rootFolder}`);
>     fecLogger(LogType.debug, `Current branch: ${gitBranch}`);
>     !generateSourceMaps && fecLogger(LogType.debug, `Source map generation for "${gitBranch}" deployment has been disabled.`);
>     fecLogger(LogType.debug, `Beta branches: ${betaBranches}`);
>     fecLogger(LogType.debug, `Using deployments: ${appDeployment}`);
>     fecLogger(LogType.debug, `Public path: ${publicPath}`);
>     fecLogger(LogType.debug, `App entry: ${appEntry}`);
>     fecLogger(LogType.debug, `Use proxy: ${configurations.useProxy ? 'true' : 'false'}`);
61,62c82,83
<       console.warn('Insights-proxy is deprecated in favor of "useProxy" or "standalone".');
<       console.warn('See https://github.com/RedHatInsights/frontend-components/blob/master/packages/config/README.md');
---
>       fecLogger(LogType.warn, 'Insights-proxy is deprecated in favor of "useProxy" or "standalone".');
>       fecLogger(LogType.warn, 'See https://github.com/RedHatInsights/frontend-components/blob/master/packages/config/README.md');
65c86
<     console.log('~~~~~~~~~~~~~~~~~~~~~');
---
>     console.groupEnd();
70c91
<     config: config({
---
>     config: createConfig({
72,73d92
<       appDeployment,
<       insights,
78c97
<     plugins: plugins({
---
>     plugins: createPlugins({
81,83c100
<       appDeployment,
<       insights,
<       publicPath,
---
>       appname: insights.appname,
86a104,106
> 
> export default createFecConfig;
> module.exports = createFecConfig;
\ No newline at end of file
```

packages/config/src/bin/fec.ts
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env node
+const { execSync } = require('child_process');
```
florkbr, line 2: Same diff issue here. Done locally:
```diff
➜  Desktop diff fec-old.js fec-new.ts -w
3,4c3,6
< const static = require('@redhat-cloud-services/frontend-components-config-utilities/serve-federated');
< const yargs = require('yargs');
---
> import serveStatic from '@redhat-cloud-services/frontend-components-config-utilities/serve-federated';
> const fs = require('fs');
> const path = require('path');
> import yargs from 'yargs';
6,8c8,11
< const devScript = require('../src/scripts/dev-script');
< const buildScript = require('../src/scripts/build-script');
< const { logError, validateFECConfig } = require('../src/scripts/common');
---
> const devScript = require('./dev-script');
> const buildScript = require('./build-script');
> const { validateFECConfig } = require('./common');
> const { fecLogger, LogType } = require('@redhat-cloud-services/frontend-components-config-utilities');
22c25
< `
---
> `;
24c27
<         execSync(command)    
---
>     execSync(command);
26c29
<         logError('Unable to patch /etc/hosts! Please to run the script as sudo.')
---
>     fecLogger(LogType.error, 'Unable to patch /etc/hosts! Please to run the script as sudo.');
31a35,79
> function checkDependencies() {
>   const requiredDependencies = ['typescript', 'ts-patch', '@redhat-cloud-services/tsc-transform-imports'];
>   const missingDependencies: string[] = [];
>   const packageJsonFile = path.resolve(cwd, 'package.json');
>   // do not destruct the json to not mutate order of attributes in the file
>   const pckJson = JSON.parse(fs.readFileSync(packageJsonFile));
>   requiredDependencies.forEach((dep) => {
>     if (!pckJson.dependencies[dep] && !pckJson.devDependencies[dep]) {
>       missingDependencies.push(dep);
>     }
>   });
> 
>   if (!pckJson.scripts.postinstall || !pckJson.scripts.postinstall.includes('ts-patch install')) {
>     // prepend patch ts to existing postinstall command
>     if (pckJson.scripts.postinstall) {
>       pckJson.scripts.postinstall = `ts-patch install && ${pckJson.scripts.postinstall}`;
>     } else {
>       pckJson.scripts.postinstall = `ts-patch install`;
>     }
> 
>     fs.writeFileSync(packageJsonFile, JSON.stringify(pckJson, null, 2));
>   }
> 
>   return missingDependencies;
> }
> function patchTs(dependencies: string) {
>   const usesYarn = fs.existsSync(path.resolve(cwd, 'yarn.lock'));
>   if (dependencies.length > 0) {
>     let command;
>     if (usesYarn) {
>       command = `yarn add -D ${dependencies}`;
>     } else {
>       command = `npm i --save-dev ${dependencies}`;
>     }
>     fecLogger(LogType.info, 'Installing missing build dependencies: ', command);
>     execSync(command, { stdio: 'inherit', cwd });
>   }
>   fecLogger(LogType.info, 'Patching TS');
>   execSync('npx ts-patch install', { stdio: 'inherit', cwd });
>   if (!fs.existsSync(path.resolve(cwd, 'tsconfig.json'))) {
>     fecLogger(LogType.info, 'Creating base tsconfig.json');
>     fs.copyFileSync(path.resolve(__dirname, './tsconfig.template.json'), path.resolve(cwd, 'tsconfig.json'));
>   }
> }
> 
35c83,84
<     yargs.positional('config', {
---
>     yargs
>       .positional('config', {
38,39c87,89
<         describe: 'Path to webpack config'
<     }).positional('port', {
---
>         describe: 'Path to webpack config',
>       })
>       .positional('port', {
43c93
<         default: 8003
---
>         default: 8003,
46c96
< .command('patch-etc-hosts', 'You may have to run this as \'sudo\'. Setup your etc/hosts allow development hosts in your browser')
---
>   .command('patch-etc-hosts', "You may have to run this as 'sudo'. Setup your etc/hosts allow development hosts in your browser")
50a101
>     });
52d102
< })
56a107
>     });
58d108
< })
68,69c118
< .help()
< .argv;
---
>   .help().argv;
71,72c120,121
< const scripts = {
<     static: (argv, cwd) => {
---
> const scripts: { [name: string]: (...args: any[]) => void } = {
>   static: (argv: any, cwd: string) => {
74,75c123,124
<         validateFECConfig(cwd)
<         static(argv, cwd)
---
>     validateFECConfig(cwd);
>     serveStatic(argv, cwd);
78,80c127,129
<     dev: (argv, cwd) => {
<         validateFECConfig(cwd)
<         devScript(argv, cwd)
---
>   dev: (argv: any, cwd: string) => {
>     validateFECConfig(cwd);
>     devScript(argv, cwd);
82c131,132
<     build: buildScript
---
>   build: buildScript,
>   'patch-ts': () => patchTs(checkDependencies().join(' ')),
88c138,142
<     if (!argv._.length || argv._.length === 0) {
---
>   const missingDependencies = checkDependencies();
>   if (missingDependencies.length > 0) {
>     patchTs(missingDependencies.join(' '));
>   }
>   if (!(argv as any)._.length || (argv as any)._.length === 0) {
93c147
<     scripts[argv._[0]](...args);
---
>   scripts[(argv as any)._[0]](...args);
```

packages/config/src/bin/build-script.ts
```diff
@@ -16,19 +18,20 @@ function buildScript(argv, cwd) {
     cwd,
     shell: true,
   });
-  subprocess.on('error', function (err) {
-    logError(err);
+  subprocess.on('error', function (err: any) {
```
florkbr, line 21: Again I think ideally we'd avoid the use of `any`. Maybe we could add todo's to follow up this migration

packages/config/src/bin/dev-script.ts
```diff
@@ -28,11 +28,18 @@ async function setEnv(cwd) {
...
+    clouddotEnv?: string;
+    uiEnv?: string;
+  },
+  cwd: string
+) {
   try {
     validateFECConfig(cwd);
 
-    const fecConfig = require(process.env.FEC_CONFIG_PATH);
+    const fecConfig = require(process.env.FEC_CONFIG_PATH!);
```
florkbr, line 42: I'm not sure if the trailing `!` is really what we want to do in these various locations where `process.env.FEC_CONFIG_PATH` is accessed... Are we actually guaranteed the env var will be set and not null/undefined?

packages/config/src/bin/dev.webpack.config.ts
```diff
@@ -1,18 +1,22 @@
...
+  return isBeta ? [new RegExp(`/beta${url.toString()}`), new RegExp(`/preview${url.toString()}`)] : [new RegExp(url.toString())];
 }
 
-function createAppUrl(appUrl) {
+function createAppUrl(appUrl: string | string[] | (string | RegExp)[]) {
   if (Array.isArray(appUrl)) {
     return appUrl
       .map((url) => {
-        if (typeof url === 'object') {
+        if (url instanceof RegExp) {
```
florkbr, line 19: :+1: for the tighter check here. Hopefully no one is relying on passing in a native object.

packages/config/tsconfig.json
florkbr: Nevermind. I see we are using strict here: https://github.com/RedHatInsights/frontend-components/pull/1892/files#diff-b87edffaa739c304571395ae03f1a69afd153afc0b56374e9c6a7c56e1baa204R12 👍

packages/config/src/lib/config.test.js
```diff
@@ -76,18 +82,7 @@ describe('rootFolder', () => {
...
-  test('first to be chrome-render-loader', () => {
-    const { module } = configBuilder({ appEntry: 'testEntry', appName: 'someName' });
-    expect(new RegExp(module.rules[0].rules).test('testEntry')).toBe(true);
-    expect(module.rules[0].options.skipChrome2).toBe(false);
-  });
-
-  test('first to be chrome-render-loader', () => {
-    const { module } = configBuilder({ appEntry: 'testEntry', appName: 'someName', skipChrome2: true });
-    expect(module.rules[0].options.skipChrome2).toBe(true);
+    expect(module.rules.length).toBe(4);
```
florkbr, line 85: Do we really need to check the _exact_ length here? This just seems fragile to me if we add other rules in the future. Maybe we should confirm the rules.length > 0

packages/config/src/lib/createPlugins.ts
```diff
@@ -7,9 +10,40 @@ const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin'
 const { glob } = require('glob');
 const path = require('path');
 
-module.exports = ({ rootFolder, insights, generateSourceMaps, plugins, definePlugin = {}, _unstableHotReload = false, useFileHash = true } = {}) => {
+export type WebpackPluginDefinition = undefined | null | false | '' | 0 | ((this: Compiler, compiler: Compiler) => void) | WebpackPluginInstance;
+
+export type CreatePluginsOptions = {
+  rootFolder: string;
```
florkbr, line 16: NIT - If some of these keys (such as `rootFolder`) are present and the same type between `CreateConfigOptions` and `CreatePluginsOptions` I wonder if we move those to a common type that can be extended to avoid duplication?

florkbr:
Overall looks good - just a few questions/comments. @Hyperkid123 is there a good way to test this locally (re: the PF module sharing list)?

packages/config-utils/src/fec-logger.ts
Hyperkid123: It is just an extension of the `console.` typings. Generally using `any` is not good. That said if we only extend an existing interface we use the original types if possible.

packages/config/.eslintrc
Hyperkid123: It's only for the `config` package, not the monorepo.

packages/config/.gitignore
Hyperkid123: Yes. https://stackoverflow.com/questions/5533050/gitignore-exclude-folder-but-include-specific-subfolder

TLDR
```
You must white-list folders first, before being able to white-list files within a given folder.
```

packages/config/src/bin/dev-script.ts
Hyperkid123: The `FEC_CONFIG_PATH`  is validated in the root of the `fec.js` binary. If it does not exist or is not valid, these scripts will not even be triggered. 

You can check the `validateFECConfig` function in the `common.ts` file.

packages/config/src/bin/dev.webpack.config.ts
Hyperkid123: I don't this so. I would not work.

Hyperkid123:
> Overall looks good - just a few questions/comments. @Hyperkid123 is there a good way to test this locally (re: the PF module sharing list)?

The only way to properly test this is within the application. I was using this PR as a testing example:https://github.com/RedHatInsights/sources-ui/pull/1147

CI Is falling because of changed PF selectors etc, but the app runs and builds as expected.

packages/config/src/lib/createPlugins.ts
Hyperkid123: I think we should.

packages/config/src/bin/build-script.ts
Hyperkid123: If we switch to an `import` statement we can use the default type. Let me do that

florkbr (approved):

PR #1900, opened 2023-08-28
fix: Fix always loading RBACProvider on empty appName

gkarat:
This fixes the RBACProvider component: now, with the empty appName property, it fetches permissions and update the state properly. This way, it is not stuck in the always loading state, and all the permissions from all scopes are fetched (empty `appName` == we request all of the user's permissions).

### DIFF
--- a/packages/components/src/RBACProvider/RBACProvider.tsx
+++ b/packages/components/src/RBACProvider/RBACProvider.tsx
@@ -19,15 +19,15 @@ const hasAccessWithUserPermissions = (userPermissions: (Access | string)[], chec
 };
 
 export interface RBACProviderProps {
-  appName: string;
-  checkResourceDefinitions: boolean;
+  appName?: string | null;
+  checkResourceDefinitions?: boolean;
 }
 
 export const RBACProvider: React.FunctionComponent<RBACProviderProps> = ({ appName, checkResourceDefinitions = false, children }) => {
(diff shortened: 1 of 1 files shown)


packages/components/src/RBACProvider/RBACProvider.tsx
```diff
@@ -38,9 +38,7 @@ export const RBACProvider: React.FunctionComponent<RBACProviderProps> = ({ appNa
   };
 
   useEffect(() => {
-    if (appName) {
```
karelhala, line 41: This would trigger the call to fetch permissions twice. It could be better to allow `null` value to be passed as `appName` and if such value request empty application.

karelhala (commented):
I have just one minor concern about firing multiple times the request to fetch permissions.

packages/components/src/RBACProvider/RBACProvider.tsx
gkarat: Yeah, this makes sense. Added `null` support

karelhala: I would flip it arround, if `null` or `string` fire a `fetchPermissions`

gkarat: @karelhala, implemented

PR #1921, opened 2023-10-03
Fix visited pages title values.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-28594

### Changes
- use Mutation observer to updated last visited data to latest valid document title

### DIFF
--- a/packages/chrome/src/ChromeProvider/ChromeProvider.test.tsx
+++ b/packages/chrome/src/ChromeProvider/ChromeProvider.test.tsx
@@ -1,7 +1,9 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { Link, MemoryRouter, Route, Routes } from 'react-router-dom';
 import { render, screen } from '@testing-library/react';
 import { act } from 'react-dom/test-utils';
+import { ScalprumContext } from '@scalprum/react-core';
+import { PluginStore } from '@openshift/dynamic-plugin-sdk';
 
 import ChromeProvider from './ChromeProvider';
 import * as fetch from '../utils/fetch';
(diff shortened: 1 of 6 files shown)


packages/chrome/src/ChromeProvider/ChromeProvider.test.tsx
```diff
@@ -53,8 +76,13 @@ describe('ChromeProvider', () => {
     await act(async () => {
       await flushPromises();
     });
+    // console.warn(postSpy.mock.calls)
```
florkbr, line 79: @Hyperkid123 can we remove this?

packages/chrome/src/ChromeProvider/ChromeProvider.tsx
```diff
@@ -7,28 +9,70 @@ import { IDENTITY_URL, LAST_VISITED_URL, get, post } from '../utils/fetch';
...
+        const currentTitle = mutations[0]?.target.textContent;
+        console.warn({ prevTitle, currentTitle, pathname });
+        // trigger only if the titles are different
+        if (typeof currentTitle === 'string' && currentTitle !== prevTitle) {
+          try {
+            prevTitle = currentTitle;
+            postData(pathname, currentTitle, scalprum.api.chrome.getBundleData().bundleTitle);
+          } catch (error) {
+            // catch sync errors
+            console.error('Unable to update last visited pages!', error);
```
florkbr, line 52: NIT (not blocking) - we have the same error catch/logger on line 24. Might be nice to create a single error handler for errors that occur anywhere in this logic.

packages/chrome/src/ChromeProvider/ChromeProvider.tsx
```diff
@@ -7,28 +9,70 @@ import { IDENTITY_URL, LAST_VISITED_URL, get, post } from '../utils/fetch';
...
+      /**
+       * Use Mutation observer to trigger the updates.
+       * Using the observer will ensure the last visited pages gets updated on document title change rather than just location change.
+       * The chrome service uses pathname as identifier and updates title according.
+       * Multiple calls with the same pathname and different title will ensure that the latest correct title is assigned to a pathname.       *
+       *  */
+      titleObserver = new MutationObserver((mutations) => {
+        // grab text from the title element
+        const currentTitle = mutations[0]?.target.textContent;
+        console.warn({ prevTitle, currentTitle, pathname });
```
florkbr, line 44: Do we have any mechanism to gate this to "debug" only scenarios?

packages/chrome/src/ChromeProvider/ChromeProvider.test.tsx
```diff
@@ -31,17 +32,39 @@ describe('ChromeProvider', () => {
...
       render(
-        <MemoryRouter initialEntries={['/']} initialIndex={0}>
-          <Routes>
-            <Route path="*" element={<Link to="/foo/bar">/foo/bar</Link>}></Route>
-          </Routes>
-          <ChromeProvider bundle="bundle-title" />
-        </MemoryRouter>
+        <ScalprumContext.Provider
+          // @ts-ignore
+          value={{
```
florkbr, line 50: NIT - adding ts-ignore here may prevent us from catching breaking API changes in the future. It might be nice to just pass in the 3 missing values:

```ts
          value={{
            initialized: false,
            config: {},
            pluginStore: new PluginStore(),
            api: {
              chrome: {
                getBundleData: () => ({
                  bundleTitle: 'bundle-title',
                }),
              },
            },
          }}
```

packages/chrome/src/ChromeProvider/ChromeProvider.tsx
Hyperkid123: @florkbr It's not really the same. The only way of being able to use the same logic would be making the observer callback async and I don't want to do that. It can cause some issues. The catch block you have highlighted is there solely for the purpose of catching this call 
```
scalprum.api.chrome.getBundleData().bundleTitle
```
Where as the top one is for the async call. Sure we could create some handle error method but I don't really see a point n wrapping a console.error statement in a function.

packages/chrome/src/ChromeProvider/ChromeProvider.tsx
Hyperkid123: No, and it was added just for testing because the MutationObserver does not exist in node env. It should be removed.

florkbr (approved):

PR #1943, opened 2023-12-04
Update getRefreshToken type

aneelac22:


packages/types/index.d.ts
```diff
@@ -125,7 +125,7 @@ export interface ChromeAPI {
   auth: {
     doOffline?: () => void;
     getOfflineToken: () => Promise<any>;
-    getRefreshToken: () => string;
+    getRefreshToken: () => Promise<string | undefined>;
```
Hyperkid123, line 128: The refresh token should be always defined. At least for KC instances.

Hyperkid123 (approved):

PR #1952, opened 2024-01-03
Use determinstic output file hashes based on git commit

adamrdrew:
We want to have hashes in filenames for cache busting, but this can cause problems with with containerized frontends and how they interact with akamai that can (and recently, have) result in outages. This patch makes the hashes based on the current git SHA instead of webpack's built-in full hash. This would ensure that if a given commit is rebuilt we will get the same files in the resultant build.

The reason we want this is we had an issue where a version of an app was rebuilt. When this new version of the build was deployed cache wasn't busted in akamai because we were using the same image tag. The resulted in the fed-mods.json in akamai and the file names on the pod to mismatch. This resulted in a production outage and customer impact in the form of a customer support case. If this patch had been in place the issue would never have happened as the resultant rebuild would have had the same filenames.

This patch shouldn't impact developers and the developer workflow because we're already conditional on hot reload. That said, this is a complex config and we've got lots of apps, so input is very much appreciated.

### DIFF
--- a/packages/config/src/lib/createConfig.ts
+++ b/packages/config/src/lib/createConfig.ts
@@ -2,6 +2,7 @@ const path = require('path');
 const fs = require('fs');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const searchIgnoredStyles = require('@redhat-cloud-services/frontend-components-config-utilities/search-ignored-styles');
+const childProcess = require('child_process');
 
 import { LogType, ProxyOptions, fecLogger, proxy } from '@redhat-cloud-services/frontend-components-config-utilities';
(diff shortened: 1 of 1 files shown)


Hyperkid123:
@adamrdrew can't do that yet. I'll switch to content hash soon. The new version of one of the dependencies we were waiting on was updated and we will be able to move forward now.

We have to start with upgrading our Scalprum dependency.

johnsonm325:
@adamrdrew @Hyperkid123 do we still want this? Is it worth closing, or is it ready to be implemented?

Hyperkid123:
We are already using content hashes which are deterministic. This can be closed.

PR #1954, opened 2024-01-08
create helper function to fetch data with concurrent requests

mkholjuraev:
This adds a simple, but useful helper function to utils package. It will help us to fetch a huge amount of data using paginated requests. Additionally, those paginated requests can be done in a certain number of concurrent API calls.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32651,6 +32651,31 @@
             "resolved": "https://registry.npmjs.org/ospath/-/ospath-1.2.2.tgz",
             "integrity": "sha512-o6E5qJV5zkAbIDNhGSIlyOhScKXgQrSRMilfph0clDfM0nEnBOlKlH4sWDmG95BW/CvwNz0vmm7dJVtU2KlMiA=="
         },
+        "node_modules/p-all": {
+            "version": "5.0.0",
+            "resolved": "https://registry.npmjs.org/p-all/-/p-all-5.0.0.tgz",
+            "integrity": "sha512-pofqu/1FhCVa+78xNAptCGc9V45exFz2pvBRyIvgXkNM0Rh18Py7j8pQuSjA+zpabI46v9hRjNWmL9EAFcEbpw==",
+            "dependencies": {
+                "p-map": "^6.0.0"
+            },
+            "engines": {
+                "node": ">=16"
+            },
+            "funding": {
+                "url": "https://github.com/sponsors/sindresorhus"
+            }
+        },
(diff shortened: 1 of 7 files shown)


Hyperkid123 (commented):
Do we need the `p-all` and `p-map` dependency? In this case, we are dealing with just a simple que.

I am also a bit curious about why you need to limit the number of concurrent requests at all?

mkholjuraev:
@Hyperkid123 thank you for having a look. I think they will be useful to avoid AKAMAI blocks when there are 10k items to be fetched. As you know if we do not limit concurrent requests, AKAMAI will block after a short time. Let me know if this is not needed

Hyperkid123:
> I think they will be useful to avoid AKAMAI blocks when there are 10k items to be fetched.

Oh, one more question. Why is it that the UI requires 10k of records? Rendering anything with 10k items (table for example) is never going to end up with good UX unless the view is virtualized.

mkholjuraev:
Yup, this is one of the major issues on the UI that we face. When for example, a customer selects all existing advisories to be remediated, there is too much data to be fetched on the UI. Another example is when customers export reports. Maybe, I am exaggerating a bit, but I have seen sometimes this indeed happens

mkholjuraev:
The ideal solution for the remediation issue would be to do changes in the architectural level, but it is requires a decent amount of effort. Discussions on this have been taken place even now.

Hyperkid123:
> The ideal solution for the remediation issue would be to do changes in the architectural level, but it is requires a decent amount of effort. Discussions on this have been taken place even now.

Yeah, I think the screens should be revisited then. I am not going to block your features, but I'd like it if the problem was properly sorted out at some point. These kinds of "optimizations" should never be required.

bastilian:
@Hyperkid123 One main area that we have been using these hooks or similar solutions are exports, like CSV, JSON and PDF.

mkholjuraev:
@Hyperkid123 just wanted to remind you about this.

Hyperkid123:
@mkholjuraev the CI is still failing.

mkholjuraev:
@Hyperkid123 fixed. PTAL

gkarat (approved):
LGTM. I think we can merge and release it now

gkarat (changes requested):
Oops, there are some conflicts. Can we resolve them? @mkholjuraev otherwise, everything looks good

mkholjuraev:
@gkarat thanks for having a look. Conflicts are resolved now

PR #1959, opened 2024-01-17
[RHCLOUD-30109] Batch updates for last-visited-pages

BlakeHolifield:
Migrates the `useLastVisitedPages` hook to only push batch updates to the API
* When the page is `hidden` for 20 seconds, the user's last 10 pages from localStorage will be saved in the DB
* Every 3 minutes, the user's last 10 pages from localStorage are saved in the DB
* In the event that localStorage is cleared, the state is set from the DB

### DIFF
--- a/packages/chrome/src/ChromeProvider/ChromeProvider.tsx
+++ b/packages/chrome/src/ChromeProvider/ChromeProvider.tsx
@@ -1,78 +1,75 @@
 import React, { useEffect, useRef, useState } from 'react';
 import { useScalprum } from '@scalprum/react-core';
-import { ChromeAPI } from '@redhat-cloud-services/types';
 import { useLocation } from 'react-router-dom';
-import debounce from 'lodash/debounce';
-
 import { ChromeContext } from '../ChromeContext';
 import chromeState, { LastVisitedPage, UserIdentity } from './chromeState';
 import { IDENTITY_URL, LAST_VISITED_URL, get, post } from '../utils/fetch';
 
 const getUserIdentity = () => get<UserIdentity>(IDENTITY_URL);
-const postDataDebounced = debounce(async (pathname: string, title: string, bundle: string) => {
-  const data = await post<LastVisitedPage[], { pathname: string; title: string; bundle: string }>(LAST_VISITED_URL, {
-    pathname,
-    title,
-    bundle,
-  });
-  return data;
-  // count page as visited after 5 second of being on the page
-  // should help limit number of API calls
-}, 5000);
 
-// FIXME: Use this hook once the issues with dead locking are resolved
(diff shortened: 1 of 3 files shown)


packages/chrome/src/ChromeProvider/ChromeProvider.tsx
```diff
@@ -1,78 +1,69 @@
...
-        if (typeof currentTitle === 'string' && currentTitle !== prevTitle) {
-          try {
-            prevTitle = currentTitle;
-            postData(pathname, currentTitle, scalprum.api.chrome.getBundleData().bundleTitle);
-          } catch (error) {
-            // catch sync errors
-            console.error('Unable to update last visited pages!', error);
-          }
+    // Save state from localStorage on an interval
+    setInterval(async () => {
```
Hyperkid123, line 38: You have to clear the interval in the useEffect cleanup phase. Currently this can cause memory leaks.

packages/chrome/src/ChromeProvider/ChromeProvider.tsx
```diff
@@ -1,78 +1,69 @@
...
+      await updateMostRecentPages();
```
Hyperkid123, line 39: No reason to await this call. We are not working with the return value.

packages/chrome/src/ChromeProvider/ChromeProvider.tsx
```diff
@@ -1,78 +1,69 @@
...
+          inactiveTimeout.current = undefined;
+        }
+      }
+    };
+    document.addEventListener('visibilitychange', handleVisibilityChange);
+
     return () => {
-      titleObserver?.disconnect();
-      postDataDebounced?.cancel();
+      document.removeEventListener('visibilitychange', handleVisibilityChange);
```
Hyperkid123, line 64: We should also clear the `inactiveTimeout.current` here if it exists.

packages/chrome/src/ChromeProvider/ChromeProvider.tsx
```diff
@@ -82,24 +73,32 @@ const useLastVisitedLocalStorage = (providerState: ReturnType<typeof chromeState
   const scalprum = useScalprum();
   const titleTarget = document.querySelector('title');
   useEffect(() => {
-    let titleObserver: MutationObserver | undefined;
-    let prevTitle: string | null;
-    const lastVisited = localStorage.getItem(LAST_VISITED_FLAG);
-    if (lastVisited) {
+    const lastVisitedLocal = localStorage.getItem(LAST_VISITED_FLAG);
+    const getInitialPages = async () => {
```
Hyperkid123, line 77: This function should be defined out of the effect.

Hyperkid123 (approved):

PR #1968, opened 2024-02-08
feat(utils): Align selectors, functions with PF5

gkarat:
This makes the cypress utilities consistent with the PF5 upgrades. Mainly selectors are affected, but so do the helping functions that use them.


Fewwy (approved):
Amazing! Great job!

gkarat:
Hmm, the automation doesn't upgrade the package for some reason

gkarat:
@RedHatInsights/platform-experience ^

PR #1969, opened 2024-02-12
fix(docs): Replace `_unstableHotReload` with `hotReload`

leSamo:
`_unstableHotReload` with `hotReload` has been deprecated according to

https://github.com/RedHatInsights/frontend-components/blob/4ef5accbd645b6a416f2e2bd31096b05771056ec/packages/config/src/lib/createConfig.ts#L46

and

https://github.com/RedHatInsights/frontend-components/blob/bbef2f308320b4b879020ac43f46a23d07df0574/packages/config/src/bin/webpack.plugins.ts#L10


Hyperkid123 (approved):
:bow:

PR #1973, opened 2024-02-22
fix(deps): Make rbac-client proper dependency

bastilian:
Both the `RBAC` and the `RBACHook`[0] modules import from `@redhat-cloud-services/rbac-client`.

Using the hook without having the `rbac-client` already as a dependency or installed in a project will cause a module not found error.

[0] https://github.com/RedHatInsights/frontend-components/blob/master/packages/utils/src/RBACHook/RBACHook.ts#L3

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43748,6 +43748,7 @@
             "version": "4.0.5",
             "license": "Apache-2.0",
             "dependencies": {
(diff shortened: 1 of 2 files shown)


Fewwy (approved):
LGTM

PR #1976, opened 2024-02-27
fix(cypress): fixing incorrect functions and imports

Fewwy:
I fixed a couple of incorrect imports and one of the filter functions

I am not sure that this is correct
cy.get(CONDITIONAL_FILTER).find('label').contains(value).parent().find('input[type=radio]').check();


gkarat (approved):
LGTM, thanks @Fewwy

PR #1978, opened 2024-03-06
chore(deps): upgrade redux version to ^9.1.0

mkholjuraev:
Upgrades react-redux version to 9.1.0. This is an important upgrade to make the package compatible with other apps that are using it and have never version of react-redux.


bastilian:
Maybe it would be better to stick to the same version as `insights-chrome`, which is at `8.1.3` at the moment.

https://github.com/RedHatInsights/insights-chrome/blob/master/package.json#L175

packages/notifications/package.json
```diff
@@ -33,11 +33,11 @@
         "prop-types": "^15.6.2",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
-        "redux": ">=4.2.0",
-        "react-redux": ">=7.2.9"
+        "redux": ">=5.0.1",
+        "react-redux": ">=9.1.0"
```
Hyperkid123, line 37: This could introduce conflicts with NPM and local versions of redux. I suggest loosening the version range even more, but the version range. `"^4.2.0 || ^5.0.0"` etc. Alternatively, we could use the "*" range but that would be probably way too much. This is specific to the notifications package as it is installed by almost every HCC service.

We will remove redux dependencies from the notifications package at some point. No free cycles any time soon though.

packages/utils/package.json
```diff
@@ -31,7 +31,7 @@
         "@patternfly/react-table": "^5.0.0",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
-        "react-redux": ">=7.0.0",
+        "react-redux": ">=9.1.0",
```
Hyperkid123, line 34: Same as the notifications package. Loosen the restriction rather than restricting it to latest versions.

mkholjuraev:
@Hyperkid123 I have tried to update the package version, but I am bumped into many more peer dependency issues across FEC packages. It seems it is better if the platform does this as you have a much better understanding of the needs. Closing this PR for now. Let me know if I can help anyhow.

PR #1986, opened 2024-03-20
fix(npm): npm audit and axios update because of vulnerabilities

Fewwy:
There is a CVE about axios version,
https://issues.redhat.com/browse/RHINENG-8636
 I think it would be better to update frontend components to the newer axios.
In addition to that I ran "npm audit fix"
Before
image

After
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -75,7 +75,7 @@
                 "@types/sanitize-html": "^2.6.2",
                 "@typescript-eslint/eslint-plugin": "^5.38.1",
                 "@typescript-eslint/parser": "^5.38.1",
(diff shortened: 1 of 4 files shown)


Hyperkid123 (commented):
@Fewwy have you tested the updates? Axios v1 introduced breaking changes and this change can affect negatively all users.

Fewwy:
@Hyperkid123 You mean by running it locally with other projects? No, not yet. How would it be better to approach this?

Hyperkid123:
Yeah, you have to run it locally. Otherwise can't verify the updates are working.

Another approach is releasing a major version, but we have some plans with the next major version of FEC packages in foreseeable future.

PR #1991, opened 2024-03-28
add eslint rule to suggest using shared axios instance

mkholjuraev:
This adds eslint rule to suggest using the shared axios instance for better error handling and also for error reporting. There are many applications that have Axios installed and do not have any integration with the interceptors that are used to handle and report errors. With this eslint rule that is not enforcing, this would hopefully be addressed


Hyperkid123 (commented):
Just one nit, can we name the file/rule to suggest it is meant for the insights bundle?

Other services do not benefit from the error reporting.

Hyperkid123 (approved):

PR #1992, opened 2024-03-28
fix(RHINENG-9046): report details remove Usechrome hook

Fewwy:
https://issues.redhat.com/browse/RHINENG-9046

### DIFF
--- a/packages/advisor-components/src/ReportDetails/ReportDetails.spec.ct.js
+++ b/packages/advisor-components/src/ReportDetails/ReportDetails.spec.ct.js
@@ -65,6 +65,7 @@ describe('report details: kba loaded', () => {
   });
 
   it('renders a loaded kba link', () => {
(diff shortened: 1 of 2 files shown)


packages/advisor-components/src/ReportDetails/ReportDetails.spec.ct.js
```diff
@@ -65,6 +65,7 @@ describe('report details: kba loaded', () => {
   });
 
   it('renders a loaded kba link', () => {
+    cy.mount(<ReportDetails {...{ ...props, kbaLoading: true, isProd: true }} />);
```
gkarat, line 68: You are calling cy.mount twice (one in beforeEach, and one in `it`). It's not healthy for the test, and we better keep only one mount command. Can we just add this prop in cy.mount in beforeEach above?

gkarat (approved):
lgtm! just one note on the tests, but overall looks good

Hyperkid123 (approved):

PR #2000, opened 2024-05-01
fix(BulkSelect): RHINENG-9831 - Don't override passed in children

bastilian:
Through the change to PF5 components a bug was introduced to the BulkSelect in the frontend components that overrides children passed in via the toggleProps. Since a number of applications use this to show a loading icon or the number of items selected, etc, we should not be overriding these when passed in.


johnsonm325 (approved):

adonispuente (approved):
LGTM! Thanks for also adding test 👍🏽

PR #2001, opened 2024-05-02
fix(fec): Make bundle analyzer optional feature

bastilian:
The webpack prod config requires webpack-bundle-analyzer, this makes it optional and adds webpack-bundle-analyzer as a peer dependency in case apps do want to enable bundle analysing.

### DIFF
--- a/packages/config/src/bin/prod.webpack.config.ts
+++ b/packages/config/src/bin/prod.webpack.config.ts
@@ -1,7 +1,7 @@
+const { fecLogger, LogType } = require('@redhat-cloud-services/frontend-components-config-utilities');
 import FECConfiguration from '../lib/fec.config';
 import config from '../lib/index';
 import commonPlugins from './webpack.plugins';
-const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');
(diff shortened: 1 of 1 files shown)


packages/config/src/bin/prod.webpack.config.ts
```diff
@@ -18,7 +18,7 @@ const { config: webpackConfig, plugins } = config({
 plugins.push(...commonPlugins, ...externalPlugins);
 
 const start = (env: { analyze?: string }): Configuration => {
-  if (env && env.analyze === 'true') {
+  if (BundleAnalyzerPlugin && env && env.analyze === 'true') {
```
Hyperkid123, line 21: I think we can just kick it out entirely. Plugins can be extended from within `fec.config.js` anyway. Can you add a deprecation warning to this condition? So we can remove it in next major release?

bastilian:
@Hyperkid123 Since we intend to remove it, I removed it again from the dependencies and just wrapped the config in a `try {} catch {}` to not fail builds in apps that don't have the package.

Hyperkid123 (approved):

PR #2007, opened 2024-05-18
refactor(fec): Fixes, improvements and moar options

bastilian:
This PR adds a (few) fix(es) and improvements to the `fec` command:

* **Enables `npm run start` commands for the config packages**
   Just for convenience. I got tired of manually building all the time.
* **Fixes the port setting**
   The `—port` flag was not fully working I thought that the `PORT` var was recognised by fec/webpack, but I realised it is something we have implemented in the apps’ fec config.
* **Adds an option to expose the “LOCAL_APPS” feature** 
  This is to make it more convenient to start an app adhoc with routes for another app, without having to use a ENV var. The code to build the routes has also been slightly refactored and now also allows to setup routes for `/api` as well via an `--api` option
* **Adds more cli flags for fec options**
   These are flags for options that we have often configured in `fec.config.js` via ENV vars to allow them to be variable then calling `fec`. For example, in many apps we use a `PROXY=true` env variable to toggle `useProxy` int the fec config.
* **Rewrites downloading insights-chrome’s index.html** 
  This is to try and make the `onBeforeSetupMiddleware` obsolete, or at least reduce its responsibilities to make migrating to `setupMiddleware` easier or even unnecessary. It's also a cleaner implementation using the HTMLWebpackPlugin. At least a tiny bit. The cloning of the repo is removed as well. We really only need the index.html file, but if we realise we need more we can easily restore this.
* **Cleans up logging** 
  Where I touched code I replaced the `console.log` calls with `fecLogger`. Even webpack now logs via `fecLogger`. For consistency. 
* **Adds an option to dump configs** 
  It has been very useful at times when debugging issues to inspect the fec config and webpack config.  This new option will make it more convenient without having to hack in some console.logs somewhere, and will show the full "fec" config passed in and the full resulting webpack config.

### DIFF
--- a/packages/config-utils/src/fec-logger.ts
+++ b/packages/config-utils/src/fec-logger.ts
@@ -1,10 +1,10 @@
 import chalk from 'chalk';
 
 const PREFIX = '[fec]';
-const ERROR_LEVEL = `${PREFIX} Error`;
-const WARN_LEVEL = `${PREFIX} Warn`;
-const INFO_LEVEL = `${PREFIX} Info`;
-const DEBUG_LEVEL = `${PREFIX} Debug`;
+const ERROR_LEVEL = 'Error';
+const WARN_LEVEL = 'Warn';
+const INFO_LEVEL = 'Info';
+const DEBUG_LEVEL = 'Debug';
 
 export enum LogType {
   error = ERROR_LEVEL,
@@ -28,9 +28,35 @@ const logFunctions = {
 };
 
 function getFecMessage(level: LogType, ...data: any[]) {
-  logFunctions[level](`${colors[level](level)}: `, ...data);
+  logFunctions[level](`${colors[level](PREFIX + ' ' + level)}: `, ...data);
 }
+
+export const fecWebpackLogger = () => {
+  function cleanUpLog(args: any) {
+    return args.map((log: any) => {
+      if (typeof log === 'string') {
+        return log.replace('[webpack-dev-server]', '').trim();
+      } else {
+        return log;
+      }
+    });
+  }
+
+  function log(level: LogType, ...data: any[]) {
+    logFunctions[level](`${colors[level]('[wds] ' + level)}: `, ...data);
+  }
+
+  return {
+    ...console,
+    error: (...args: any[]) => log(LogType.error, chalk.bold.red(...cleanUpLog(args))),
+    warn: (...args: any[]) => log(LogType.warn, ...cleanUpLog(args)),
+    info: (...args: any[]) => log(LogType.info, ...cleanUpLog(args)),
+    debug: (...args: any[]) => log(LogType.debug, ...cleanUpLog(args)),
+    log: (...args: any[]) => log(LogType.debug, ...cleanUpLog(args)),
+  };
+};
 
 export default getFecMessage;
 module.exports = getFecMessage;
 module.exports.LogType = LogType;
+module.exports.fecWebpackLogger = fecWebpackLogger;
--- a/packages/config/src/bin/dev-script.ts
+++ b/packages/config/src/bin/dev-script.ts
@@ -8,6 +8,55 @@ import serveChrome from './serve-chrome';
 import { LogType, fecLogger } from '@redhat-cloud-services/frontend-components-config-utilities';
 const DEFAULT_CHROME_SERVER_PORT = 9998;
 
+export const flags = (yargs: any) => {
+  yargs
+    .positional('webpack-config', {
+      type: 'string',
+      describe: 'Path to webpack config',
+    })
+    .option('port', {
+      type: 'number',
+      alias: 'p',
+      describe: 'Webpack dev server port',
+      default: 1337,
+    })
+    .option('proxy', {
+      type: 'boolean',
+      default: false,
+      describe: 'Enable proxying',
+    })
+    .option('apps', {
+      type: 'string',
+      describe:
+        'A coma seperated string of frontend applications with ports and optional protocol to create routes for (APP_NAME:APP_PORT[~APP_PROTOCOL])',
+    })
+    .option('apis', {
+      type: 'string',
+      describe:
+        'A coma seperated string of application APIs with ports and optional protocol to create routes for (APP_NAME:APP_PORT[~APP_PROTOCOL])',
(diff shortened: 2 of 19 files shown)


packages/config-utils/src/fec-logger.ts
```diff
@@ -28,9 +28,34 @@ const logFunctions = {
...
+      }
+    });
+  }
+
+  function log(level: LogType, ...data: any[]) {
+    logFunctions[level](`${colors[level]('[wds] ' + level)}: `, ...data);
+  }
+
+  return {
+    ...console,
```
bastilian, line 50: ... it works, but I'm a bit unsure if that's right.

packages/config/src/bin/webpack.plugins.ts
```diff
@@ -35,6 +36,10 @@ const plugins = [
 
 // Save 20kb of bundle size in prod
 if (process.env.NODE_ENV === 'production') {
+  // TODO This is questionable...
```
Hyperkid123, line 39: Another left over of stand alone. I'd just remove it.

packages/config/src/lib/createConfig.ts
```diff
@@ -3,29 +3,31 @@ const fs = require('fs');
...
 
-import { LogType, ProxyOptions, fecLogger, proxy } from '@redhat-cloud-services/frontend-components-config-utilities';
+import { LogType, ProxyOptions, fecLogger, fecWebpackLogger, proxy } from '@redhat-cloud-services/frontend-components-config-utilities';
 type Configuration = import('webpack').Configuration;
 type CacheOptions = import('webpack').FileCacheOptions | import('webpack').MemoryCacheOptions;
 type ProxyConfigArrayItem = import('webpack-dev-server').ProxyConfigArrayItem;
 type ClientConfiguration = import('webpack-dev-server').ClientConfiguration;
 type ResolveOptions = import('webpack').ResolveOptions;
 
+export type FrontendEnv = 'stage-stable' | 'prod-stable' | 'ci-stable' | 'qa-stable' | 'stage-beta' | 'prod-beta' | 'ci-beta' | 'qa-beta';
```
Hyperkid123, line 13: I think you can comfortably remove `ci` and `qa` envs. They are dead/

packages/config/src/lib/createConfig.ts
```diff
@@ -267,6 +282,7 @@ export const createConfig = ({
         disableDotRule: true,
       },
       devMiddleware: {
+        // TODO Figure out if this helps in any way or if it is required for something
```
Hyperkid123, line 285: Used to be able to modify the build files in dev env.

packages/config/src/lib/createConfig.ts
```diff
@@ -287,14 +303,12 @@ export const createConfig = ({
         proxyVerbose,
         target,
         registry,
-        onBeforeSetupMiddleware: ({ chromePath }) => {
```
Hyperkid123, line 290: This has to stay. Apps do not have their own HTML templates.

packages/config/src/lib/createPlugins.ts
```diff
@@ -11,12 +12,19 @@ const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin'
 const { glob } = require('glob');
 const path = require('path');
 
+const chromeIndexHtmlUrl = (env = 'dev-stable') => {
```
Hyperkid123, line 15: Can't use this. We will be removing the build repositories. Old build pipeline is coming to an end. We will be suing the container images as the source of truth for local chrome.

packages/config/src/lib/fec.config.ts
```diff
@@ -1,12 +1,13 @@
...
 import { CreateConfigOptions } from './createConfig';
 import { CreatePluginsOptions, WebpackPluginDefinition } from './createPlugins';
+type ProxyConfigArrayItem = import('webpack-dev-server').ProxyConfigArrayItem;
 
 export interface FECConfiguration
   extends Partial<Omit<CreateConfigOptions, 'appUrl' | 'appName' | 'env' | 'rootFolder'>>,
     Partial<Omit<CreatePluginsOptions, 'rootFolder' | 'appName'>> {
   appName: never;
-  appUrl: string | string[] | (string | RegExp)[];
+  appUrl?: string | string[] | (string | RegExp)[];
```
Hyperkid123, line 10: This is mandatory.

packages/config/src/lib/index.ts
```diff
@@ -51,9 +53,12 @@ const createFecConfig = (
   config: ReturnType<typeof createConfig>;
   plugins: ReturnType<typeof createPlugins>;
 } => {
+  // TODO ... sus.
   configurations.isProd = configurations.isProd || process.env.NODE_ENV === 'production';
   const isProd = configurations.isProd;
   const { insights } = require(`${configurations.rootFolder}/package.json`);
+
+  // TODO We should deprecated building based upon git branches
```
Hyperkid123, line 61: Yes. Remove only once all apps are on the current build pipeline.

packages/config/src/bin/dev-script.ts
```diff
@@ -67,18 +124,57 @@ async function devScript(
...
+      // TODO This is neat,
+      // but we should just default to stage-stable, which is what we need in dev 99% of the time.
       await setEnv(cwd);
     }
 
     if (argv.port) {
       process.env.PORT = argv.port;
     }
 
+    if (argv.proxy) {
```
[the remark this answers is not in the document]
bastilian: I've been thinking about this. I am worried about that too, but the issue is that due to the way `fec` spawns `webpack` some of these flags that are webpack( dev server) flags that we usually could adhoc append to `npm run start` (and `:proxy`) are now not usable anymore.

However, we could try to change the implementation and "inherit" webpack flags, as in we will pass along all flags that are not used/recognised by the `fec` command itself.

Should we go for that?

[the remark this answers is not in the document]
Hyperkid123: > However, we could try to change the implementation and "inherit" webpack flags, as in we will pass along all flags that are not used/recognised by the fec command itself.

Yes. I'd do this instead.

PR #2010, opened 2024-05-22
chore: set AsyncComponent height to full while loading

mkholjuraev:
This is a minor fix to properly lay the loading icon in the center of the page. Previously, due to missing height, the loading icon was displayed on top of the page. Ping me if screen video is needed to explain what this fix is

### DIFF
--- a/packages/components/src/AsyncComponent/index.tsx
+++ b/packages/components/src/AsyncComponent/index.tsx
@@ -3,7 +3,6 @@ import { ScalprumComponent, ScalprumComponentProps } from '@scalprum/react-core'
 import { Bullseye, Spinner } from '@patternfly/react-core';
 import classNames from 'classnames';
 import { ChromeAPI } from '@redhat-cloud-services/types';
-
 export type ExcludeModulesKeys = 'appName' | 'module' | 'scope';
 
 export interface AsyncComponentProps extends React.DetailedHTMLProps<React.HTMLAttributes<HTMLElement>, HTMLElement> {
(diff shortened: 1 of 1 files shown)


bastilian:
@mkholjuraev I'm not sure if the `AsyncComponent` should be concerned about that height.

We should much rather address the height issue where it occurs. The AsyncComponent does only append a class name to scope css, but it itself shouldn't apply any styling.

mkholjuraev:
@bastilian you may be right. I do not have a strong opinion on this. The thing is this component is used in many places and those pages/components also suffer from this problem. Also, I think this change only touches the height 100% style which is most probably expected for all the use cases. Thus, I thought this style may be an exception

bastilian:
@mkholjuraev We should be looking at where the loading state comes from and make sure that it expands to full height. The `AsyncComponent` is just concerned about loading a federated module. As part of it, it appends app specific CSS classes in some cases, but it should avoid changing the styling. It is more to think of like a `div`, that has no real styling by itself.

mkholjuraev:
The loading state is coming from here: https://github.com/RedHatInsights/frontend-components/blob/1cff3e121fd922fa9bb2204a4ba12058baab6d85/packages/components/src/AsyncComponent/index.tsx#L32 

I think the problem is that Bullseye does not stretch the height, but it only takes the inherited height that is not 100%. One more fix can be setting this height from the chroming app, the main content wrapper component (where we put our tenant apps). But, I am not sure if this would be a proper fix.

packages/components/src/AsyncComponent/index.tsx
```diff
@@ -49,7 +48,7 @@ const BaseAsyncComponent: React.FunctionComponent<BaseAsyncComponentProps> = ({
     ...props,
   };
   return (
-    <Cmp className={classNames(className, appName)}>
+    <Cmp className={classNames(className, appName, 'pf-v5-u-h-100')}>
```
Hyperkid123, line 51: This is not something we can. You can have multiple async components on a page and this will affect the overall layout of a page. You need to do this in your loading components that required the extra styling.

mkholjuraev:
Thank you both! Will fix the issue in those proper places

PR #2016, opened 2024-06-06
Add SASS prefix handler

JetyAdam:
Jira ticket: RHCLOUD-24774

### DIFF
--- a/packages/config/src/lib/addPrefixToContent.test.ts
+++ b/packages/config/src/lib/addPrefixToContent.test.ts
@@ -1,0 +1,51 @@
+import addPrefixToContent from './addPrefixToContent';
+
+describe('shouldnt prefix nested selectors', () => {
+  test('should add prefix above content', () => {
+    const content = `.foo {
+            .bar {
+                color: blue;
+            }
+        }`;
+    const prefix = '.learningResources, .learning-resources';
+    const output = `.learningResources, .learning-resources {
+            .foo {
+                .bar {
+                    color: blue;
+                }
+            }
+        }`.replace(/\s+/g, ``);
+    expect(addPrefixToContent(content, prefix).replace(/\s+/g, ``)).toEqual(output);
+  });
+
+  test('should handle multiple top-level selectors in one string', () => {
+    const content = `.learningResources-learningResources{overflow-y:auto}.widget-learning-resources{column-width:300px}`;
+    const prefix = '.learningResources, .learning-resources';
+    const output =
+      `.learningResources, .learning-resources { .learningResources-learningResources {overflow-y:auto}}.learningResources, .learning-resources { .widget-learning-resources {column-width:300px}}`.replace(
+        /\s+/g,
+        ``
+      );
+    expect(addPrefixToContent(content, prefix).replace(/\s+/g, ``)).toEqual(output);
+  });
+
+  test('should ignore nested selector containing prefix', () => {
+    const content = `.foo {
+        .bar {
+            .learningResources, .learning-resources {
+                    color: red;
+            }
+        }}`;
+    const prefix = `.learningResources, .learning-resources`;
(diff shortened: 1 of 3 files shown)


packages/config/src/lib/createConfig.ts
```diff
@@ -109,6 +109,41 @@ export const createConfig = ({
...
+        const exactPrefix = new RegExp(`^${prefix}(\\s|\\{|$)`);
+        if (exactPrefix.test(selector)) {
+          return false;
+        }
+      }
+      const exactAppNamePrefix = new RegExp(`^\\.${appName}(\\s|\\{|$)`);
+      return !exactAppNamePrefix.test(selector);
+    }
+
+    function addPrefixToSelector(selector: string): string {
```
Hyperkid123, line 127: I would create a constant and remove the `shouldPrependPrefix`

packages/config/src/lib/createConfig.ts
```diff
@@ -109,6 +109,41 @@ export const createConfig = ({
...
+      if (shouldPrependPrefix(selector)) {
+        const prefixes: string[] = sassPrefixes && sassPrefixes.length > 0 ? sassPrefixes : [`.${appName}`];
```
Hyperkid123, line 129: This can be removed once appName and sassPrefix is merged into one argument.

packages/config/src/lib/createConfig.ts
```diff
@@ -109,6 +109,41 @@ export const createConfig = ({
...
+        return prefixes.map((prefix: string) => `${prefix} ${selector}`).join(', ');
+      }
+      return selector;
+    }
+
+    // Process the content to add prefixes to all selectors
+    const prefixedContent = content.replace(/([^{}]+)\s*\{/g, (match, selectors) => {
```
Hyperkid123, line 136: This matches even nested rules.

Hyperkid123 (commented):
Could you extract te function to a separate file. Also write some tests for the input/output.

packages/config/src/lib/addPrefixToContent.ts
```diff
@@ -0,0 +1,40 @@
+//TODO - iterate through a string and save indexes of the curly parentheses [openingIndex, closingIndex]
```
Hyperkid123, line 1: Do you need the TODO?

packages/config/src/lib/addPrefixToContent.test.ts
Hyperkid123, line 1: I'd also add a mixed example. A content with rules that should be prefixed and rules that should not be prefixed.

And mix up the order of the rules.

Hyperkid123:
Add a custom matcher pealse to remove the replace for all the strings.

packages/config/src/lib/addPrefixToContent.ts
```diff
@@ -0,0 +1,40 @@
...
+// slice the selector before that, if it's top level selector, add prefix in a scope above, if its nested, dont care
+
+const addPrefixToContent = (content: string, sassPrefix: string): string => {
+  // Stack to store pairs of opening and closing curly brace indexes
```
bastilian, line 5: Maybe instead of building a custom parser, we can utilise sass' `compileString` function? https://sass-lang.com/documentation/js-api/functions/compilestring/

bastilian: Hm. I thought maybe compileString would allow to have a more structured representation of the css, but it's just a string too.

Hyperkid123: Oh, we don't want to compile, we will leave that to saas. We just want to not prefix css rules that match the app name. To allow styling the root element. I think is a manual process. I wish there was some nice AST for scss that is not 10 years old.

Hyperkid123 (approved):

PR #2018, opened 2024-06-14
fix(config): Remove obsolete fec config bits

bastilian:
This removes the `redux-promise-middleware` from being shared, as this might cause issues when incompatible versions of `(react-)redux` and `redux-promise-middleware`. 

It also removed the stubbing of `redux-logger` in prod. We should not need to do this, as `redux-logger` should not be imported in a prod asset in the first place. If it is, the apps should properly setup entry points for dev and prod.


AsToNlele (approved):
LGTM, tested on malware

Hyperkid123 (approved):

PR #2020, opened 2024-06-24
Bump @redhat-cloud-services/rbac-client to 2.0.3

ezr-ondrej:
Update rbac-client so we stop pulling old axios client.

This pulls mostly https://github.com/RedHatInsights/javascript-clients/pull/261

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8319,21 +8319,43 @@
                 "form-data": "^4.0.0"
             }
         },
-        "node_modules/@redhat-cloud-services/rbac-client": {
-            "version": "1.2.5",
-            "resolved": "https://registry.npmjs.org/@redhat-cloud-services/rbac-client/-/rbac-client-1.2.5.tgz",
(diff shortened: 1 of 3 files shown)


ezr-ondrej:
@RedHatInsights/platform-experience could you take a look? I dunno how to work with monorepos, happy to adjust this, but bumping this will IMO solve most repos axios vulnerabilities ;)

Hyperkid123 (commented):
This might be problematic as TS change on FEC side, can mean a requirement for a major version upgrade on the consumer size.


Can we loosen the version of the dependency? To both v1 and v2? Or we will have to release a major version of the affected packages.

ezr-ondrej:
Sure. Loose as this looks good?

Hyperkid123:
I'd say we want to keep the `^1.0.111` just to make sure. 2.x is OK for now. I the next major release (its coming) we will just use the 2.x

ezr-ondrej:
Gotcha, make sense. Done `^1.0.111 || 2.x`

Hyperkid123 (approved):

ezr-ondrej:
Right, I've missed the direct dependency: https://github.com/RedHatInsights/frontend-components/pull/2021 :face_exhaling:

ezr-ondrej:
I missed two, not one: #2022 :face_exhaling:

PR #2023, opened 2024-06-27
Update axios

ezr-ondrej:
Apparently I needed to run `npm update axios` to take effect in `package-lock.json` or did we need to release and only then npm update?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8328,16 +8328,6 @@
                 "tslib": "^2.6.2"
             }
         },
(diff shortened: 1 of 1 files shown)


ezr-ondrej:
I don't even know if this is needed, https://github.com/RHEnVision/provisioning-frontend/pull/430 looks like updating the dependencies helped and we do not this.

As you can see I've got no idea how npm dependency resolution works :stuck_out_tongue:

PR #2038, opened 2024-07-19
Add new chrome API boolean for ITLess

florkbr:


Hyperkid123:
Do we have to do this? Can't we just use feature flags? How much work it would be to do this instead? As far as I see this, its only a temporary variable. I don't want this here to stick for ever.

florkbr:
Closing as we should remove this logic and not just rename/update.

PR #2039, opened 2024-07-22
feat: Jest test wrapper component created, CypressTestUtils folder relocated

mkholjuraev:
This PR is intended to share our TestWrapper component that is used to render React components in Jest tests. I have also relocated the CypressTestUtils into the new TestUtils folder which is supposed to hold all test related utils. I made sure exports are fixed after relocation.


mkholjuraev:
@Fewwy I am tagging you as you were the author of CypressTestUtils. @gkarat This is the continuation of our discussion in the [PR for vulnerability](https://github.com/RedHatInsights/vulnerability-ui/pull/2124#discussion_r1675419960).

packages/utils/src/TestingUtils/JestUtils/TestWrapper.js
```diff
@@ -0,0 +1,33 @@
...
+import { Provider } from 'react-redux';
+import { MemoryRouter, Route, Routes } from 'react-router-dom';
+import { IntlProvider } from 'react-intl';
+
+const TestWrapper = ({ renderOptions = {}, children }) => {
+  const mockStore = configureStore();
+
+  return (
+    <IntlProvider locale="en">
+      <Provider store={renderOptions?.store || mockStore()}>
```
Hyperkid123, line 13: Any reason why the context props are nested in the `renderOptions` object?

mkholjuraev: There is no firm intention. This made this component as dynamic as possible so we do not have to pass all the options. Rather according to rendering needs, we can pass render options and the component renders declaratively. I am open to any improvements, feel free to share your ideas :)

Hyperkid123: > This made this component as dynamic as possible so we do not have to pass all the options

Not sure if I follow. Props can be optional. I know its a nitpick, but I'd just make the props flat. I can understand props nesting when a specific prop encompasses props if a child component and you just want to spread them, but this seems unnecessary.

gkarat: I think Martin's comment makes sense: we can extract `store`, `initialEntries`, some other keys to separate props for TestWrapper. This way we don't need to look up for `renderOptions` first to find the needed values.

gkarat: We actually have the props more "plain" in Inventory: https://github.com/RedHatInsights/insights-inventory-frontend/blob/e2123605436df6b8c84de8c39711d0be2a5a142a/src/Utilities/TestingUtilities.js#L9

packages/utils/src/TestingUtils/JestUtils/TestWrapper.js
```diff
@@ -0,0 +1,33 @@
...
+        <MemoryRouter initialEntries={renderOptions?.initialEntries || ['/']}>
```
gkarat, line 14: MemoryRouter already has `initialEntries` defaulting to []: https://reactrouter.com/en/main/router-components/memory-router.

Let's also extract props for MemoryRouter and do simple spread like this:
```suggestion
        <MemoryRouter {...routerProps}>
```

packages/utils/src/TestingUtils/JestUtils/TestWrapper.js
```diff
@@ -0,0 +1,33 @@
...
+          {renderOptions?.componentPath ? (
+            <Routes>
+              <Route path={renderOptions?.elementPath || '/'}>{children}</Route>
```
gkarat, line 17: Shouldn't it be using `componentPath` instead? I would also restrain from giving any route (even /) if it's not explicitly given, just render the component without route wrapping:

```suggestion
              <Route path={renderOptions.componentPath}>{children}</Route>
```

packages/utils/src/TestingUtils/JestUtils/TestWrapper.js
```diff
@@ -0,0 +1,33 @@
...
+          )}
+        </MemoryRouter>
+      </Provider>
+    </IntlProvider>
+  );
+};
+
+TestWrapper.propTypes = {
+  children: PropTypes.element,
+  renderOptions: PropTypes.object,
```
gkarat, line 30: Let's describe the props more in detail with `PropTypes.shape(...)` so that the code is better documented.

mkholjuraev:
@Hyperkid123 @gkarat PTAL, I forgot about this PR. Now, it has all the necessary changes.

packages/utils/src/TestingUtils/JestUtils/TestWrapper.js
```diff
@@ -5,29 +5,31 @@ import { Provider } from 'react-redux';
...
-            </Routes>
-          ) : (
-            children
-          )}
-        </MemoryRouter>
-      </Provider>
-    </IntlProvider>
-  );
-};
+const TestWrapper = (children, routerProps = { initialEntries: ['/'] }, componentPath, store = mockStore()) => (
```
Hyperkid123, line 10: I think the props should be wrapped in an object.

mkholjuraev: Ah, yes!

mkholjuraev: Fixed!

gkarat (approved):
LGTM, thanks @mkholjuraev

PR #2046, opened 2024-08-09
fix(downloadFile): Use 'csv' value as a default format

gkarat:
No jira.

### DIFF
--- a/packages/utils/src/helpers/helpers.ts
+++ b/packages/utils/src/helpers/helpers.ts
@@ -11,7 +11,7 @@ export function mergeArraysByKey(arrays: Record<string, unknown>[][], key = 'id'
   return Object.values(mergedObject);
 }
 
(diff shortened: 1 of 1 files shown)


packages/utils/src/helpers/helpers.ts
```diff
@@ -11,7 +11,7 @@ export function mergeArraysByKey(arrays: Record<string, unknown>[][], key = 'id'
   return Object.values(mergedObject);
 }
 
-export function downloadFile(data: unknown, filename = `${new Date().toISOString()}`, format = CSV_TYPE) {
+export function downloadFile(data: unknown, filename = `${new Date().toISOString()}`, format = 'csv') {
```
bastilian, line 14: ;) https://github.com/RedHatInsights/frontend-components/blob/master/packages/utils/src/helpers/helpers.test.ts#L58

Hyperkid123 (approved):

PR #2051, opened 2024-08-15
Add quay fallback for chrome image sha.

Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-34649

TLDR, GH api rate limiting can sometimes prevent us from getting the commit. We will get latest sha from quay. It might not match the current latest version, but better than nothing.


aneelac22 (approved):

leSamo (approved):
Thank you, the issue is resolved for me

PR #2063, opened 2024-09-12
feat(devtool): Ability to pass custom devtool option to config

adonispuente:
Because apps now use fec as the default config, I want to add the ability to pass in a devtool option. With this we can pass in 'hidden-source-map' for apps that will use sentry. Eventually I want this to just be the default value but I want to use advisor as a test application first.
Now that more of the sentry efforts are unblocked this can move along much faster.

https://github.com/RedHatInsights/insights-advisor-frontend/pull/1268 related PR

### DIFF
--- a/packages/config/src/lib/createConfig.ts
+++ b/packages/config/src/lib/createConfig.ts
@@ -97,7 +97,7 @@ export const createConfig = ({
   nodeModulesDirectories = [],
   stripAllPfStyles = false,
   blockLegacyChrome,
-  devtool,
+  devtool = false,
 }: CreateConfigOptions): Configuration => {
   if (typeof _unstableHotReload !== 'undefined') {
     fecLogger(LogType.warn, `The _unstableHotReload option in shared webpack config is deprecated. Use hotReload config instead.`);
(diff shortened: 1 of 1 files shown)


packages/config/src/lib/createConfig.ts
```diff
@@ -96,6 +97,7 @@ export const createConfig = ({
   nodeModulesDirectories = [],
   stripAllPfStyles = false,
   blockLegacyChrome,
+  devtool,
```
Hyperkid123, line 100: Can we do `devtool = false` to follow other configs with default values?

adonispuente: yeah sure thing, just sent this up

packages/config/src/lib/createConfig.ts
```diff
@@ -117,7 +119,7 @@ export const createConfig = ({
   const devServerPort = typeof port === 'number' ? port : useProxy || standalone ? 1337 : 8002;
   return {
     mode: mode || (isProd ? 'production' : 'development'),
-    devtool: false,
+    devtool: devtool || false,
```
Hyperkid123, line 122: The default value is set in the argument now you can simply use `devtool: devtool,` or just `devtool,`.

adonispuente: I got a bit too caught up in the meeting, thank you!

Hyperkid123 (approved):

PR #2066, opened 2024-09-19
chore(deps): Correct utils package deps

bastilian:
react-intl and redux mock store are used in the `TestWrapper` component and should be proper dependencies.

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -43,13 +43,13 @@
         "commander": "^2.20.3",
         "mkdirp": "^1.0.4",
         "p-all": "^5.0.0",
-        "react-content-loader": "^6.2.0"
+        "react-content-loader": "^6.2.0",
+        "react-intl": "^6.6.8",
+        "redux-mock-store": "^1.5.4"
     },
     "devDependencies": {
         "@types/react": "^18.0.0",
-        "glob": "10.3.3",
-        "react-intl": "^6.6.8",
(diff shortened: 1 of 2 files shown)


packages/utils/package.json
```diff
@@ -43,13 +43,13 @@
         "commander": "^2.20.3",
         "mkdirp": "^1.0.4",
         "p-all": "^5.0.0",
-        "react-content-loader": "^6.2.0"
+        "react-content-loader": "^6.2.0",
+        "react-intl": "^6.6.8",
+        "redux-mock-store": "^1.5.4"
```
mkholjuraev, line 48: should these packages be direct dependency even if we use them for testing purposes?

bastilian: @mkholjuraev If they are imported and used they should be dependencies. If we could expect that all apps, always have these dependencies as well, we should still add them as dependencies, but in that case we'd make them "peerDependencies". 

However, we can not expect our apps to really always, for sure, to have these dependencies and we are not making it possible to use the Wrapper without these dependencies. So we should add them, or apps may unnecessarily have to add these as a dependency themselves, which we should avoid.

mkholjuraev: I am just concerned that some devDepencies in nature are being added as direct dependency. Honestly, I do not have a clear understanding of the problem stated thus, I would like to request someone else as a reviewer.

PR #2071, opened 2024-10-01
Allow spdy to be used as webpack dev server.

Hyperkid123:
### Changes
- update webpack and its dependencies to latest versions
- enable setting the `_unstableSpdy` option from fec.config.js to enable spdy as a dev server


fhlavac (approved):

packages/config/src/lib/createPlugins.ts
```diff
@@ -2,7 +2,7 @@ import { LogType, fecLogger } from '@redhat-cloud-services/frontend-components-c
 import { Compiler, WebpackPluginInstance } from 'webpack';
 import { CommonConfigOptions } from './createConfig';
 
-const { SourceMapDevToolPlugin } = require('webpack');
+import { SourceMapDevToolPlugin } from 'webpack';
```
bastilian, line 5: The constructor below was still raising an error when starting webpack. Changing this import to just `import webpack form 'webpack'` and the constructor below to `new webpack.SourceMapDevToolPlugin(` seemed to solve the issues.

bastilian (approved):
LGTM! I only had to change the Sourcemap thingy, maybe that needs some attention still, but I tested Compliance with the `_unstableSpdy` option and everything is zooooomin, aka it's fast, so fast! Thank you for looking into it @Hyperkid123!

packages/config/src/lib/createPlugins.ts
Hyperkid123: OK let me change it.

PR #2079, opened 2024-10-14
Revert "fix: typo in version check"

mkholjuraev:
Reverts RedHatInsights/frontend-components#2076

We are seeing this error on our local environment provided. This affecting our workflow. Thus, I am opening this revert change. I believe there should be more investigation to fix the actual issue in #2076.

image


Hyperkid123 (approved):

Hyperkid123:
I think we are dealing with the mismatched versions of the SDK package. We should revert the changes and make sure there are correct version numbers in the consuming package.

PR #2083, opened 2024-10-21
fix(RHINENG-13796): Fix vertical alignment Inv tags badge

LightOfHeaven1994:
before:
image

after:
image


mkholjuraev (approved):
LGTM! Tested locally, alignements look good

PR #2084, opened 2024-10-21
fix(RHINENG-8385): Make group filter look correct

LightOfHeaven1994:
Fixed size of group filter and adapted it to `typeahead` variant when search is available
Before:
image

After:
image

### DIFF
--- a/packages/components/src/ConditionalFilter/GroupFilter.tsx
+++ b/packages/components/src/ConditionalFilter/GroupFilter.tsx
@@ -13,10 +13,11 @@ import {
   Radio,
   TextInputGroup,
   TextInputGroupMain,
+  TextInputGroupUtilities,
   TreeView,
   TreeViewDataItem,
 } from '@patternfly/react-core';
-import { CloseIcon } from '@patternfly/react-icons';
+import { TimesIcon } from '@patternfly/react-icons';
 import {
   FilterMenuItemOnChange,
   GroupFilterItem,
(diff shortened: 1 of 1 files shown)


packages/components/src/ConditionalFilter/GroupFilter.tsx
```diff
@@ -155,9 +156,14 @@ const GroupFilter: React.FunctionComponent<GroupFilterProps> = (props) => {
   };
 
   const handleClickOutside = (event: MouseEvent) => {
-    if (isOpen && !menuRef.current?.contains(event.target as Node)) {
-      setIsOpen(false);
-    }
+    if (!isOpen) return;
+
```
adonispuente, line 160: Nit pick, could you add a comment here just explaining that 
`//If you click insight the input or menu, function does nothing, otherwise, close the menu`

This is more so just if someone doesnt understand what this is checking. I havent seen code similar to this in any of the repos

adonispuente (approved):
LGTM, just left a comment about adding a note but its not necessary.

packages/components/src/ConditionalFilter/GroupFilter.tsx
LightOfHeaven1994: @adonispuente I was thinking about it and will follow approach that when code is clear there is no need to comment it. Having there clear variable names like `clickedInsideInput` and `setIsOpen(false)` also directly tells what's going on.

PR #2064, opened 2024-09-12
pageheader fix

epwinchell:
- Make FlexItem fill the available space

https://issues.redhat.com/browse/RHCLOUD-34965

Old
image-2024-09-03-13-26-23-653
New
Screenshot 2024-09-12 at 14 15 52

PR #2088, opened 2024-10-22
Fix: PageHeader component

epwinchell:
Jira: https://issues.redhat.com/browse/RHCLOUD-34965

Follow-up to: https://github.com/RedHatInsights/frontend-components/pull/2064

Replace FlexItem with div so that the previously added class takes effect.

Old
Screenshot 2024-10-22 at 2 58 56 PM

New
Screenshot 2024-10-22 at 2 59 09 PM


fhlavac (commented):
@epwinchell just curious, was there any issue with using the ContentHeader from component-groups instead?

epwinchell:
@fhlavac No, but this is marked as a major issue and a simple fix. I thought I would revisit using ContentHeader afterward.

fhlavac:
@epwinchell sounds good! Thanks Eric!

fhlavac (approved):

PR #2094, opened 2024-10-30
fix(RHINENG-8385): Unblock opening menu by toggle click

LightOfHeaven1994:
This is going to help with preventing opening menu when toggle is clicked even for the first time. We need to consider it as we are clicking inside and do nothing


gkarat (approved):
LGTM, thanks @LightOfHeaven1994

PR #2126, opened 2024-12-02
fix(fec): fix missing template.json when running fec dev

florkbr:
When running locally with the newly published version of FEC via our NX publish pipeline, we get the error:

image

This should resolve that issue


apinkert (approved):
LGTM 👍

florkbr:
Fixes https://issues.redhat.com/browse/RHCLOUD-36557

PR #2131, opened 2024-12-06
feat(deps): upgrade PatternFly to v6

wise-king-sullyman:
Upgrades this repo to use PatternFly 6 as part of the process of adding PF6 support to the chrome.


Hyperkid123:
Just a note. This will be a breaking change and we will be publishing a new major version.

Hyperkid123:
Merge this to the pf6 branch and will make a pre-release version(s) of the affected packages.

PR #2143, opened 2025-01-13
feat(config-utils): plugin feo nav interceptor

Hyperkid123:
Prepares the proxy for intercepting the messages. We need to enable the FEO-generated navigation first. That is handled separately.


Hyperkid123:
Need to adjust to consume the generated assets instead of regular config files

PR #2146, opened 2025-01-14
fix(deps): use latest version of fec package

bastilian:
Same as https://github.com/RedHatInsights/frontend-components/pull/2145 but for all packages.


bastilian:
@Hyperkid123 There are a few other packages also still referring to the v4 package versions. Should we bump these too?

bastilian:
@Hyperkid123 All the `package.json`s are now updated to refer to the latest version(s).

Hyperkid123 (approved):

PR #2147, opened 2025-01-15
fix(tagmodal): revert the PF5/Modal import path change

gkarat:
This makes the TagModal and related components use the standard PF5/Modal import path. It was changed to use the "next" version of PF5/Modal and I believe it was not an intended change since it breaks the integration with other apps that still use the parameters for the standard version of PF5/Modal.


bastilian (approved):
LGTM! Tested with Inventory and all looks good again! Thank you @gkarat!

PR #2148, opened 2025-01-16
fix(config-utils): fix 403s when proxying to stage for fed modules in chrome

florkbr:
I believe this bug was introduced during some of our refactoring in support of NX.

While testing https://github.com/RedHatInsights/insights-chrome/pull/2990 locally with @wise-king-sullyman - we were seeing 403's targeting `stage-beta` or `stage-stable` via our `webpack.config.js` proxy settings.

Upon deeper investigation I found that our config utils package was no longer passing the proxy `agent` key through the config object - resulting in these requests throwing 403's with the squid proxy access denied HTML as the response.

### Before:
Screenshot 2025-01-15 at 6 50 49 PM
Screenshot 2025-01-15 at 7 03 44 PM

### After:
Screenshot 2025-01-15 at 6 53 02 PM

### DIFF
--- a/packages/config-utils/src/proxy.ts
+++ b/packages/config-utils/src/proxy.ts
@@ -241,6 +241,9 @@ const proxy = ({
 
         return null;
       },
+      ...(agent && {
+        agent,
+      }),
     });
   }
 


florkbr:
Resolves https://issues.redhat.com/browse/RHCLOUD-37345

florkbr:
FWIW I don't think we've picked up on this being an issue as insights-chrome master is currently targeting config-utils `^3.0.7` (https://github.com/RedHatInsights/insights-chrome/blob/master/package.json#L59). The PF6 PR pulls in `^4.0.4`.

Hyperkid123 (approved):

PR #2149, opened 2025-01-16
fix(config-utils): 403 cors - headers and origin fix for stage env

florkbr:
@bastilian can you see if this resolves your issue? Thanks!


bastilian (approved):
Nice! This seems to fix the issue! Thank you @florkbr!

florkbr:
Related to https://github.com/RedHatInsights/frontend-components/pull/2148

PR #2150, opened 2025-01-17
feat: add single select filter component

johnsonm325:
This PR adds a single select filter dropdown component for use with the PrimaryToolbar. It is used on conjuction with the custom conditional filter type. Docs are included in the PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36617,7 +36617,7 @@
     },
     "packages/advisor-components": {
       "name": "@redhat-cloud-services/frontend-components-advisor-components",
-      "version": "2.0.8",
+      "version": "2.0.9",
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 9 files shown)


bastilian:
We should add an entry to the Conditional Filter[0] in order for it to be easily usable by just changing the type of "radio" to (maybe) "singleSelect". For that we should also make sure that the component complies to the same API as the other filters in the Conditional Filter[1].

[0] https://github.com/RedHatInsights/frontend-components/blob/master/packages/components/src/ConditionalFilter/conditionalFilterConstants.ts#L7-L21
[1] https://github.com/RedHatInsights/frontend-components/blob/master/packages/components/doc/conditionalFilter.md

bastilian (approved):
LGTM! Awesome work! Thank you @johnsonm325!

PR #2154, opened 2025-01-22
fix(Shield): remove Unknown question mark icon

xmicha82:
Removed QuestionIcon from Shield component as per [RHINENG-12939](https://issues.redhat.com/browse/RHINENG-12939)

### DIFF
--- a/packages/components/src/Shield/Shield.tsx
+++ b/packages/components/src/Shield/Shield.tsx
@@ -11,10 +11,17 @@ export interface ShieldProps {
   impact: keyof typeof impactList | 'N/A';
   hasLabel?: boolean;
   hasTooltip?: boolean;
+  disableQuestionIcon?: boolean;
   size?: IconComponentProps['size'];
 }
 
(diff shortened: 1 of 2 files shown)


packages/components/src/Shield/Shield.tsx
```diff
@@ -22,7 +21,12 @@ const Shield: React.FunctionComponent<ShieldProps> = ({ impact = 'N/A', hasLabel
     color: attributes.color,
   };
 
-  const badge = <Icon size={size}>{attributes.title === 'Unknown' ? <QuestionIcon {...badgeProps} /> : <SecurityIcon {...badgeProps} />}</Icon>;
```
bastilian, line 25: We might want to add a prop to show an icon and which icon, since there might be places where we actually do want to show an icon. 

The default behaviour should still remain the same.

johnsonm325 (changes requested):
Please update the documentation for this component as well: https://github.com/RedHatInsights/frontend-components/blob/master/packages/components/doc/shield.md

johnsonm325 (approved):

PR #2165, opened 2025-02-06
chore(deps): add more major versions to react-redux

AsToNlele:
Some of our apps (eg Advisor) use newer versions of Redux


AsToNlele:
there's a pr already 💀

PR #2168, opened 2025-02-11
fix(components): bulkselect ouia id

AsToNlele:
one more fix,
in the same file the `data-ouia-component-id` works for `<DropdownList data-ouia-component-id={listOuiaId ?? 'BulkSelectList'}>`
so left it like it is


AsToNlele:
Closing because of #2169

PR #2171, opened 2025-02-19
fix(Shield): fix Shield spacing without QuestionIcon

xmicha82:
Adjusted Shield to not render Icon container when question icon is disabled

### DIFF
--- a/packages/components/src/Shield/Shield.tsx
+++ b/packages/components/src/Shield/Shield.tsx
@@ -29,11 +29,18 @@ const Shield: React.FunctionComponent<ShieldProps> = ({
     color: attributes.color,
   };
 
-  const badge = (
-    <Icon size={size}>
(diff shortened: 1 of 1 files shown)


AsToNlele (commented):
Now the icon is smaller, is this expected?

johnsonm325:
This is not intended. Definitely needs to be fixed. Not sure how it's happening.

packages/components/src/Shield/Shield.tsx
```diff
@@ -29,11 +29,18 @@ const Shield: React.FunctionComponent<ShieldProps> = ({
...
-  );
+  const badge =
+    attributes.title === 'Unknown' ? (
+      disableQuestionIcon ? null : (
+        <Icon size={size}>
+          <QuestionIcon {...badgeProps} />
+        </Icon>
+      )
+    ) : (
+      <Icon>
```
johnsonm325, line 40: Add the size in here as well.
`<Icon size={size}>`

johnsonm325 (changes requested):
@xmicha82 ahh, this is the issue.

johnsonm325 (approved):

PR #2172, opened 2025-02-20
fix(components): groupFilterConstants remove key

AsToNlele:
When running Inventory locally and opening OS filter dropdown, I'm getting an error `Cannot read properties of undefined (reading 'getStackAddendum')`, after debugging the error is caused by this additional key.


bastilian (approved):
LGTM! Thank you @AsToNlele!

PR #2188, opened 2025-03-31
Fix: "TypeError: can't access property "setExtraStackFrame", ReactDebugCurrentFrame$1 is undefined" in Conditional filter

Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-38925

The custom conditional filter type is a React.Fragment and we can only pass key and children props t a fragment.

### DIFF
--- a/packages/components/src/ConditionalFilter/ConditionalFilter.tsx
+++ b/packages/components/src/ConditionalFilter/ConditionalFilter.tsx
@@ -153,7 +153,9 @@ const ConditionalFilter: React.FunctionComponent<ConditionalFilterProps> = ({
       );
     } else if (activeItem.type === 'custom' && identifyComponent<Record<string, any>>(activeItem.type, activeItem.filterValues)) {
       const C = typeMapper.custom;
-      return <C {...activeItem.filterValues} />;
+      // make sure no invalid props are passed to the Fragment element which is mapped to the custom component
+      const { key, children } = activeItem.filterValues;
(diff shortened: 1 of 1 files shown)


karelhala (commented):
I've had similar issue for `text` filter as well. Might be worth taking the key out from filterItems and passing both remainder of filerValues and key to all filter types.

bastilian:
@Hyperkid123 While this might fix the issue with the ConditionalFilter, the error is probably gonna be raised by any part that passes incorrect props to React.Fragment. I still believe that the root cause is related to the mode react assets are built with and probably the mix of prod and dev assets. Usually in development passing unknown props in React will log an error, I believe that this is failing here and raising the error, because it expects certain functions that would be available when React is built in development mode.

Hyperkid123:
@karelhala I don't think we want to do this one by one right now; it a symptom of a larger problem. A long term solution is described in the ticket.

bastilian:
@Hyperkid123 Sorry, I did't see the Jira comment till now. :)

Hyperkid123:
@bastilian Yeah I think its pretty clear what the issue is. I've talked to @florkbr, and I hope the services team will builda  dev container soon. If Konflux envs allows it (I hope it will). My intention is to inbloc your development specific to the workspaces.

justinorringer (approved):
LGTM as a quick fix! Thank you for figuring it out Martin :)

PR #2195, opened 2025-04-12
feat(config-utils): use the module interceptor when using local proxy

florkbr:
We've had some confusion around not being able to preview changes to fed-modules locally. Curious on thoughts about enabling the modules interceptor.


Hyperkid123 (approved):

Hyperkid123:
Oh yeah, I probably forgot to add it. Should have been there from the start.

PR #2202, opened 2025-04-22
fix: make webpack/http-server bin lookup compatible with monorepo setup

rawagner:
Let `npm` figure out the correct path for webpack/http-server binaries


Hyperkid123:
Hello. Yup, I'll add it to my queue today. I think it should be fine. I thought we had this in place before but removed it (might be wrong there is too many things in my head). I just have to verify it does not break some other logic.

Hyperkid123 (approved):
Oh NVM we forgot to use it here as we do use the exec in other places.

PR #2205, opened 2025-04-29
fix(config): add fec support for using dev chrome image

florkbr:
Changes:
* When proxying to stage - the new dev chrome image will be used
* The `isChrome` fec config boolean is now deprecated (defaults to `true` now)
* We will attempt to copy the chrome index.html from both the src (old) and srv (new) container locations (with logging)

@Hyperkid123 can you PTAL? We now have a development build of chrome pushed to https://quay.io/repository/redhat-services-prod/hcc-platex-services-tenant/insights-chrome-dev

~I tried pulling these changes into landing-page-frontend. The build works locally - but I believe since chrome will now request the development assets/hashes - there are errors requesting chrome-root etc.~ **Edit** fixed in second commit

### DIFF
--- a/packages/config-utils/src/proxy.ts
+++ b/packages/config-utils/src/proxy.ts
@@ -103,6 +103,7 @@ export type ProxyOptions = {
   target?: string;
   keycloakUri?: string;
   registry?: ((...args: any[]) => void)[];
+  /** @deprecated Is now turned on by default */
   isChrome?: boolean;
   onBeforeSetupMiddleware?: (opts: { chromePath?: string }) => void;
   bounceProd?: boolean;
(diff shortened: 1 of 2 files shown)


Hyperkid123:
I'll take a look. it might be just some stale index file.

Hyperkid123:
OK, so the issue is actually in the proxy. It does not proxy to the copied HTML template nor to the container. There is an "option" to force it that we are using in Chrome today, but I want to see if there is a better way.

Hyperkid123:
@florkbr I've added a commit with a change that fixes the issue. Feel free to do whatever with it you feel is best.

Hyperkid123:
While we are messing around with this, we might also want to try and copy the index from the new container location so we can start migrating chrome sooner to build-tools.

florkbr:
Tested changes in insights-inventory-frontend @bastilian. The minified react error 130 no longer crashes the page - and we see the correct dev error in the console (when running against stage).

For now I am merging - if we need to revert let me know ASAP.

PR #2211, opened 2025-05-15
feat: add support for proxying to a custom ephemeral environment

riccardo-forina:
This adds an `ephemeral` option to the list of environments supported by `fec dev`. It will prompt for an url, which is going to be the same returned by `bonfire namespace describe`. You'll know it works because the provided url will be printed in the output decribing the proxy config.

[asciicast](https://asciinema.org/a/KQakKwIMwliR83sosPZWBtott)

### DIFF
--- a/packages/config/src/bin/dev-script.ts
+++ b/packages/config/src/bin/dev-script.ts
@@ -15,11 +15,20 @@ async function setEnv(cwd: string) {
         type: 'list',
         name: 'clouddotEnv',
         message: 'Which platform environment you want to use?',
-        choices: ['stage', 'prod', 'dev'],
+        choices: ['stage', 'prod', 'dev', 'ephemeral'],
       },
     ])
-    .then((answers) => {
+    .then(async (answers) => {
       const { clouddotEnv } = answers;
+
+      if (clouddotEnv === 'ephemeral') {
+        const answer = await inquirer.prompt([{
+          type: 'input',
+          name: 'clouddotEnv',
+          message: 'Please provide the gateway route of your ephemeral environment:',
+        }]);
+        process.env.EPHEMERAL_TARGET = answer.clouddotEnv
+      }
       process.env.CLOUDOT_ENV = clouddotEnv ? clouddotEnv : 'stage';
       process.env.FEC_ROOT_DIR = cwd;
     });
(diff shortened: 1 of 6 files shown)


packages/config/src/bin/dev-script.ts
```diff
@@ -15,12 +15,23 @@ async function setEnv(cwd: string) {
...
-      process.env.CLOUDOT_ENV = clouddotEnv ? clouddotEnv : 'stage';
+
+      if (clouddotEnv === 'ephemeral') {
+        const answer = await inquirer.prompt([{
+          type: 'input',
+          name: 'clouddotEnv',
+          message: 'Please provide the URL of your ephemeral environment:',
+        }]);
+        process.env.EPHEMERAL_TARGET = answer.clouddotEnv
+        process.env.CLOUDOT_ENV = 'stage';
```
florkbr, line 31: @riccardo-forina are we intentionally setting "stage" here? Shouldn't this be ephemeral as well?

packages/config/src/bin/dev-script.ts
```diff
@@ -15,12 +15,23 @@ async function setEnv(cwd: string) {
...
-    .then((answers) => {
+    .then(async (answers) => {
       const { clouddotEnv } = answers;
-      process.env.CLOUDOT_ENV = clouddotEnv ? clouddotEnv : 'stage';
+
+      if (clouddotEnv === 'ephemeral') {
+        const answer = await inquirer.prompt([{
+          type: 'input',
+          name: 'clouddotEnv',
+          message: 'Please provide the URL of your ephemeral environment:',
```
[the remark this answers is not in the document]
riccardo-forina: Right, it’s the console UI one. Do we have a name for it that’s well understood by everybody?

packages/config/src/bin/dev-script.ts
riccardo-forina: Uhm i was being lazy. If we use ephemeral we end up with ephemeral.foo.redhat.com. I think we’ll have to update the hosts file (and the patch script) and I’m not 100% sure chrome will be happy about it

riccardo-forina: I’ll try tomorrow tho

packages/config/src/bin/dev-script.ts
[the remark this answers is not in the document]
riccardo-forina: "Gateway route" is what `bonfire namespace describe` uses, and it's the one we need. I'll use the same wording

packages/config/src/bin/dev-script.ts
riccardo-forina: 🤦 I forgot to update the hosts file, that's why it was hanging...

karelhala (approved):

PR #2212, opened 2025-05-29
fix(config): use dev image for local fec development against prod

florkbr:
Verified the commit currently running in prod is available in the parallel dev build (we build both dev and prod images for every commit). 

This will unblock PF6 dev for those who cannot use stage auth. If necessary we can revisit to add support for a CLI option to use the prod image.


justinorringer (approved):
Makes sense!

Hyperkid123 (approved):

PR #2213, opened 2025-05-29
chore(config): deprecate config option for interceptChromeConfig

florkbr:
While interceptChromeConfig is not _technically_ doing the same thing as our new FEO config interceptors - it has created confusion when migrating to the FEO (and what is automagically intercepted locally).

I'm proposing we deprecate this option to avoid confusion, however open to other approaches (renaming, etc).


florkbr:
Related to https://issues.redhat.com/browse/RHCLOUD-39513

PR #2218, opened 2025-06-26
feat: updating patternfly components to 6.2.2

tahmidefaz:


.cache_1rcnss
Hyperkid123, line 1: This file should not be pushed.

tahmidefaz:
> @tahmidefaz for this: https://github.com/RedHatInsights/frontend-components/actions/runs/15913012827/job/44884778120?pr=2218#step:6:1577
> 
> Do you know where we may be losing the spy call?

@Hyperkid123 So, I just tried it out on the `master` branch locally with a snapshot update, and it looks like it's happening there as well. It's probably not caused by the patternfly update here. Maybe, the test is broken? idk... 🤔

Hyperkid123:
@tahmidefaz ok, that makes sense. I guess we should update the assertion then. Just make sure to add a comment to the test explaining why there is the number of calls. What is "us" and what is the "them".

PR #2219, opened 2025-07-02
feat: frontend-crd spec update with API version info

justinorringer:
https://issues.redhat.com/browse/RHCLOUD-37859

### DIFF
--- a/packages/config-utils/src/feo/spec/frontend-crd.schema.json
+++ b/packages/config-utils/src/feo/spec/frontend-crd.schema.json
@@ -32,6 +32,26 @@
       ],
       "additionalProperties": false
     },
+    "apiSpecInfo": {
+      "type": "object",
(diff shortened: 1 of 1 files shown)


packages/config-utils/src/feo/spec/frontend-crd.schema.json
```diff
@@ -32,16 +32,44 @@
       ],
       "additionalProperties": false
     },
+    "apiSpecInfo": {
```
florkbr, line 35: Chatted with Justin offline - I'd like to make sure our AJV validation in frontend components mirrors what FEO expects (FEO is enforcing that either `versions` or `specs` is defined when specifying an `API` clause, otherwise an error is thrown). Can we try `anyOf` here?

florkbr (approved):

PR #2224, opened 2025-07-28
chore!: Bump JS client to new version - RBAC client

justinorringer:
https://issues.redhat.com/browse/RHCLOUD-39077

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8048,24 +8048,25 @@
       "resolved": "packages/utils",
       "link": true
     },
-    "node_modules/@redhat-cloud-services/rbac-client": {
-      "version": "1.4.5",
-      "resolved": "https://registry.npmjs.org/@redhat-cloud-services/rbac-client/-/rbac-client-1.4.5.tgz",
-      "integrity": "sha512-7kryZfzEYuw1V7c8bO5esXnfeYQO4Skz1XhluJ2whZNji8HcVbwDjBsQxy0+ndaSVirh5d86fufb00vK6fyUjw==",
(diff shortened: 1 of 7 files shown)


florkbr:
Chatted with Justin offline. This is a necessary change, I just want to be cautious not to break consumers of the effected packages.

1. I'd like to test these changes in a UI that actually uses the packages to make sure the latest RBAC client behaves as expected (I know we have some type issues open against it). It's not exactly clear to me how the rbac-client dependency makes it into the rbac provider either (not listed as a peer dep)
2. I think the safest bet would be to bump major versions of these packages to avoid breaking consumers since we are removing support for v1/v2 rbac client

florkbr:
@justinorringer I think the commit itself needs to include the `!` for major breaking change

florkbr:
@justinorringer I tested this in notifications-frontend. One thing of note, some other frontend packages in this repo pull in axios via dependencies on the utils/components package itself. For example see: https://github.com/RedHatInsights/frontend-components/blob/master/packages/notifications/package.json#L29-L30

Once this major version bump publishes, we will need to update those locations to pull in the new major versions of the utils/components packages, then publish a major version of those as well.

I did also validate that removing `"^1.0.111 || 2.x"` does in fact cause `invalid dependency` issues (at least in notifications frontend), so it is good we are publishing a major version.

florkbr (approved):

florkbr:
Well the jscutlery/semver package did it for us - however it only bumped a patch version (which is a problem). We'll have to take a look: https://github.com/RedHatInsights/frontend-components/commit/5bf28317fc64db685f6edd586955d8685902db8e

PR #2229, opened 2025-08-04
chore: adding landscape pdf param

justinorringer:
https://issues.redhat.com/browse/RHCLOUD-41602

### DIFF
--- a/packages/types/index.ts
+++ b/packages/types/index.ts
@@ -25,6 +25,7 @@ export declare type PDFRequestPayload = {
   scope: string;
   module: string;
   importName?: string;
+  landscape?: boolean;
   fetchDataParams?: Record<string, any>;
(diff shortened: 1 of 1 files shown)


packages/types/index.ts
```diff
@@ -25,6 +25,7 @@ export declare type PDFRequestPayload = {
   scope: string;
   module: string;
   importName?: string;
+  landscape?: boolean;
```
catastrophe-brandon, line 28: Based on the name `PDFRequestPayload` could changes to this type affect the API spec for pdf-generator? Or is this change isolated to frontend-components?

catastrophe-brandon (approved):

florkbr:
FWIW we also need https://github.com/RedHatInsights/pdf-generator/pull/217

PR #2234, opened 2025-08-20
Chrome search api

JetyAdam:
Extending ChromeAPI with Search API.

https://issues.redhat.com/browse/RHCLOUD-37865


JetyAdam:
@catastrophe-brandon I don't think that tests are needed in this repository because I'm just adding some interfaces that will be used and tested in `insights-chrome`.

PR #2242, opened 2025-09-18
feat(useVA): add use VA to chrome API

karelhala:
### Description

As a consumer I should be able to use VA anywhere on the site. Open it from any place or pass a message to it.


Hyperkid123:
@karelhala do we have a ticket for this? Does something require opening VA in a non standard way?

justinorringer:
There were a bunch of tickets for this, I forgot to add it to the type here a few months ago.

justinorringer (approved):
LGTM!

PR #2245, opened 2025-09-23
fix: Group filter & Filter chip fixes

bastilian:
This PR fixes some things in for the PrimaryToolbar/ConditionalFilter.

 * Restores proper groupSelectable behaviour as introduced in https://github.com/RedHatInsights/frontend-components/pull/1112
 * Changes the `FilterChips` to only show a group deletion if there are more than one item

https://github.com/user-attachments/assets/65c7994f-390b-48a9-b8c5-61e0629098ce


raswonders (approved):
Please disregard my approval for now. I've tested the change only in InventoryTable not yet in SystemsTable or tableTools.

PR #2246, opened 2025-10-06
fix(rhcloud-42449): update rbac client inside other packages

Fewwy:
https://issues.redhat.com/browse/RHCLOUD-42449
we should update the package so it won't cause issues in the consumers of frontend components

example https://github.com/RedHatInsights/insights-advisor-frontend/pull/1582


bastilian (approved):
LGTM! Thank you @Fewwy for bumping these!

PR #2254, opened 2025-10-21
feat(RHINENG-15592): Create shared sentry plugin

adonispuente:
I want all apps to be able to adopt a shared plugin so that the 1 package can be managed in far less places

### DIFF
--- a/packages/config/src/lib/createPlugins.ts
+++ b/packages/config/src/lib/createPlugins.ts
@@ -11,24 +11,10 @@ const { glob } = require('glob');
 const path = require('path');
 
 export type WebpackPluginDefinition = undefined | null | false | '' | 0 | ((this: Compiler, compiler: Compiler) => void) | WebpackPluginInstance;
-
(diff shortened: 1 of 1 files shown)


packages/config/src/lib/createPlugins.ts
```diff
@@ -68,6 +83,31 @@ export const createPlugins = ({
     ...(hasTsConfig ? [new ForkTsCheckerWebpackPlugin()] : []),
     ...(plugins || []),
     ...(internalHotReload ? [new ReactRefreshWebpackPlugin()] : []),
+    // Put the sentry plugin at the end of the plugins array
+    ...(process.env.ENABLE_SENTRY && initSentry
```
bastilian: What is the initSentry for? I think we can also leave that out and have no sentry sepcific config within the fec.config.

I'm not sure if it's still needed, but we need to pass through the environment variable in the `fec` script(s), as it's running in a sub-shell that might not have those vars.

PR #2262, opened 2025-12-19
fix(security): ensure vulnerable glob versions are not in use

charlesmulder:
https://issues.redhat.com/browse/RHCLOUD-43646

- Removed glob from packages not using it
- Installed secure version of glob
- Updated nx to resolve incompatible packages
- Updated snapshot so tests will pass


justinorringer (approved):
Looking good here!

PR #2283, opened 2026-03-12
chore(release): explicitly push git commit and tags after nx release

charlesmulder:
## Problem
  After switching to the unified `npx nx release` command, packages were being published to npm but the `chore(release): publish` commit was never pushed back to the repo, preventing nacho-bot from creating PRs to update version field in package.json

  ## Root Cause
  The unified `nx release` command ignores `release.git.push` in nx.json. It only pushes automatically when `createRelease` (GitHub/GitLab releases) is enabled. Since we use `createReleaseInGitHub: false`, no push ever happened — the commit existed only in the ephemeral CI environment.

  This affects nx 21.5.3 (javascript-clients) and nx 22.3.3 (frontend-components) identically.

  ## Fix
  - Add explicit `git push --follow-tags` step after `npx nx release --yes`
  - Remove misleading `push: true` from `release.git` in nx.json since it is ignored by the unified command
  - Sync package.json versions that were published to npm but never committed to the repo: advisor-components, components, notifications, remediations, rule-components

### DIFF
--- a/.github/actions/release/action.yml
+++ b/.github/actions/release/action.yml
@@ -32,4 +32,12 @@ runs:
         # See: https://github.com/nrwl/nx/blob/5ea2e47acfa9d175546daa922ce40905533f1074/packages/nx/src/utils/package-manager.ts#L207-L213
         npm_config_legacy_peer_deps: false
       run: npx nx release --yes
+    - name: Push release commit and tags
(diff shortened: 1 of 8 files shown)


justinorringer (approved):
LGTM!

PR #2300, opened 2026-04-10
fix(release): use npm trusted publisher with OIDC provenance

karelhala:
### Description

Add permissions to perform releases after recent NPM OIDC token auth.

## Summary by CodeRabbit

* **Chores**
  * Tightened CI workflow permissions to limit access to only what's needed for automated releases, reducing surface area for accidental or excessive privileges.
  * Enabled package provenance for release publishing so users can independently verify package origin and integrity before installation.


LightOfHeaven1994 (approved):
LGTM thank you!

PR #2366, opened 2026-07-06
chore(components): add Storybook with stories for all active components [RHCLOUD-48969]

riccardo-forina:
## Summary
- Add Storybook 10 to the components package with 24 story files covering all active, standalone components
- Add test-runner (smoke tests) and CI integration (build, test, Chromatic upload)
- Update `.npmignore` to exclude stories from published packages
- Remove stale `.storybook` entry from `.gitignore`

## Jira
https://redhat.atlassian.net/browse/RHCLOUD-48969

## Test plan
- [ ] `npx nx run @redhat-cloud-services/frontend-components:storybook` starts and all stories render
- [ ] `npx nx run @redhat-cloud-services/frontend-components:test:storybook` passes
- [ ] `npx nx run @redhat-cloud-services/frontend-components:build-storybook` produces static output
- [ ] Existing Cypress and Jest tests unaffected
- [ ] Stories excluded from npm package (`npm pack` in dist folder)
- [ ] CI pipeline runs build-storybook and test-storybook jobs
- [ ] Chromatic upload workflow triggers after CI success

## Summary by CodeRabbit

* **New Features**
  * Added Storybook setup (React Webpack 5) with UI styling and component control matching.
  * Added/expanded component stories to cover common states and variants.
  * Integrated automated Chromatic publishing with PR build URLs and PR comment updates.

* **Tests / CI**
  * CI now builds Storybook to a static bundle and runs Storybook tests against the served build.

* **Chores / Packaging**
  * Updated ignore rules for Storybook development assets and to track the generated Storybook output in the repo workflow.


riccardo-forina:
https://www.chromatic.com/builds?appId=6a4b8018b769a0f10b492cf8

tahmidefaz:
@riccardo-forina Can you look into the coderabbitai's suggestion to see if there are any values to it?

riccardo-forina:
> @riccardo-forina Can you look into the coderabbitai's suggestion to see if there are any values to it?

Yeah I think it's worth doing

charlesmulder (commented):
Claude found some potential issues:

Found 6 issues in this PR (medium effort review):

   ### 🔴 Critical

   **1. Storybook tests block npm releases**
   - **File**: `.github/workflows/ci.yaml:161`
   - **Issue**: `test-storybook` in release job `needs` array blocks publishing
   - **Impact**: Security fix merged to `packages/utils`, unrelated Storybook test fails → cannot publish patched package to npm until Storybook fixed
   - **Fix**: Remove `test-storybook` from release dependencies or make it non-blocking

   **2. Artifact download race condition**
   - **File**: `.github/workflows/chromatic-upload.yml:26`
   - **Issue**: `workflow_run.completed` fires before artifact finalized server-side
   - **Impact**: CI completes → chromatic-upload starts → artifact still processing → download fails with 'Artifact not found'
   - **Fix**: Add `continue-on-error: true` with retry step, or merge workflows to avoid cross-workflow artifact dependency

   ### 🟡 High

   **3. Regex metacharacter injection**
   - **File**: `packages/components/.storybook/main.ts:26`
   - **Issue**: `moduleName` used in RegExp without escaping
   - **Impact**: Module name 'my.module' creates \`/my.module$/\` matching 'myXmodule' instead of literal 'my.module'
   - **Fix**: Escape regex metacharacters: \`moduleName.replace(/[.*+?^${}()|[\]\\\\]/g, '\\\\\\$&')\`

```ts
import { escapeRegExp } from 'lodash';
// ...
const escapedModuleName = escapeRegExp(moduleName);
return [`${url.replace(new RegExp(`${escapedModuleName}$`), `_${moduleName}`)}.scss`, `${url}.scss`];
  ```

   **4. PR comment posted to wrong PR**
   - **File**: `.github/workflows/chromatic-upload.yml:57`
   - **Issue**: PR search limited to 10 most recent, force-push can match wrong PR by SHA
   - **Impact**: 11+ open PRs → no comment. Force-push makes two PRs share SHA → comment on wrong PR
   - **Fix**: Remove fallback search (lines 51-61), rely on \`context.payload.workflow_run.pull_requests\`

   ### 🟠 Medium

   **5. Empty string serialization bug**
   - **File**: `.github/workflows/chromatic-upload.yml:61`
   - **Issue**: github-script returns \`''\`, serializes to \`'\"\"'\` passing truthy check
   - **Impact**: No PR found → returns \`''\` → serializes to two quote chars → passes \`!= ''\` check → API call fails
   - **Fix**: Return \`null\` instead of \`''\`

   **6. Artifact retention too short**
   - **File**: `.github/workflows/ci.yaml:132`
   - **Issue**: \`retention-days: 3\` too short for chromatic reruns
   - **Impact**: Day 4+: manual rerun → artifact expired → download fails
   - **Fix**: Increase to 7-30 days for cross-workflow dependencies

**Comment by riccardo-forina** [comment]
@coderabbit review

File: packages/components/.storybook/main.ts [hunk 1]
```diff
@@ -0,0 +1,98 @@
+import type { StorybookConfig } from '@storybook/react-webpack5';
+import path from 'path';
+import { fileURLToPath } from 'url';
+import { globSync } from 'glob';
+import { readFileSync, existsSync } from 'fs';
+
+const currentDir = path.dirname(fileURLToPath(import.meta.url));
```
karelhala, line 7: **Nit: `import.meta.url` requires ESM context**

This file uses `import.meta.url` which is only available in ES modules. The project's `tsconfig.json` targets `CommonJS` modules — verify that Storybook's build pipeline resolves this correctly (it likely does since Storybook 10 uses its own ESM-aware loader for config files, but worth confirming it works with `npx nx run @redhat-cloud-services/frontend-components:storybook` locally).

packages/components/.storybook/main.ts
```diff
@@ -0,0 +1,98 @@
...
+                      const segments = repoPackage.split('/');
+                      const scopedName = segments[0].startsWith('@') ? `${segments[0]}/${segments[1]}` : segments[0];
+                      const sourcePackage = localPackages[scopedName];
+                      if (sourcePackage) {
+                        const rest = segments[0].startsWith('@') ? segments.slice(2) : segments.slice(1);
+                        return new URL(`file://${path.resolve(sourcePackage, ...rest)}`);
+                      }
+                    }
+
+                    if (url.startsWith('~')) {
```
karelhala, line 66: **Nit: duplicated `url.startsWith('~')` guard**

This is the second `if (url.startsWith('~'))` block — the first one (line 55) handles local packages, this one handles `node_modules`. Both run independently, which works because the first returns early on match. But at a glance it reads like a copy-paste issue.

Consider combining into a single `if (url.startsWith('~'))` block with the local-package check first (already returns on match) and the `node_modules` fallback after, or adding a brief comment explaining the two-phase resolution. Minor readability improvement.

packages/components/src/BulkSelect/BulkSelect.stories.tsx
```diff
@@ -0,0 +1,62 @@
+import type { Meta, StoryObj } from '@storybook/react-webpack5';
+import { fn } from 'storybook/test';
+import BulkSelect from './BulkSelect';
+
+const meta: Meta<typeof BulkSelect> = {
```
karelhala, line 5: **Suggestion (applies to all 24 story files): consider adding `play` functions to key interactive components**

All stories are currently static render-only. The test-runner infrastructure being set up in this PR would provide much more value with `play` functions on interactive components like `BulkSelect`, `ConditionalFilter`, `PrimaryToolbar`, and `FilterDropdown` — e.g., opening dropdowns, toggling checkboxes, verifying ARIA attributes.

This doesn't need to block the PR — the current stories already provide useful smoke-test and Chromatic snapshot coverage. But it would be a good follow-up to leverage the test-runner beyond "does it render."

packages/components/src/CullingInfo/CullingInformation.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+  component: CullingInformation,
+};
+export default meta;
+type Story = StoryObj<typeof CullingInformation>;
+
+export const Fresh: Story = {
+  render: () => (
+    <CullingInformation
+      content=""
+      stale={new Date(Date.now() + 30 * 24 * 60 * 60 * 1000).toISOString()}
```
karelhala, line 15: **Bug: `Date.now()` will cause flaky Chromatic snapshots**

This story (and `DateFormat`'s `Relative` story) compute dates relative to `Date.now()`, producing different rendered output on every build. Chromatic will flag a visual diff on every CI run even when no code changed, creating snapshot noise.

Consider using fixed ISO date strings instead:
```tsx
stale={new Date('2099-01-15T00:00:00Z').toISOString()}
```
Or mock `Date.now` via a Storybook decorator/loader so snapshots are deterministic.

package.json
```diff
@@ -98,7 +98,14 @@
     "ts-node": "^10.9.2",
     "ts-patch": "^3.3.0",
     "typescript": "^5.9.3",
-    "whatwg-fetch": "^3.6.20"
+    "whatwg-fetch": "^3.6.20",
+    "storybook": "^10.4.0",
```
karelhala: **Nit: dependencies not alphabetically sorted**

The new Storybook dependencies are appended after `whatwg-fetch` rather than inserted in alphabetical order. The rest of `devDependencies` follows alphabetical sorting. Consider moving `@storybook/*` entries up near the other `@`-scoped packages and `chromatic`/`storybook` into their sorted positions.

packages/components/.storybook/main.ts
riccardo-forina: Storybook uses its own config, which is ESM capable

packages/components/.storybook/main.ts
riccardo-forina: Agree with this being a nit, won't change much

packages/components/src/BulkSelect/BulkSelect.stories.tsx
riccardo-forina: That would be an improvement for later, this PR only adds the tooling to start testing components with Storybook. Smoke tests of the components are enough for this PR (if a story fails to render, the test will fail)

packages/components/src/CullingInfo/CullingInformation.stories.tsx
riccardo-forina: Agree but we will not use Chromatic's visual snapshots as it's a pay feature (after a small quota)

karelhala (approved):

charlesmulder (approved):

PR #2373, opened 2026-07-09
ci: add stale workflow to auto-close inactive issues and PRs

charlesmulder:
Add stale PR workflow to autoclose stale PRs

## Summary by CodeRabbit

* **Chores**
  * Added an automated workflow to manage stale issues and pull requests.
  * Runs daily at midnight UTC and can also be triggered manually.
  * Supports optional configuration for tuning when items are marked stale and for exempt labels.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,15 @@
+name: Handle stale issues and PRs
+
+on:
+  schedule:
+    - cron: '0 0 * * *' # Run daily at midnight UTC
+  workflow_dispatch: # Allow manual triggers
+
+jobs:
+  call-stale-workflow:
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,15 @@
+name: Handle stale issues and PRs
+
+on:
+  schedule:
+    - cron: '0 0 * * *' # Run daily at midnight UTC
+  workflow_dispatch: # Allow manual triggers
+
+jobs:
+  call-stale-workflow:
+    uses: RedHatInsights/shared-workflows/.github/workflows/stale.yml@master
```
karelhala: I think it's safer to use github SHA in here, so if the shared workflows repository is compromised we will e using safe version.

karelhala (approved):

PR #2375, opened 2026-07-10
ci: add pre-release version verification check

charlesmulder:
The version verification script acts as a pre-release safety gate by ensuring `package.json` versions, git tags, and npm registry versions are aligned before `nx release` runs. It prevents silent corruption scenarios like: (1) incorrect tags being pushed that would cause version skipping on npm, (2) release commit push failures that leave the remote in an inconsistent state (packages published but tags/package.json not updated), and (3) malicious or accidental npm publishes without corresponding git tags. By failing fast with clear diagnostics when versions are misaligned, it catches state corruption before nx release compounds the problem, saving manual cleanup effort and preventing consumer-facing version gaps.

  Script: `.github/scripts/verify-versions.sh`
  - Handles tag disambiguation (frontend-components vs frontend-components-config)
  - Reports mismatches via GitHub Actions error annotations
  - Verified locally: all 14 packages aligned, mismatch detection works

## Summary by CodeRabbit

* **Bug Fixes**
  * Added an automated release validation step to detect inconsistent package versions before publishing.
  * Reports mismatches between package metadata, Git tags, and npm versions.
  * Flags newly added or unpublished packages that lack a corresponding release version.

### DIFF
--- a/.github/scripts/verify-versions.sh
+++ b/.github/scripts/verify-versions.sh
@@ -56,7 +56,7 @@ for dir in "$REPO_ROOT"/packages/*/; do
   [[ -z "$tag_ver" ]] && tag_ver="NONE"
 
   # --- npm registry version ---
-  npm_ver="$(npm view "$pkg" version 2>/dev/null)" || true
+  npm_ver="$(npm view "$pkg" version --fetch-timeout=10000 2>/dev/null)" || true
   [[ -z "$npm_ver" ]] && npm_ver="NOT_ON_NPM"
 
   # --- Compare ---
--- a/.github/scripts/verify-versions.sh
+++ b/.github/scripts/verify-versions.sh
@@ -7,6 +7,8 @@
 #
 # Exit 0: all versions aligned
 # Exit 1: at least one mismatch detected
+# Note: When running locally in a fork, ensure tags are synced first:
+#   git fetch upstream --tags
 
 set -euo pipefail
 
@@ -21,11 +23,6 @@ printf '\n'
 for dir in "$REPO_ROOT"/packages/*/; do
   dirname="$(basename "$dir")"
 
-  # Skip executors -- not released (matches nx.json release.projects exclusion)
-  if [[ "$dirname" == "executors" ]]; then
-    continue
-  fi
-
   pkg_json="$dir/package.json"
   if [[ ! -f "$pkg_json" ]]; then
     continue
@@ -35,6 +32,12 @@ for dir in "$REPO_ROOT"/packages/*/; do
   # avoids jq dependency and handles edge cases better than grep/sed)
   pkg="$(node -p "require('$pkg_json').name")"
   disk_ver="$(node -p "require('$pkg_json').version")"
+
+  # Skip private packages — they are not published to npm or tagged
+  private="$(node -p "require('$pkg_json').private || false")"
+  if [[ "$private" == "true" ]]; then
+    continue
+  fi
 
   # --- Git tag version ---
   # Tag pattern from nx.json: {projectName}-{version} where projectName = npm package name
--- a/.github/scripts/verify-versions.sh
+++ b/.github/scripts/verify-versions.sh
@@ -1,0 +1,86 @@
+#!/usr/bin/env bash
+# verify-versions.sh
+#
+# Compares package versions across three sources (package.json, git tags, npm registry)
+# for all releasable packages in this monorepo. Intended to run before `nx release`
+# in CI to prevent releasing from an inconsistent state.
+#
+# Exit 0: all versions aligned
+# Exit 1: at least one mismatch detected
+
+set -euo pipefail
+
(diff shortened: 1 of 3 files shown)


.github/scripts/verify-versions.sh
```diff
@@ -0,0 +1,86 @@
...
+  # --- Compare ---
+  if [[ "$disk_ver" == "$tag_ver" && "$disk_ver" == "$npm_ver" ]]; then
+    status="OK"
+    marker="✓"
+  elif [[ "$tag_ver" == "NONE" && "$npm_ver" == "NOT_ON_NPM" ]]; then
+    # New package: no tag, not published. Flag as informational but still a mismatch.
+    status="NEW_PACKAGE"
+    marker="!"
+    mismatch=1
+    echo "::error::${pkg}: new/unpublished package (disk=${disk_ver}, tag=NONE, npm=NOT_ON_NPM). Verify this is intentional."
```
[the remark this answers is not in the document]
charlesmulder: Current blocking behavior for NEW_PACKAGE is correct given CI setup.
  
  `nx release --first-release` creates initial tags for new packages ([docs](https://nx.dev/docs/features/manage-releases)), but `.github/actions/release/action.yml:46` runs `npx nx release --yes` without `--first-release` flag. 

 So even if version check passed NEW_PACKAGE, nx release would fail when it can't find existing git tags to determine next version.
 
 Current workflow requires manual creation of first release.

.github/scripts/verify-versions.sh
```diff
@@ -0,0 +1,86 @@
...
+  # avoids jq dependency and handles edge cases better than grep/sed)
+  pkg="$(node -p "require('$pkg_json').name")"
+  disk_ver="$(node -p "require('$pkg_json').version")"
+
+  # --- Git tag version ---
+  # Tag pattern from nx.json: {projectName}-{version} where projectName = npm package name
+  # Use glob "${pkg}-*" then strip the prefix and filter for version-like results.
+  # This handles the ambiguity where e.g. "@redhat-cloud-services/frontend-components-config-*"
+  # also matches "@redhat-cloud-services/frontend-components-config-utilities-*".
+  # After stripping "${pkg}-", the utilities tags become "utilities-4.12.0" which does NOT
```
karelhala, line 47: Missing `git fetch --tags` — GitHub Actions `actions/checkout` does a shallow clone by default and does not fetch tags. `git tag --list` will return nothing, causing every package to show `tag_ver=NONE`. Verify that the calling workflow uses `fetch-depth: 0`, or add an explicit `git fetch --tags` before this script runs.

.github/scripts/verify-versions.sh
```diff
@@ -0,0 +1,86 @@
...
+    git tag --list "${pkg}-*" \
+      | sed "s|^${pkg}-||" \
+      | grep '^[0-9]' \
+      | sort -V \
+      | tail -1
+  )" || true
+  [[ -z "$tag_ver" ]] && tag_ver="NONE"
+
+  # --- npm registry version ---
+  npm_ver="$(npm view "$pkg" version 2>/dev/null)" || true
```
karelhala: Consider adding a timeout: `timeout 30 npm view "$pkg" version 2>/dev/null || true`. With 14 sequential calls and no timeout, a slow/unreachable registry can stall the pipeline for a long time.

.github/scripts/verify-versions.sh
```diff
@@ -0,0 +1,86 @@
...
+  [[ -z "$npm_ver" ]] && npm_ver="NOT_ON_NPM"
+
+  # --- Compare ---
+  if [[ "$disk_ver" == "$tag_ver" && "$disk_ver" == "$npm_ver" ]]; then
+    status="OK"
+    marker="✓"
+  elif [[ "$tag_ver" == "NONE" && "$npm_ver" == "NOT_ON_NPM" ]]; then
```
karelhala, line 66: This creates a chicken-and-egg problem for new packages — `nx release` is the mechanism that creates the first tag and publishes to npm, but this check blocks `nx release` from running. Consider treating `NEW_PACKAGE` as informational (don't set `mismatch=1`), or adding an env-var escape hatch like `ALLOW_NEW_PACKAGES=1`.

.github/scripts/verify-versions.sh
charlesmulder: Fetching of tags is handled by the checkout step in `.github/worksflows/ci.yaml`. I added a comment to the script about manually fetching tags when running the script locally.

.github/scripts/verify-versions.sh
charlesmulder: In this repo, first releases are created manually because trusted publishing (OIDC provenance) requires initial npm  setup that nx release can't bootstrap. So NEW_PACKAGE blocking is intentional — if a new package directory shows up unexpectedly, we want the release to stop and a human to verify it's intentional.

.github/scripts/verify-versions.sh
charlesmulder: I've added 10 sec timeout.

karelhala (approved):

PR #2392, opened 2026-07-23
feat(@redhat-cloud-services/frontend-components-config): release mkcert support

charlesmulder:
Original ticket https://redhat.atlassian.net/browse/RHCLOUD-49050

When PR https://github.com/RedHatInsights/frontend-components/pull/2384 was squash merged, the commit message was changed to a format that failed commitlint thereby preventing release. 

This PR is to retrigger a release using correct commit message.


charlesmulder:
Was released as patch version bump instead of minor version dump. Leaving as is.

